PR #11, opened 2019-06-21
#10 Update Examples

JinHai-CN:
#10

### DIFF
--- a/examples/example.py
+++ b/examples/example.py
@@ -1,5 +1,5 @@
 from milvus import Milvus, Prepare, IndexType, Status
-import random
+import random, time
 from pprint import pprint
 
 _HOST = 'localhost'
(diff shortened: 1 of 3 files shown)


examples/example.py
```diff
@@ -22,40 +22,38 @@ def main():
...
-    # Check if `test01` exists, if not, create a table test01
-    table_name = 'test01'
+    table_name = 'table01'
     res_status, table = milvus.describe_table(table_name)
     print('# Describe table status: {}'.format(res_status))
     print('# Describe table:{}'.format(table))
 
     # Create table
-    #   01.Prepare data
+    # Check if `test01` exists, if not, create a table test01
```
XuanYang-cn: table name should be `table01`

PR #78, opened 2019-08-03
add indextype and upgrade version 0.1.25

BossZou:


BossZou:
#79 fix indextype and modify changlog

PR #355, opened 2020-11-10
fix class Entity `str` func error

2h4dl:
Fix a bug that `Entity.__str__()` not work as expected.


BossZou:
@2h4dl you are recommended to pull this request on branch master.

PR #525, opened 2021-05-26
Copy search params to avoid changing the passed params

DragonDriver:


DragonDriver:
This pr will fix [#5491](https://github.com/milvus-io/milvus/issues/5491).

PR #537, opened 2021-06-02
Remove tests

scsven:


milvus/client/pool.py
```diff
@@ -9,7 +9,6 @@
 
 from . import __version__
 from .grpc_handler import GrpcHandler, RegistryHandler
-from .http_handler import HttpHandler
```
XuanYang-cn, line 12: Are we not going to support Http request?

Makefile
```diff
@@ -0,0 +1,6 @@
+
+unittest:
+	python -m pytest tests
+
+package:
+	python -m build --sdist --wheel --outdir dist/.
```
XuanYang-cn, line 6: Maybe better if we use `python3`

We need to add `build` module into requirements.txt

XuanYang-cn (changes requested):

milvus/client/pool.py
scsven: These codes are outdated. And we will not support HTTP in near future.

Makefile
scsven: Changed to python3.
build already in requirements.txt.

PR #542, opened 2021-06-03
Rename package

scsven:


XuanYang-cn (changes requested):
Please change README.md to notify everyone

Maybe commit message could be more specific like `Rename package name to pymilvus`

XuanYang-cn (approved):

PR #546, opened 2021-06-04
Add query rpc in proto

scsven:


XuanYang-cn:
`milvus.proto` in milvus and pymilvus **should be identical**, if not, further maintenance would be like in hell.

As least every time when I need to update something in `milvus.proto`, I'll just copy the file into pymilvus.

So please make sure they are identical.

scsven:
> `milvus.proto` in milvus and pymilvus **should be identical**, if not, further maintenance would be like in hell.
> 
> As least every time when I need to update something in `milvus.proto`, I'll just copy the file into pymilvus.
> 
> So please make sure they are identical.

These two files are consistent. Please check: https://github.com/milvus-io/milvus/blob/master/internal/proto/milvus.proto

XuanYang-cn:
> > `milvus.proto` in milvus and pymilvus **should be identical**, if not, further maintenance would be like in hell.
> > As least every time when I need to update something in `milvus.proto`, I'll just copy the file into pymilvus.
> > So please make sure they are identical.
> 
> These two files are consistent. Please check: https://github.com/milvus-io/milvus/blob/master/internal/proto/milvus.proto

That'll be great, just a reminder.

BTW It would be much less misleading if you attach an origin PR or ISSUE to this proto change.

scsven:
Related PR in milvus: https://github.com/milvus-io/milvus/pull/5592

PR #564, opened 2021-06-18
Retry when rpc failure occurs

DragonDriver:

### DIFF
--- a/pymilvus/client/stub.py
+++ b/pymilvus/client/stub.py
@@ -5,6 +5,7 @@ import functools
 import logging
 import time
 
+import grpc
 from urllib.parse import urlparse
 
 from . import __version__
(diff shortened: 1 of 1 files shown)


pymilvus/client/stub.py
```diff
@@ -41,6 +42,30 @@ def inner(self, *args, **kwargs):
...
 
+def retry_on_rpc_failure(retry_times=10, wait=1):
+    def wrapper(func):
+        @functools.wraps(func)
+        def handler(self, *args, **kwargs):
+            counter = 1
+            try:
+                return func(self, *args, **kwargs)
+            except grpc.RpcError as e:
+                if e.code() != grpc.StatusCode.DEADLINE_EXCEEDED and e.code() != grpc.StatusCode.UNAVAILABLE:
```
scsven, line 53: If the exception is DEADLINE_EXCEEDED or UNAVAILABLE, these code behavior? 
I think this decorator need some comments to explain behavior.

scsven (changes requested):

pymilvus/client/stub.py
DragonDriver: According to the [doc of grpc](https://grpc.github.io/grpc/python/grpc.html#grpc-status-code), DEADLINE_EXCEEDED means the task was not completed, UNAVAILABLE means that service is not reachable currently.

PR #571, opened 2021-06-22
Complement entity class

fishpenguin:
Signed-off-by: fishpenguin <kun.yu@zilliz.com>

### DIFF
--- a/pymilvus/client/abstract.py
+++ b/pymilvus/client/abstract.py
@@ -125,7 +125,7 @@ class CollectionSchema:
         self.params = dict()
         self.fields = list()
         self.statistics = dict()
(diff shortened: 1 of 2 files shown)


pymilvus/client/abstract.py
```diff
@@ -184,13 +184,19 @@ def id(self):
...
+        for k, v in self._row_data.items():
+            fields.append(k)
+        return fields
 
     def get(self, field):
         return self.value_of_field(field)
 
     def value_of_field(self, field):
-        raise NotImplementedError('TODO: support field in Hits')
+        if field in self._row_data:
```
czs007, line 196: Do the fields contain the primary field?

fishpenguin: yes

PR #594, opened 2021-07-05
Optimize the workflow of future object

DragonDriver:


DragonDriver:
Reference: [future object of grpc](https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/framework/foundation/future.py)

PR #660, opened 2021-08-24
Revealing shards in create_collection

filip-halt:
Revealing the shards argument when creating a collection since it is useful for larger scale datasets. Does not solve/limit the problem of too many growing segments.


XuanYang-cn:
@filip-halt Great! Could you please squash the 5 commits into 1 commit? We want to make our commit history as simple and clear as possible.

PR #673, opened 2021-08-30
Remove retry logic when drop collection with deadline error

DragonDriver:

issue: https://github.com/milvus-io/milvus/issues/7306

### DIFF
--- a/pymilvus/client/stub.py
+++ b/pymilvus/client/stub.py
@@ -42,7 +42,7 @@ def check_connect(func):
     return inner
 
 
-def retry_on_rpc_failure(retry_times=10, wait=1):
+def retry_on_rpc_failure(retry_times=10, wait=1, retry_on_deadline=True):
     def wrapper(func):
         @functools.wraps(func)
         def handler(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


pymilvus/client/stub.py
```diff
@@ -42,7 +42,7 @@ def inner(self, *args, **kwargs):
     return inner
 
 
-def retry_on_rpc_failure(retry_times=10, wait=1):
+def retry_on_rpc_failure(retry_times=10, wait=1, retry_on_deadline=True):
```
czs007, line 45: just rename retry_on_deadline to  `enable`, deafult  `enable=True`

DragonDriver: so what if we will distinguish the behavior of handling `UNAVAILABLE`?
maybe you mean `enable_on_deadline` and default `enable_on_deadline=True`?

czs007: Add it（`enable_on_deadline`） when we really want to distinguish. Anyway, this `enable` to seems reasonable and will be reserved even if we add `enable_on_dealine`.

Issue #651, opened 2021-08-20
Timeout not modifiable


natemccoy:
https://github.com/milvus-io/pymilvus/blob/229f5aa48eb4ef6020508714f562b855b291a1f6/pymilvus/client/stub.py#L179

Milvus client does not allow for timeout changes even thought the kwargs could support it on instantiation of the `Milvus` class.

Issue occurs as no args are passed to `_wait_for_healthy` function

XuanYang-cn:
Thanks for pointing it out.

I think it's a simple change

```python
def _wait_for_healthy(self, timeout=30, retry=10):
    _timeout_on_every_retry = self._kw.get("timeout", None)
    _timeout = _timeout_on_every_retry if _timeout_on_every_retry else timeout
    ...
        if (time.time() - start_time > _timeout):
    ...
            status = handler.fake_register_link(_timeout)
```

Would you like to fix the bug?

BTW, If this's no feedbacks for several days, I'll fix it myself.

PR #681, opened 2021-08-31
Make wait for healthy timeout controllable

XuanYang-cn:
Resolves: #651

### DIFF
--- a/pymilvus/client/stub.py
+++ b/pymilvus/client/stub.py
@@ -134,17 +134,19 @@ class Milvus:
 
     @check_connect
     def _wait_for_healthy(self, timeout=30, retry=10):
+        _timeout_on_every_retry = self._kw.get("timeout", None)
+        _timeout = _timeout_on_every_retry if _timeout_on_every_retry else timeout
         with self._connection() as handler:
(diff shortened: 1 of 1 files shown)


pymilvus/client/stub.py
```diff
@@ -134,17 +134,19 @@ def __init__(self, host=None, port=None, handler="GRPC", pool="SingletonThread",
 
     @check_connect
     def _wait_for_healthy(self, timeout=30, retry=10):
+        _timeout_on_every_retry = self._kw.get("timeout", None)
```
czs007, line 137: @check_connect
    def _wait_for_healthy(self, timeout=3, retry=10):
        with self._connection() as handler:
            while retry > 0:
                is_except = False
                try:
                    status = handler.fake_register_link(timeout)
                    if status.error_code == 0:
                        self._deploy_mode = status.reason
                        return
                except:
                    is_except = True
                finally:
                    sleep_time = 1 if is_except else timeout
                    time.sleep(sleep_time)
                    retry -= 1
            raise Exception("server is not healthy, please try again later")
            
          Can you refer to the logic above?

Issue #689, opened 2021-09-03
[BUG] The value of standalone deployment key differs from server


DragonDriver:
**Describe the bug**

`pymilvus/client/types.py`:
```python
class DeployMode:
    Distributed = "DISTRIBUTED"
    StandAlone = "SINGLE_NODE"
```

However, the value of `StandAlone` defined in server now is `STANDALONE`.

PR #691, opened 2021-09-06
Change value of standalone deployment key to keep same as server.

JackLCL:
Fixes: #689


DragonDriver:
/lgtm
thx for your contribution to PyMilvus and we will appreciate if you continuously help the community to be better.

PR #706, opened 2021-09-13
Update SchemaNotReady exception msg when collection not exist

cydrain:
Signed-off-by: yudong.cai <yudong.cai@zilliz.com>

https://github.com/milvus-io/milvus/issues/7165

### DIFF
--- a/pymilvus/orm/pymilvus_orm/collection.py
+++ b/pymilvus/orm/pymilvus_orm/collection.py
@@ -133,7 +133,7 @@ class Collection:
 
         else:
             if schema is None:
-                raise SchemaNotReadyException(0, ExceptionsMessage.NoSchema)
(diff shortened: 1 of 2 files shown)


pymilvus/orm/pymilvus_orm/collection.py
```diff
@@ -133,7 +133,7 @@ def __init__(self, name, schema=None, using="default", shards_num=2, **kwargs):
 
         else:
             if schema is None:
-                raise SchemaNotReadyException(0, ExceptionsMessage.NoSchema)
+                raise SchemaNotReadyException(0, ExceptionsMessage.CollectionNotExistNoSchema % name)
```
scsven, line 136: This is an old syntax, please use f-string instead. It looks like 
```
raise Exception(0, "Collection {name} not exist")
```

scsven (changes requested):

PR #707, opened 2021-09-14
Adding examples and partition info

filip-halt:
Adding simple and advanced example. We should primarily keep just these two as they incorporate all of milvus's features and are easier to understand line by line. Also aligning partition return to match docs.

### DIFF
--- a/examples/advanced_example.py
+++ b/examples/advanced_example.py
@@ -1,0 +1,81 @@
+import random
+import pymilvus.orm.pymilvus_orm as orm
+from pymilvus import DataType
+from pprint import pprint
+random.seed(1234567890)
+
+
+def main():
+   
+    orm.connections.connect("default", host = "127.0.0.1", port= 19530)
+    orm.connections.connect("collection_con", host = "127.0.0.1", port= 19530)
+
+    try:
+        collection = orm.collection.Collection(name="advanced_example")
+        collection.drop()
+    except:
+        pass
+
+    dim = 128
+
+    id_field = orm.schema.FieldSchema(name="ids_field", dtype=DataType.INT64, description="primary ids", auto_id=True, is_primary=True)
+    vector_field = orm.schema.FieldSchema(name="vectors_field", dtype=DataType.FLOAT_VECTOR, dim=dim, description="vectors")
+    scalar_field = orm.schema.FieldSchema(name="scalar_field", dtype=DataType.INT64, description="numbers")
+    int_boolean_field = orm.schema.FieldSchema(name="int_boolean_field", dtype=DataType.INT64, description="0 or 1")
+
+    schema = orm.schema.CollectionSchema(fields =[id_field, vector_field, scalar_field, int_boolean_field], description="collection of vectors")
+    
+    collection = orm.collection.Collection(name="advanced_example", schema=schema, using="collection_con")
+
+    partitions = []
+
+    for x in range(1,4):
+        partition = orm.partition.Partition(collection=collection, name="partition_" + str(x), descrition="partition " + str(x) + " of 3")
+        partitions.append(partition)
+
+    print("Partitions before insert: ")
+    pprint(collection.partitions)
+
+    for x in partitions:
+        amount = 100000
+        store_vectors = [[random.random() for _ in range(dim)] for _ in range(amount)]
+        store_a = [random.randint(1,amount) for _ in range(amount)]
+        store_b = [random.randint(0,1) for _ in range(amount)]
+
+        x.insert([store_vectors, store_a, store_b])
+
+    print("Partitions after insert: ")
+    pprint(collection.partitions)
+
+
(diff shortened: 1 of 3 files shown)


examples/advanced_example.py
```diff
@@ -0,0 +1,81 @@
...
+        collection.drop()
+    except:
+        pass
+
+    dim = 128
+
+    id_field = orm.schema.FieldSchema(name="ids_field", dtype=DataType.INT64, description="primary ids", auto_id=True, is_primary=True)
+    vector_field = orm.schema.FieldSchema(name="vectors_field", dtype=DataType.FLOAT_VECTOR, dim=dim, description="vectors")
+    scalar_field = orm.schema.FieldSchema(name="scalar_field", dtype=DataType.INT64, description="numbers")
+    int_boolean_field = orm.schema.FieldSchema(name="int_boolean_field", dtype=DataType.INT64, description="0 or 1")
```
XuanYang-cn, line 24: According the import statement reviewed above, these could be simplified into

```python
id_field = FieldSchema(...)
...
```

examples/advanced_example.py
```diff
@@ -0,0 +1,81 @@
...
+
+    schema = orm.schema.CollectionSchema(fields =[id_field, vector_field, scalar_field, int_boolean_field], description="collection of vectors")
```
XuanYang-cn, line 26: According the import statement reviewed above, this could be simplified into
```python
schema = CollectionSchema(...)
```

examples/advanced_example.py
```diff
@@ -0,0 +1,81 @@
...
+    
+    collection = orm.collection.Collection(name="advanced_example", schema=schema, using="collection_con")
```
XuanYang-cn, line 28: According the import statement reviewed above, this could be simplified into
```python
examle_collection = Collection(...)
```

examples/advanced_example.py
```diff
@@ -0,0 +1,81 @@
...
+import pymilvus.orm.pymilvus_orm as orm
+from pymilvus import DataType
+from pprint import pprint
+random.seed(1234567890)
+
+
+def main():
+   
+    orm.connections.connect("default", host = "127.0.0.1", port= 19530)
+    orm.connections.connect("collection_con", host = "127.0.0.1", port= 19530)
```
XuanYang-cn, line 11: According to the import statement reviewed above, this could be simplified into

```python
connections.connect("collection_con", host="127.0.0.1", port=19530)
```
Since "default" and "collection_con" are aliases for the same server address, line10 should be removed.

XuanYang-cn (changes requested):

pymilvus/orm/pymilvus_orm/partition.py
```diff
@@ -42,6 +42,7 @@ def __repr__(self):
             'name': self.name,
             'collection_name': self._collection.name,
             'description': self.description,
+            'num_entities': self.num_entities
```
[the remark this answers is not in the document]
filip-halt: Should we change https://milvus.io/docs/v2.0.0/create.md and https://milvus.io/api-reference/pymilvus/v2.0.0rc6/api/collection.html#pymilvus.Collection.partitions? Currently the tutorial and api reference shows that collections.partitions displays num_entities

examples/advanced_example.py
```diff
@@ -0,0 +1,81 @@
+import random
+import pymilvus.orm.pymilvus_orm as orm
```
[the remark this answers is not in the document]
filip-halt: This is one where i am a bit conflicted. As a tutorial/example, we should show what we are calling and where we are getting it from. I believe pep8 recommends absolute paths, but I may be wrong. . In [sqalchemy](https://github.com/sqlalchemy/sqlalchemy/blob/master/examples/generic_associations/discriminator_on_association.py) examples they show full paths, as it makes everything easier to understand and look up as a new user. This gets even more useful as we currently have orm and pymilvus in the same directory. How would one immediately know which one is being used?

examples/advanced_example.py
filip-halt: This was to show how to use alternate connections, relating to line 28. There is very little guidance on how to use different connections for Collections, Index, and Partitions, so I wanted to show the example. Its useful because when you are doing threaded inserts you dont want threads to use the same default connection.

pymilvus/orm/pymilvus_orm/partition.py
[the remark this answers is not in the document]
XuanYang-cn: Good catch, num_entites with the flush is a problem. To solve it, the Milvus server also needs a major change, and there are still no MEPs to solve this problem yet.

In PyMilvus, I tend to keep what's already there and don't put `num_entities` in new places. Just don't make things more complicated while it's still under discussion.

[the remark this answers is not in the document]
filip-halt: Makes sense, Ill remove it.

XuanYang-cn:
2 months since the last response, closing this PR.

Feel free to reopen one.

Issue #701, opened 2021-09-08
Add __str__ methods for SearchResult and Hits objects


NotRyan:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. E.g. I wish I could use Milvus to do [...]
There are no __str__ methods for SearchResult or Hits objects; making it more difficult to print out their contents when debugging. This is only a minor inconvenience when debugging and isn't a high priority, but it would be nice to have.

**Describe the solution you'd like**
Add __str__ methods to SearchResult and Hits objects that list out their contents.

**Describe alternatives you've considered**
Iterating through the SearchResult and Hits objects works fine; it is just slightly less convenient than a __str__ method when inspecting results manually for debugging. 

**Additional context**
There is currently already a __str__ method for individual Hit objects.

PR #715, opened 2021-09-17
Adds __str__ for baseloop (Search results)

NotRyan:
In reference to #701.
This adds a __str__ method to baseloop, which is a parent of our hits object. Now hits can be printed directly, which will simply list out a string representation of all the contents. See image below. 
Screen Shot 2021-09-16 at 6 50 31 PM

Signed-off-by: NotRyan <ryan.chan@zilliz.com>

### DIFF
--- a/pymilvus/client/abstract.py
+++ b/pymilvus/client/abstract.py
@@ -45,7 +45,7 @@ class LoopBase(object):
         raise StopIteration()
 
     def __str__(self):
-        return str(list(map(str, self.__getitem__(slice(None)))))
+        return str(list(map(str, self.__getitem__(slice(0, 10)))))
 
     @abc.abstractmethod
     def get__item(self, item):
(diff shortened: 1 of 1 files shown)


pymilvus/client/abstract.py
```diff
@@ -44,6 +44,9 @@ def __next__(self):
         self.__index = 0
         raise StopIteration()
 
+    def __str__(self):
+        return str(list(map(str, self.__getitem__(slice(None)))))
```
XuanYang-cn, line 48: This could be a catastrophe if the top-k/limits is massive, for example, when topk is greater than 10.

Please make a boundary for the returned string. As

```python
return str(list(map(str, self.__getitem__(slice(0, 10)))))
```

XuanYang-cn (changes requested):

pymilvus/client/abstract.py
NotRyan: Requested changes have been made.

NotRyan:
Requested changes have been made.

XuanYang-cn (commented):
/lgtm
/approve

PR #736, opened 2021-09-29
Update ORM SearchResult and Hits objects to use consistent types for index gets and slice gets

NotRyan:
Currently, getting the contents of an ORM SearchResult or an ORM Hits object will result in a differently typed version of the contents depending on whether the get is done with a slice or an index. See this image for an example:

Screen Shot 2021-09-29 at 4 34 43 PM

It does not make sense for one to return objects of different types; logically they should be the same. This PR changes it so that all objects are returned as the corresponding ORM type. See the following image: 

Screen Shot 2021-09-29 at 4 37 39 PM

If we would rather have the objects return the other type, let me know and I can change it to that. But whatever they are, it should be consistent. 

Signed-off-by: NotRyan <ryan.chan@zilliz.com>

### DIFF
--- a/pymilvus/orm/search.py
+++ b/pymilvus/orm/search.py
@@ -180,7 +180,7 @@ class Hits:
             for i in range(_start, _end):
                 elements.append(self.on_result(s[i]))
             return elements
-        return s
+        return self.on_result(s)
 
     def __len__(self) -> int:
         """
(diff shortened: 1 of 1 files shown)


XuanYang-cn:
@NotRyan the commit message is too long, please make it lesser than 50 characters

NotRyan:
@XuanYang-cn I have amended the commit message

PR #743, opened 2021-10-04
Improve client.check

gruebel:
Most of the `len()` checks are not needed and by removing them the performance is also improved. Using a `tuple` instead of a `list` for lookups is also faster. 

I also made the code `pylint` complaint and added missing type hints.

Are you interested in making your whole project `pylint` complaint and need some help with it?


XuanYang-cn:
@gruebel Great enhancement! But I'm afraid you have to do one more change.

Please squash your 2 commits into 1, The commit history would be so much clearer if it's 1 commit/PR.

After the commits being squashed, it's ready to be tested on GitHub actions, once passed, this PR is ready to merge.

Good day.

XuanYang-cn:
BTY, just let me know if you need a Hacktoberfest label   ; -)

Issue #813, opened 2021-11-12
[FEATURE]: Change default consistency level to session-consistency


DragonDriver:
### Is there an existing issue for this?

- [X] I have searched the existing issues

DragonDriver:
related: https://github.com/milvus-io/milvus/issues/9618

PR #814, opened 2021-11-12
Set default consistency level to session

DragonDriver:
issue: https://github.com/milvus-io/pymilvus/issues/813 https://github.com/milvus-io/milvus/issues/9618


DragonDriver:
Closed because we need to maintain the `last_write_ts` for every collection.
So it's better to keep this state in `orm`.

PR #823, opened 2021-11-18
Revert "Set default consistency level to session"

DragonDriver:
Reverts milvus-io/pymilvus#821


DragonDriver:
My mistake, I forgot to add callback on result of delete. By doing so, we can still search the deleted records.

Issue #881, opened 2022-02-09
[FEATURE]: gRPC secure channel support


zwd1208:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

pymilvus is not support secure gRPC channel now. but AWS ALB only supported listener protocol is HTTPS for gRPC protocol version.
Reference documentation:
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-protocol-version

### Describe the solution you'd like

So I think it need to support secure gRPC channel as an optional

XuanYang-cn:
- [ ] #883 This PR makes it work to communicate with AWS. @DragonDriver 
- [ ] Supporting init secure channel with `credentials` and `private key`.

XuanYang-cn:
#967

PR #883, opened 2022-02-09
Support secure gRPC channel for AWS

DragonDriver:
issue: #881

### DIFF
--- a/pymilvus/client/grpc_handler.py
+++ b/pymilvus/client/grpc_handler.py
@@ -76,6 +76,7 @@ class GrpcHandler:
             self._uri = uri
         self._max_retry = kwargs.get("max_retry", 5)
 
+        self._secure = kwargs.get("secure", False)
         self._setup_grpc_channel()
 
     def __enter__(self):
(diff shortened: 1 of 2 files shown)


DragonDriver:
You can just establish a secure connection via:
```python
connections.connect(host, port, secure=True)
```
In old style API:
```python
client = Milvus(host, port, secure=True)
```

pymilvus/client/grpc_handler.py
```diff
@@ -100,13 +101,24 @@ def close(self):
...
+            if not self._secure:
+                self._channel = grpc.insecure_channel(
+                    self._uri,
+                    options=[(cygrpc.ChannelArgKey.max_send_message_length, -1),
+                             (cygrpc.ChannelArgKey.max_receive_message_length, -1),
+                             ('grpc.enable_retries', 1),
+                             ('grpc.keepalive_time_ms', 55000)]
+                )
+            else:
+                creds = grpc.ssl_channel_credentials(root_certificates=None, private_key=None, certificate_chain=None)
```
XuanYang-cn, line 113: Is it valid that we don't provide root certificates and private_key?

XuanYang-cn:
@DragonDriver Mostly lgtm, could you please change the commit message to something like `Support gRPC secured channel for AWS`

This PR seems a partial impl for #881 for AWS only.

PR #885, opened 2022-02-12
support for numpy/ Tensorflow/ Pytorch vectors

aquibjaved:
Added the support for different vector types, using this any vector can be passed for indexing.

### DIFF
--- a/pymilvus/orm/collection.py
+++ b/pymilvus/orm/collection.py
@@ -534,10 +534,10 @@ class Collection:
         # The following code will take care of different vector type
         if not isinstance(data[2], list):
             try:
-                data[2]=data[2].tolist() # numpy or torch tensor
+                data[2]=data[2].tolist() # check for numpy or torch tensor and convert to list
             except AttributeError:
                 try:
-                    data[2] = np.array(data[2]).tolist() #Tf tensor
+                    data[2] = np.array(data[2]).tolist() # for tensorflow tensor to list
                 except AttributeError:
                     raise DataTypeNotSupportException(0, ExceptionsMessage.DataTypeNotSupport)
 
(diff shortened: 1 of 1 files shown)


anukaal (approved):

XuanYang-cn:
@anukaal This is a great idea, since we support Dataframe, at least we can support numpy.ndarry for vector fields.

However, I've checked the code, and it's a lot more complicated than this commit; your's patch has no difference than just transferring `ndarray` to `list` from **outside of pymilvus**, and it doesn't cover Partition.py and Milvus() class.

To support `ndarrary` at the bottom of the PyMilvus, there're several places to modify, the first one is: https://github.com/milvus-io/pymilvus/blob/a6a8b8e9da2152f9109a539891a94b12bf243ad9/pymilvus/orm/prepare.py#L60, and so on.

It'll be more complicated than this PR, I've opened issue #890 for this task, would you like to work on it?

anukaal:
> @anukaal This is a great idea, since we support Dataframe, at least we can support numpy.ndarry for vector fields.
> 
> However, I've checked the code, and it's a lot more complicated than this commit; your's patch has no difference than just transferring `ndarray` to `list` from **outside of pymilvus**, and it doesn't cover Partition.py and Milvus() class.
> 
> To support `ndarrary` at the bottom of the PyMilvus, there're several places to modify, the first one is:
> 
> https://github.com/milvus-io/pymilvus/blob/a6a8b8e9da2152f9109a539891a94b12bf243ad9/pymilvus/orm/prepare.py#L60
> 
> , and so on.
> It'll be more complicated than this PR, I've opened issue #890 for this task, would you like to work on it?

Yeah I would like to work on this issue

XuanYang-cn:
@anukaal Great! I'll close this PR and we could discuss the solutions on #890. Maybe we could reach a consensus today : )

PR #908, opened 2022-03-06
increase pandas range for python 3.8

wallies:


wallies:
https://pandas.pydata.org/pandas-docs/version/1.4.1/whatsnew/v1.4.0.html#backwards-incompatible-api-changes
```
pandas 1.4.0 supports Python 3.8 and higher.
```

XuanYang-cn:
@wallies Supporting a higher version of Pandas is definitely a great idea. 

I'm looking into the codes of the PR, remind me if I don't respond to this PR in three days, you can get my email from the sign-off of my commits.

wallies:
@XuanYang-cn how do we move this along. We use pymilvus-legacy library which is just this PR with pymilvus 2.0.1. We use thid in production and works really well with Python 3.9

.github/workflows/nightly_ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6]
+        python-version: [3.6, 3.8]
```
XuanYang-cn, line 21: nightly cannot pass with python3.8, so please remove this.

.github/workflows/publish_dev_package.yml
```diff
@@ -9,6 +9,9 @@ jobs:
   build-n-publish:
     name: Build and publish Python 🐍 distributions 📦 to TestPyPI
     runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8]
```
XuanYang-cn, line 14: This action is to publish dev package in testpypi, we certainly don't need to publish twice, so please remove 3.7 here

XuanYang-cn (changes requested):

setup.py
```diff
@@ -27,6 +27,7 @@
         "mmh3>=2.0,<=3.0.0",
         "pandas==1.1.5; python_version<'3.7'",
         "pandas>=1.2.4,<=1.3.5; python_version>'3.6'",
+        "pandas>=1.2.4,<=1.4.1; python_version>='3.8'",
```
XuanYang-cn, line 30: I've tested with "pandas>=1.2.4; python_version>='3.6'", and it worked perfectly fine.

So please remove L30 and alter L29 into
```
"pandas>1.2.4; python_version>3.6"
```

.github/workflows/publish_dev_package.yml
XuanYang-cn: It can be built perfectly locally, but the built package does not need to be uploaded into testpypi. Besides, the publishing python environment doesn't influence the actual building process. 

So please remove 3.7 here, just leave python3.8 here.

.github/workflows/publish_dev_package.yml
```diff
@@ -17,16 +20,19 @@ jobs:
...
-        python-version: 3.7
+        python-version: ${{ matrix.python-version }}
+
     - name: Install pypa/build
       run: >-
         python -m
         pip install
         build
         --user
+
```
XuanYang-cn, line 35: This is not needed since we don't need a publishing matrix.

XuanYang-cn (changes requested):

XuanYang-cn:
> @XuanYang-cn how do we move this along. We use pymilvus-legacy library which is just this PR with pymilvus 2.0.1. We use thid in production and works really well with Python 3.9

@wallies This PR needs some refinement to be qualified to merge, please help to improve the quality and remove some unnecessary GitHub action changes.

XuanYang-cn:
Tell me if you encounter any problems. I'm very happy to help.

setup.py
wallies: thanks @XuanYang-cn. Just make sure when building the package it is built with different python version, as the new pandas > 1.4 only works with python 3.8 and above

XuanYang-cn: Yes, pip will figure it out automatically. With python3.7, pandas1.4.1 cannot be downloaded, you could try locally.

pip will only download the highest pandas version that satisfies python3.7, which is 1.3.5. 

Here's a ref https://pypi.org/project/pandas/1.4.1/ , please pay attention the programming language part.

image

setup.py
```diff
@@ -23,10 +23,10 @@
     install_requires=[
         "grpcio==1.37.1",
         "grpcio-tools==1.37.1",
-        "ujson>=2.0.0,<=5.1.0",
+        "ujson>=1.3.5,<=5.1.0",
```
XuanYang-cn, line 26: I'm afraid ujson doesn't have version 1.3.5, please change back to 2.0.0

Here's a ref https://pypi.org/project/ujson/#history

XuanYang-cn (changes requested):

setup.py
```diff
@@ -23,10 +23,10 @@
...
         "mmh3>=2.0,<=3.0.0",
         "pandas==1.1.5; python_version<'3.7'",
-        "pandas>=1.2.4,<=1.3.5; python_version>'3.6'",
+        "pandas>1.2.4; python_version>3.6"
```
XuanYang-cn, line 29: Sorry for my misleading line, there should be a pair of single quote around `3.6`, just like '3.6'

XuanYang-cn (changes requested):

setup.py
```diff
@@ -23,10 +23,10 @@
     install_requires=[
         "grpcio==1.37.1",
         "grpcio-tools==1.37.1",
-        "ujson>=2.0.0,<=5.1.0",
+        "ujson>=1.35,<=5.1.0",
```
XuanYang-cn, line 26: ujson1.35 at most only support python3.2, it's not a proper required version for pymilvus, which supports python >=3.6.

So please change it back to  2.0.0, I don't see any reason why we need to change ujson's version here.

image

https://pypi.org/project/ujson/1.35/

image

https://pypi.org/project/ujson/2.0.0/

XuanYang-cn (changes requested):

XuanYang-cn:
The CI team requires it to be fixed as soon as possible, So I created another PR #919 to fix this problem;

This will be available on pymilvus==2.0.2 soon and a dev version(pymilvus==2.0.2.dev9) is already available on `test.pypi`.

So I'm closing this PR now. Thanks again for pointing this out! Wish you wonderful day!

PR #930, opened 2022-03-31
Fix index_name in Index.drop()

DragonDriver:
/kind bug


xige-16:

PR #943, opened 2022-04-04
update requirements

filipecaixeta:
For some reason, all the requirements have a fixed version
The main dependencies for pymilvus are the grpc packages so I fixed a minimum and maximum version
I'm having problems installing pymilvus with other google libraries because pymilvus has a fixed version with an old version of grpc libraries.


Jacobh2:
Any updates on this? 😄  Would be nice to get in, we're using `grpcio` lib for other reasons in combo with PyMilvus, where we need the higher versions!

aditya-nambiar:
Any update on this ? I'm unable to install pymilvus cause of grpcio 1.37 fails but i am able to install a higher version 1.46

XuanYang-cn:
#1033 would fix this problem, thanks for this PR

PR #948, opened 2022-04-08
Implement BulkLoad APIs

XuanYang-cn:
See also: https://wiki.lfaidata.foundation/display/MIL/MEP+24+--+Support+bulk+load

### DIFF
--- a/pymilvus/client/prepare.py
+++ b/pymilvus/client/prepare.py
@@ -880,3 +880,23 @@ class Prepare:
             with_shard_nodes=True,
         )
         return request
+
+    @classmethod
+    def bulk_load(collection_name: str, partition_name: str, is_row_based: bool, files: list, **kwargs):
+        req = milvus_types.ImportRequest(
+            collection_name=collection_name,
+            partition_name=partition_name,
+            row_based=is_row_based,
+            files=files,
+        )
+
+        for k, v in kwargs.items():
+            if k in ("bucket",):
+                kv_pair = common_types.KeyValuePair(key=str(k), value=str(v))
+        req.options.append(kv_pair)
+        return req
+
+    @classmethod
+    def get_import_state(task_id):
+        req = milvus_types.GetImportStateRequest(task_id)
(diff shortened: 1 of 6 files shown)


pymilvus/client/prepare.py
```diff
@@ -880,3 +880,23 @@ def get_replicas(cls, collection_id: int):
...
+        req = milvus_types.ImportRequest(
+            collection_name=collection_name,
+            partition_name=partition_name,
+            row_based=is_row_based,
+            files=files,
+        )
+
+        for k, v in kwargs.items():
+            if k in ("bucket",):
+                kv_pair = common_types.KeyValuePair(key=str(k), value=str(v))
```
DragonDriver, line 895: `if` here should only be hitted once, maybe better to break immediately.

pymilvus/client/stub.py
```diff
@@ -1068,3 +1068,43 @@ def get_replicas(self, collection_name: str, timeout=None, **kwargs) -> Replica:
...
+        :param collection_name: the name of the collection
+        :type  collection_name: str
+
+        :param partition_name: the name of the partition
+        :type  partition_name: str
+
+        :param is_row_based: indicate whether the files are row-based or coloumn based.
+        :type  is_row_based: bool
+
+        :param files: file names to bulk load
```
DragonDriver, line 1084: Maybe we should point out that the file is located in `Milvus Storage`, not in local disk.

pymilvus/client/stub.py
```diff
@@ -1068,3 +1068,43 @@ def get_replicas(self, collection_name: str, timeout=None, **kwargs) -> Replica:
...
+        :type  files: list[str]
+
+        :param timeout: The timeout for this method, unit: second
+        :type  timeout: int
+
+        :param kwargs: other infos
+
+        :return: ids of tasks
```
DragonDriver, line 1092: What does the `ids` mean? Every file corresponds to an id?

PR #961, opened 2022-04-22
Support create index for scalar field

xige-16:
Signed-off-by: xige-16 <xi.ge@zilliz.com>


xige-16:

PR #967, opened 2022-04-28
Support tls for self-signed certificate

xiyichan:
Signed-off-by: xiyichan <2863768433@qq.com>


DragonDriver:
Hello, @xiyichan , welcome to che community and thx for the contribution. Lgtm for the code changes generally. Besides, can you help to complete the documentation of the newly-added parameter. You can find the documentation in `connections.connect`. Btw, maybe the title of this patch should be more detailed, since tls was already supported. How about "Support to configure credential for tls"?

xiyichan:
> Lgtm
If it is a self-signed certificate, we need to specify the certificate when we use it.

pymilvus/orm/connections.py
```diff
@@ -140,6 +140,10 @@ def connect(self, alias=DefaultConfig.DEFAULT_USING, user="", password="", **kwa
             * *password* (``str``) --
                 Optional and required when user is provided. The password corresponding to
                 the user.
+            * *secure* (``bool``) --
+                Optional. Default is false. If set to true, tls will be enabled.
+            * *cert_path* (``str``) --
+                Optional. If use a self-signed certificate, need to write the server.crt path.
```
aakejiang, line 146: Will it need to specify ca.crt if CA exists?

xiyichan: I run the example.py is not need to specify ca.crt.

aakejiang: You should also test the case with CA.

xiyichan: i should build a ca server? I don't know how to test the case with CA.

pymilvus/client/grpc_handler.py
```diff
@@ -123,14 +127,26 @@ def _setup_grpc_channel(self):
                              ('grpc.keepalive_time_ms', 55000)]
                 )
             else:
-                creds = grpc.ssl_channel_credentials(root_certificates=None, private_key=None, certificate_chain=None)
+                if self._client_pem_path != "" and self._client_key_path != "" and self._ca_pem_path != "" \
```
DragonDriver, line 130: Should these four parameters appear at the same time?

pymilvus/client/grpc_handler.py
```diff
@@ -123,14 +127,26 @@ def _setup_grpc_channel(self):
...
+                        root_certificates = f.read()
+                    creds = grpc.ssl_channel_credentials(root_certificates, private_key, certificate_chain)
+                else:
+                    creds = grpc.ssl_channel_credentials(root_certificates=None, private_key=None,
+                                                         certificate_chain=None)
                 self._channel = grpc.secure_channel(
                     self._uri,
                     creds,
-                    options=[(cygrpc.ChannelArgKey.max_send_message_length, -1),
+                    options=(('grpc.ssl_target_name_override', self._server_name, ),
```
DragonDriver, line 145: What if `server_name` here is empty?

xiyichan: server_name can't be empty

pymilvus/client/grpc_handler.py
xiyichan: Yes, I forget add server_name .

PR #969, opened 2022-04-29
Fix err msg when max_length_per_row is too large

DragonDriver:
issue: https://github.com/milvus-io/milvus/issues/16584


xige-16:

PR #1004, opened 2022-05-18
support one-way authentication

xiyichan:
https://github.com/milvus-io/milvus/issues/17073
Signed-off-by: xiyichan <2863768433@qq.com>


pymilvus/client/grpc_handler.py
```diff
@@ -86,6 +86,7 @@ def __init__(self, uri=config.GRPC_ADDRESS, host=None, port=None, channel=None,
         self._client_pem_path = kwargs.get("client_pem_path", "")
         self._client_key_path = kwargs.get("client_key_path", "")
         self._ca_pem_path = kwargs.get("ca_pem_path", "")
+        self._server_pem_path = kwargs.get("server_pem_path", "")
         self._server_name = kwargs.get("server_name", "")
```
longjiquan, line 90: I wonder if we can just take these variables as temporary instead of saving them? Maybe it's not safe to save them.

pymilvus/client/grpc_handler.py
```diff
@@ -142,7 +143,11 @@ def _setup_grpc_channel(self):
                     with open(self._ca_pem_path, 'rb') as f:
                         root_certificates = f.read()
                     creds = grpc.ssl_channel_credentials(root_certificates, private_key, certificate_chain)
-
+                elif self._server_pem_path != "" and self._server_name != "":
```
longjiquan, line 146: OK, I got that we don't need another parameter to distinguish one-way tls or two-way tls. If `server_pem_path` and `server_name` were passed, then we should use two-way tls.

longjiquan:
/lgtm
/approve
@xiyichan thx for the contribution.

pymilvus/client/grpc_handler.py
xiyichan: Just read from kwargs.get ?

PR #1014, opened 2022-05-23
Set guarantee_timestamp to 2 in case of Bounded consistency level

### LINKED ISSUE #1005: [FEATURE]: Set guarantee_timestamp to 2 in case of Bounded consistency level

### Is your feature request related to a problem? Please describe.

Currently for Bound consistency, the client takes the client's time minus the default configured 5 seconds as the gurarantee_timestamp.
When the client time and server time are very different, this leads to the failure of Bound consistency.  see [ #16841](https://github.com/milvus-io/milvus/issues/16841)

### Describe the solution you'd like

I would like to change the strategy to:
For Bounded Consistency, client-side set guarantee_timestamp to `2`.
On the server side, set the timestamp to the arrival time of the proxy minus the default configured 5 second interval if recevied guarantee_timestamp is `2`.

Currently we have made some special settings for guarantee_timestamp.
When guarantee_timestamp is 0, the proxy will change it to the time when the request arrives at the server.
When the guarantee_timestamp is 1, the proxy does not change it. Since the service time of querynode  will be greater than 1, it is the effect of Eventually consistency.

Now, we add a new special setting for guarantee_timestmap. 
if guarantee_timestamp is `2`, proxy will change it to the arrival time of the proxy minus the default configured 5 seconds.

The settings of guarantee_timestamp and travel_timestamp will affect the effect of search grouping, so this logic needs to be clarified.

### Describe alternatives you've considered

_No response_

### Anything else?

_No response_

czs007:
fix #1005 
Signed-off-by: zhenshan.cao <zhenshan.cao@zilliz.com>

### DIFF
--- a/pymilvus/client/constants.py
+++ b/pymilvus/client/constants.py
@@ -2,6 +2,6 @@ from ..grpc_gen.common_pb2 import ConsistencyLevel
 
 LOGICAL_BITS = 18
 LOGICAL_BITS_MASK = (1 << LOGICAL_BITS) - 1
-DEFAULT_GRACEFUL_TIME = 5000  # in ms
 EVENTUALLY_TS = 1
+BOUNDED_TS = 2
 DEFAULT_CONSISTENCY_LEVEL = ConsistencyLevel.Bounded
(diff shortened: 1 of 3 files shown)


czs007:
see alse : https://github.com/milvus-io/milvus/issues/17087

PR #1063, opened 2022-07-07
Upgrade grpc version to 1.47.0

jaime0815:
Signed-off-by: yun.zhang <yun.zhang@zilliz.com>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
 build==0.4.0
 certifi==2021.5.30
 chardet==4.0.0
-grpcio>=1.37.1
-grpcio-testing>=1.37.1
-grpcio-tools>=1.37.1,<=1.43.0
+grpcio==1.47.0
+grpcio-testing==1.47.0
+grpcio-tools==1.47.0
 idna==2.10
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,9 +1,9 @@
 build==0.4.0
 certifi==2021.5.30
 chardet==4.0.0
-grpcio>=1.37.1
-grpcio-testing>=1.37.1
-grpcio-tools>=1.37.1,<=1.43.0
+grpcio==1.47.0
+grpcio-testing==1.47.0
+grpcio-tools==1.47.0
```
longjiquan, line 6: Please also change these versions in `setup.py`.

longjiquan: By the way, any suggestions why did we need this?

longjiquan (changes requested):

requirements.txt
jaime0815: > By the way, any suggestions why did we need this?

https://github.com/grpc/grpc/issues/24069 https://github.com/grpc/grpc/pull/29050

Issue #1115, opened 2022-08-23
[Enhancement]: add tracing interfaces


Zach41:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

add interfaces to allow user to pass:
- `log level` variables to change output logs
- `client trace id` to trace requests

PR #1116, opened 2022-08-23
add client log level and client quest id for rpc calls

Zach41:
ISSUE #1115

### DIFF
--- a/pymilvus/decorators.py
+++ b/pymilvus/decorators.py
@@ -26,6 +26,7 @@ def retry_on_rpc_failure(retry_times=10, initial_back_off=0.01, max_back_off=60,
     def wrapper(func):
         @functools.wraps(func)
         @error_handler(func_name=func.__name__)
+        @tracing_request()
         def handler(self, *args, **kwargs):
             # This has to make sure every timeout parameter is passing throught kwargs form as `timeout=10`
             _timeout = kwargs.get("timeout", None)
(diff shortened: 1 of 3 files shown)


pymilvus/decorators.py
```diff
@@ -26,6 +26,7 @@ def retry_on_rpc_failure(retry_times=10, initial_back_off=0.01, max_back_off=60,
     def wrapper(func):
         @functools.wraps(func)
         @error_handler(func_name=func.__name__)
+        @tracing_request()
```
longjiquan, line 29: I think it's not reasonable to combine retry decorator and tracing decorator.

pymilvus/decorators.py
```diff
@@ -108,3 +109,18 @@ def handler(*args, **kwargs):
                 raise e
         return handler
     return wrapper
+
+def tracing_request():
+    def wrapper(func):
+        @functools.wraps(func)
+        def handler(self, *args, **kwargs):
+            level = kwargs.get("log_level", None)
+            req_id = kwargs.get("client_request_id", None)
```
longjiquan, line 118: Three question:
- 1, what's the type of `client_request_id`?
- 2, how should users construct the `client_request_id`?
- 3, what's the behavior if two same `client_request_id`s were met in server?

Zach41: 1. client_request_id is a user defined string to trace request
2. if two same client_request_id are same, the corresponding requests have same trace id in their logs. It's recommended to use random string as the request id in order to avoid this problem.

pymilvus/decorators.py
longjiquan: Just keep this way. We can adjust the decorators later if necessary.

pymilvus/decorators.py
longjiquan: OK, got it. Better to give a simple documentation to guide users how to use this feature when everything is ready.

PR #1164, opened 2022-09-28
Fix partial orm api missing kwargs

bigsheeper:
Signed-off-by: bigsheeper <yihao.dai@zilliz.com>

### DIFF
--- a/pymilvus/orm/collection.py
+++ b/pymilvus/orm/collection.py
@@ -112,9 +112,9 @@ class Collection:
         self._kwargs = kwargs
         conn = self._get_connection()
 
-        has = conn.has_collection(self._name)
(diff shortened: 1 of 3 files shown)


pymilvus/orm/collection.py
```diff
@@ -272,7 +272,7 @@ def aliases(self) -> list:
             Aliases of the collection.
         """
         conn = self._get_connection()
-        resp = conn.describe_collection(self._name)
+        resp = conn.describe_collection(self._name, **self._kwargs)
```
longjiquan, line 275: Do not use `self._kwargs` here. We'd better to customize parameters for every dependent APIs.

longjiquan (changes requested):

Issue #1196, opened 2022-10-20
[Bug]: Failed to connect to milvus with secure, user and password if adding connections first


yanliang567:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

1. add connections with alias, host and port
2. connect with alias, secure, user and password
3.  fail to connect

BTW, if connect directly with alias, host, port, secure, user and password, it succeeds

### Expected Behavior

connect to milvus in both ways

### Steps/Code To Reproduce behavior

```markdown
>>> from pymilvus import utility, connections, DataType, \
...     Collection, FieldSchema, CollectionSchema
>>> connections.connect(alias="default11", host="xxx.com", port=19541, user="root", password="12345", secure=True)
>>> 
>>> 
>>> connections.add_connection(default={"host": "xxx.com", "port": 19541})
>>> connections.connect("default", user="root", password="12345", secure=True)
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/pymilvus/client/grpc_handler.py", line 114, in _wait_for_channel_ready
    grpc.channel_ready_future(self._channel).result(timeout=3)
  File "/usr/local/lib/python3.8/dist-packages/grpc/_utilities.py", line 139, in result
    self._block(timeout)
  File "/usr/local/lib/python3.8/dist-packages/grpc/_utilities.py", line 85, in _block
    raise grpc.FutureTimeoutError()
grpc.FutureTimeoutError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/pymilvus/orm/connections.py", line 306, in connect
    connect_milvus(**self._alias[alias], password=password)
  File "/usr/local/lib/python3.8/dist-packages/pymilvus/orm/connections.py", line 271, in connect_milvus
    gh._wait_for_channel_ready()
  File "/usr/local/lib/python3.8/dist-packages/pymilvus/client/grpc_handler.py", line 117, in _wait_for_channel_ready
    raise MilvusException(Status.CONNECT_FAILED, f'Fail connecting to server on {self._address}. Timeout')
pymilvus.exceptions.MilvusException: <MilvusException: (code=2, message=Fail connecting to server on in01-909c24c01dc99e9.aws-us-west-2.vectordb-uat.zillizcloud.com:19541. Timeout)>
```

### Environment details

```markdown
pymilvus              2.2.0.dev42
```

[...]

PR #1203, opened 2022-10-21
Fix unable to connect to secure server after add

XuanYang-cn:
See also: #1196


XuanYang-cn:
截屏2022-10-24 17 44 10
/unhold

PR #1210, opened 2022-10-26
Update index interface

xiaocai2333:
Signed-off-by: cai.zhang <cai.zhang@zilliz.com>
/kind improvement


pymilvus/exceptions.py
```diff
@@ -68,6 +68,10 @@ class IndexNotExistException(MilvusException):
     """ Raise when index doesn't exist """
 
 
+class MultipleIndexNameNotSpecify(MilvusException):
```
czs007, line 71: Better to rename as IndexNameNotSpecify

PR #1223, opened 2022-11-15
Optimize some low efficient code

liliu-z:
Signed-off-by: Li Liu <li.liu@zilliz.com>

This PR fix some low efficient code:
1. Call `to_dict()` of a collection at construction time and store it for search use.
2. Use a map instead of endless `if else` to check param.
3. For sync search, simply use serial execution instead of `async + get`
4. Remove deep copy in search path

For performance enhancement, I did some test in my 2C8M container:
*Pymilvus*: Set `grpc` to return empty response immediately, the QPS of `Pymilvus` increase from ~1200 to ~2100, increased about 75%
*End 2 End*: For nq1 serial execution, latency improved ~10%.

Notice: Still lots of similar style in other parts of the code, like deep copy, `to_dict()` call, `if else` param check and mixed sync/async call. We need to fix them all if needed.

### DIFF
--- a/pymilvus/client/check.py
+++ b/pymilvus/client/check.py
@@ -3,6 +3,7 @@ import datetime
 from typing import Any, Union
 from ..exceptions import ParamError
 from ..grpc_gen import milvus_pb2 as milvus_types
+from .singleton_utils import Singleton
 from .utils import (
     valid_index_types,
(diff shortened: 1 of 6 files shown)


XuanYang-cn:
Marvelous!
/lgtm
/approve

PR #1273, opened 2023-01-31
Asyncio support

belkka:
Commits with commit message "draft" will be overridden, so ignore formatting/style issues in those commits.

### DIFF
--- a/pymilvus/orm/connections.py
+++ b/pymilvus/orm/connections.py
@@ -14,11 +14,12 @@ import os
 import copy
 import re
 import threading
+import typing
 from urllib import parse
 from typing import Tuple
 
 from ..client.check import is_legal_host, is_legal_port, is_legal_address
-from ..client.grpc_handler import GrpcHandler
+from ..client.grpc_handler import GrpcHandler, AbstractGrpcHandler
 
 from .default_config import DefaultConfig, ENV_CONNECTION_CONF
 from ..exceptions import ExceptionsMessage, ConnectionConfigException, ConnectionNotExistException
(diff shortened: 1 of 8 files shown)


belkka:
Started moving reusable parts of `GrpcHandler` into `AbstractGrpcHandler`.

pymilvus/orm/connections.py
```diff
@@ -211,8 +218,16 @@ def remove_connection(self, alias: str):
         if not isinstance(alias, str):
             raise ConnectionConfigException(message=ExceptionsMessage.AliasType % type(alias))
 
-        self.disconnect(alias)
+        # TODO: does order matter?
+        # original sync implementation was
+        #     self.disconnect(alias)
+        #     self._alias.pop(alias, None)
```
Andrew-Chen-Wang, line 224: there is a possibility that it could be because a disconnect was unsuccessful which raised an error such that the alias wasn't popped in that errored event. But not sure.

filip-halt: I assume its as you have stated, make sure disconnected before removing so that we are sure there are no ghost connections left over.

belkka: Could you describe an example scenario?

Current implementation of `.disconnect()` actually removes a GrpcHandler first and *then* calls `.close()`:
https://github.com/milvus-io/pymilvus/blob/94c210c7f0385d0d4c76190b74cde40a491fc966/pymilvus/orm/connections.py#L203

In case that some exception is raised from `GrpcHandler.close()`, the grpc handler instance will be "lost" anyway (removed from `._connected_alias` dict). Is it meaningful to keep a corresponding value in `self._alias` in that case (instead of removing as requested by user)?

Anyway, I've updated the code to keep original statements order. But it may turn out that we actually must use some asyncio synchronization primitives for reliable behavior in case of errors.

mistborn17:
Just to followup, I'm interested in the asyncio support for pymilvus. Is this PR still relevant ?

belkka:
@mistborn17 I stopped working on this PR when suddenly noticed https://github.com/milvus-io/pymilvus/pull/1288/files, which adds "deprecation" markings on code that is being "adapted" for asyncio in this PR (e. g. `pymilvus.orm.collection.Collection`). I had some communication on this topic in slack (milvusio.slack.com), but now it does not let me read messages older than 90 days. IIRC, they said they are working on new API (in a separate branch) and they are gonna provide asyncio interface in that API (but maybe not full support/test coverage).

czs007:
refer to https://github.com/milvus-io/pymilvus/pull/2411

Issue #1324, opened 2023-03-07
[Bug]: Failed to insert data to Milvus: 'int' object has no attribute 'name'


tom729:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

I reuse the reverse_image_search example to build a image similar search engine, I have succeed to insert about 50w image to the milvus, however, today I insert new data to milvus failed, I am not sure wether I did some configuraiton changes or some code changes. how can I find more useful info about the error, is there any other guy have the same problem?
the below is the insert function, which i rewrite to meet our needs.

### Steps/Code To Reproduce behavior

> ``def insert(self, collection_name, vectors,work_list,path_list,uid_list):
        # Batch insert vectors to milvus collection
        try:
            self.create_collection(collection_name)
            self.set_collection(collection_name)
            #LOGGER.debug(self.collection.schema)
            data = [vectors,work_list,path_list,uid_list]
            mr = self.collection.insert(data)
            LOGGER.debug(f"Insert vectors to Milvus in collection: {collection_name} with {len(vectors)} rows")
            ids = mr.primary_keys
            #self.collection.flush() 
            self.collection.load()
            LOGGER.debug("ids %s" %ids)
            return ids
        except Exception as e:
            LOGGER.error(f"Failed to insert data to Milvus: {e}")
            return(-1)``

_No response_

### Environment details

```markdown
- Hardware/Softward conditions (ubuntu, tencent cloud 4 CORE 16G memory):
- Method of installation (Docker):
- Milvus version (2.2.2)
- pymilvus version (2.2.2)
```

tom729:
create colleciont codes: 
`    def create_collection(self, collection_name):
        # Create milvus collection if not exists
        try:
            if not self.has_collection(collection_name):
                field1 = FieldSchema(name="id", dtype=DataType.INT64, descrition="int64", is_primary=True, auto_id=True)
                field2 = FieldSchema(name="embedding", dtype=DataType.FLOAT_VECTOR, descrition="float vector",
                                     dim=VECTOR_DIMENSION, is_primary=False)
                field3 = FieldSchema(name="work_id", dtype=DataType.INT64, descrition="work id", is_primary=False,auto_id=False)
                field4 = FieldSchema(name='path', dtype=DataType.VARCHAR, description='path to image', max_length=200, is_primary=False, auto_id=False)
[...]

PR #1333, opened 2023-03-23
Fix error message when insert mismatch datatypes

XuanYang-cn:
See also: #1324


longjiquan:
Good catch. @XuanYang-cn

Issue #1329, opened 2023-03-15
[FEATURE]: remove support for python3.6


XuanYang-cn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

According to [PEP 494 – Python 3.6 Release Schedule](https://peps.python.org/pep-0494/#lifespan)

> As of 2021-12-23, 3.6 has reached the [end-of-life phase](https://devguide.python.org/devcycle/#end-of-life-branches) of its release cycle. 3.6.15 was the final security release. The code base for 3.6 is now frozen and no further updates will be provided nor issues of any kind will be accepted on the bug tracker.

We'll stop supporting python3.6 on pymilvus from **pymilvus==2.3.0**

PR #1344, opened 2023-04-03
support latest grpcio

XuanYang-cn:
See also: #1329, 1308


pymilvus/client/constants.py
```diff
@@ -1,7 +1,7 @@
-from ..grpc_gen.common_pb2 import ConsistencyLevel
+from ..grpc_gen import common_pb2
 
 LOGICAL_BITS = 18
 LOGICAL_BITS_MASK = (1 << LOGICAL_BITS) - 1
 EVENTUALLY_TS = 1
 BOUNDED_TS = 2
-DEFAULT_CONSISTENCY_LEVEL = ConsistencyLevel.Bounded
```
XuanYang-cn, line 7: This change is due to `pylint` unable to identify the check imported from a `*.pyi` file.

longjiquan: Can we define `ConsistencyLevel = common_pb2.ConsistencyLevel`, so the codes depend on the `ConsistencyLevel` can keep the compatibility?

longjiquan (changes requested):

pymilvus/client/constants.py
longjiquan: In this way, it seems that the changes related to the `ConsistencyLevel` can be also saved.

XuanYang-cn: It's a enum defined by proto, won't be changed from sdk, we're just using its enum functions; But I edit it anyway to reduce some code changes.

longjiquan:
LGTM
/assign @czs007

PR #1374, opened 2023-04-20
Milvus Client

filip-halt:
Looking to add a wrapper around Pymilvus for easier usage.

### DIFF
--- a/pymilvus/milvus_client/defaults.py
+++ b/pymilvus/milvus_client/defaults.py
@@ -1,0 +1,12 @@
+"""Default MilvusClient args."""
+
+DEFAULT_SEARCH_PARAMS = {
+    "IVF_FLAT": {"metric_type": "L2", "params": {"nprobe": 10}},
+    "IVF_SQ8": {"metric_type": "L2", "params": {"nprobe": 10}},
+    "IVF_PQ": {"metric_type": "L2", "params": {"nprobe": 10}},
+    "HNSW": {"metric_type": "L2", "params": {"ef": 10}},
+    "RHNSW_FLAT": {"metric_type": "L2", "params": {"ef": 10}},
+    "RHNSW_SQ": {"metric_type": "L2", "params": {"ef": 10}},
+    "RHNSW_PQ": {"metric_type": "L2", "params": {"ef": 10}},
+    "AUTOINDEX": {"metric_type": "L2", "params": {}},
+}
--- a/pymilvus/milvus_client/milvus_client.py
+++ b/pymilvus/milvus_client/milvus_client.py
@@ -1,0 +1,869 @@
+"""MilvusClient for dealing with simple workflows."""
+import logging
+import threading
+from typing import Dict, List, Union
+from uuid import uuid4
+
+from tqdm import tqdm
+
+from pymilvus.exceptions import MilvusException
+from pymilvus.milvus_client.defaults import DEFAULT_SEARCH_PARAMS
+from pymilvus.orm import utility
+from pymilvus.orm.collection import Collection, CollectionSchema, FieldSchema
+from pymilvus.orm.connections import connections
+from pymilvus.orm.types import DataType, infer_dtype_bydata
+
+logger = logging.getLogger()
+logger.setLevel(logging.DEBUG)
+
+
+class MilvusClient:
+    """The Milvus Client"""
+
+    # pylint: disable=logging-too-many-args, too-many-instance-attributes
+
+    def __init__(
+        self,
+        collection_name: str = "ClientCollection",
+        pk_field: str = None,
+        vector_field: str = None,
+        uri: str = None,
+        shard_num: int = None,
+        partitions: List[str] = None,
+        consistency_level: str = "Bounded",
+        replica_number: int = 1,
+        index_params: dict = None,
+        timeout: int = None,
+        overwrite: bool = False,
+    ):
+        """A client for the common Milvus use case.
+
+        This client attempts to hide away the complexity of using Pymilvus. In a lot ofcases what
+        the user wants is a simple wrapper that supports adding data, deleting data, and searching.
+        This wrapper can autoinfer the schema from a previous collection or newly inserted data,
+        can update the paritions, can query, and can delete by pk.
+
+        Args:
+            pk_field (str, optional): Which entry in data is considered the primary key. If None,
+                an auto-id will be created. Will be overwritten if loading from a previous
+                collection. Defaults to None.
+            vector_field (str, optional): Which entry in the data is considered the vector field.
+                Will get overwritten if loading from previous collection. Required if not using
+                already made collection.
+            uri (str, optional): The connection address to use to connect to the
+                instance. Defaults to "http://localhost:19530". Another example:
+                "https://username:password@in01-12a.aws-us-west-2.vectordb.zillizcloud.com:19538
+            shard_num (int, optional): The amount of shards to use for the collection. Unless
+                dealing with huge scale, recommended to keep at default. Defaults to None and allows
+                server to set.
+            partitions (List[str], optional): Which paritions to create for the collection.
+                Defaults to None.
+            consistency_level (str, optional): Which consistency level to use for the Client.
+                The options are "Strong", "Bounded", "Eventually", "Session". Defaults to "Bounded".
+            replica_number (int, optional): The amount of in memomory replicas to use.
+                Defaults to 1.
+            index_params (dict, optional): What index parameteres to use for the Collection.
+                If none, will use a default one. If collection already exists, will overwrite
+                using this index.
+            timeout (int, optional): What timeout to use for function calls. Defaults
+                to None.
+            overwrite (bool, optional): Whether to overwrite existing collection if exists.
+                Defaults to False
+        """
+        self.uri = uri
+        self.collection_name = collection_name
+        self.shard_num = shard_num
+        self.partitions = partitions
+        self.consistency_level = consistency_level
+        self.replica_number = replica_number
+        self.index_params = index_params
+        self.timeout = timeout
+        self.pk_field = pk_field
+        self.vector_field = vector_field
+
+        # TODO: Figure out thread safety
+        # self.concurrent_counter = 0
+        self.concurrent_lock = threading.RLock()
+        self.default_search_params = None
+        self.collection = None
+        self.fields = None
+
+        self.alias = self._create_connection()
+        self.is_self_hosted = bool(
+            utility.get_server_type(using=self.alias) == "milvus"
+        )
+        if overwrite and utility.has_collection(self.collection_name, using=self.alias):
+            utility.drop_collection(self.collection_name, using=self.alias)
+
+        self._init(None)
+
+    def __len__(self):
+        return self.num_entities()
+
+    def num_entities(self):
+        """return the number of rows in the collection.
+
+        Returns:
+            int: Number for rows.
+        """
+        if self.collection is None:
+            return 0
+
+        self.collection.flush()
+        return self.collection.num_entities
+
+    def insert_data(
+        self,
+        data: List[Dict[str, any]],
+        timeout: int = None,
+        batch_size: int = 100,
+        partition: str = None,
+        progress_bar: bool = False,
+    ) -> List[Union[str, int]]:
+        """Insert data into the collection.
+
+        If the Milvus Client was initiated without an existing Collection, the first dict passed
+        in will be used to initiate the collection.
+
+        Args:
+            data (List[Dict[str, any]]): A list of dicts to pass in. If list not provided, will
+                cast to list.
+            timeout (int, optional): The timeout to use, will override init timeout. Defaults
+                to None.
+            batch_size (int, optional): The batch size to perform inputs with. Defaults to 100.
+            partition (str, optional): Which partition to insert into. Defaults to None.
+            progress_bar (bool, optional): Whether to display a progress bar for the input.
+                Defaults to False.
+
+        Raises:
+            DataNotMatchException: If the data has misssing fields an exception will be thrown.
+            MilvusException: General Milvus error on insert.
+
+        Returns:
+            List[Union[str, int]]: A list of primary keys that were inserted.
+        """
+        # If no data provided, we cannot input anything
+        if len(data) == 0:
+            return []
+
+        if batch_size < 1:
+            logger.error("Invalid batch size provided for insert.")
+
+            raise ValueError("Invalid batch size provided for insert.")
+
+        # If the collection hasnt been initialized, initialize it
+        with self.concurrent_lock:
+            if self.collection is None:
+                self._init(data[0])
+
+        # Dont include the primary key if auto_id is true and they included it in data
+        ignore_pk = self.pk_field if self.collection.schema.auto_id else None
+        insert_dict = {}
+        pks = []
+
+        for k in data:
+            for key, value in k.items():
+                if key in self.fields:
+                    insert_dict.setdefault(key, []).append(value)
+
+        # Insert the data in batches
+        for i in tqdm(range(0, len(data), batch_size), disable=not progress_bar):
+            # Convert dict to list of lists batch for insertion
+            try:
+                insert_batch = [
+                    insert_dict[key][i : i + batch_size]
+                    for key in self.fields
+                    if key != ignore_pk
+                ]
+            except KeyError as ex:
+                logger.error(
+                    "Malformed data, at least one of the inserts does not contain all"
+                    " the required fields."
+                )
+                raise KeyError(
+                    f"Malformed data, at least one of the inserts does not"
+                    f" the required fields: {ex}",
+                ) from ex
+            # Insert into the collection.
+            try:
+                res = self.collection.insert(
+                    insert_batch,
+                    timeout=timeout or self.timeout,
+                    partition_name=partition,
+                )
+                pks.extend(res.primary_keys)
+            except MilvusException as ex:
+                logger.error(
+                    "Failed to insert batch starting at entity: %s/%s",
+                    str(i),
+                    str(len(data)),
+                )
+                raise ex
+        return pks
+
+    def upsert_data(
+        self,
+        data: List[Dict[str, any]],
+        timeout: int = None,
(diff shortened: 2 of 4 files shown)


filip-halt:
fixes 1380

xiaofan-luan:
please squash your commit record into 1 to make the commit history easier to understand.

pymilvus/milvus_client/defaults.py
```diff
@@ -0,0 +1,26 @@
...
+DEFAULT_LOAD_ARGS = {
+    "partition": None,
+    "replica_number": 1,
+}
+
+DEFAULT_CONNECTION_ARGS = {
+    "alias": "default",
+}
+
+DEFAULT_SEARCH_PARAMS = {
```
xiaofan-luan, line 15: we should change all the default metric type to cosine in the future.

pymilvus/milvus_client/defaults.py
```diff
@@ -0,0 +1,26 @@
...
+    "IVF_FLAT": {"metric_type": "L2", "params": {"nprobe": 10}},
+    "IVF_SQ8": {"metric_type": "L2", "params": {"nprobe": 10}},
+    "IVF_PQ": {"metric_type": "L2", "params": {"nprobe": 10}},
+    "HNSW": {"metric_type": "L2", "params": {"ef": 10}},
+    "RHNSW_FLAT": {"metric_type": "L2", "params": {"ef": 10}},
+    "RHNSW_SQ": {"metric_type": "L2", "params": {"ef": 10}},
+    "RHNSW_PQ": {"metric_type": "L2", "params": {"ef": 10}},
+    "IVF_HNSW": {"metric_type": "L2", "params": {"nprobe": 10, "ef": 10}},
+    "ANNOY": {"metric_type": "L2", "params": {"search_k": 10}},
```
xiaofan-luan, line 24: remove ANNOY, and probably IVF_HNSW

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+
+class MilvusClient:
+    """The Milvus Client"""
+    def __init__(
+        self,
+        collection_name: str = "ClientCollection",
+        pk_field: str = None,
+        vector_field: str = None,
+        uri: str = None,
+        shard_num: int = 2,
```
xiaofan-luan, line 30: leave the shard_num to be empty, server side has default value of shard number

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+        partitions: List[str] = None,
+        consistency_level: str = "Session",
```
xiaofan-luan, line 32: use bounded, it's actually much faster under most cases

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+            self._create_collection(input_data)
+        # Nothin to init from
+        else:
+            logger.debug(
+                "No information to perform init from for collection %s",
+                self.collection_name,
+            )
+            return
+        self._create_index()
+        # Partitions only allowed on Milvus at the mmoment
```
xiaofan-luan, line 569: typo: moment

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+                If none, will use a default one. Defaults to None.
+            timeout (Optional[int], optional): What timeout to use for function calls. Defaults
+                to None.
+            drop_old (bool, optional): If a collection with the same name already exists, drop it.
+                Defaults to False.
+        """
+        self.uri = uri
+        self.collection_name = collection_name
+        self.shard_num = shard_num
+        self.partitions = partitions
```
xiaofan-luan, line 72: we should just ignore the partition here.

Later on user can specify the partition key here

xiaofan-luan: Milvus will automatially create 64 or 256 partition for the user

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+            self.vector_field,
+        )
+
+    def _create_index(self) -> None:
+        """Create a index on the collection"""
+        if self._get_index() is None:
+            # If no index params, use a default HNSW based one
+            if self.index_params is None:
+                # TODO: Once segment normalization we can default to IP
+                metric_type = (
```
xiaofan-luan, line 699: user has to be able to set default metric type I guess

xiaofan-luan: we are gonna to support cosine soon,so far the default set to L2 works for me but it has to been changed soon.

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+        pk_field: str = None,
+        vector_field: str = None,
+        uri: str = None,
+        shard_num: int = 2,
+        partitions: List[str] = None,
+        consistency_level: str = "Session",
+        replica_number: int = 1,
+        index_params: dict = None,
+        timeout: Optional[int] = None,
+        drop_old: bool = False,
```
xiaofan-luan, line 36: drop may not be a good idea on init, especially on multi client situation.
I would rather let the user to drop the collection on close if they'd like to

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+            )
+            return
+        self._create_index()
+        # Partitions only allowed on Milvus at the mmoment
+        if self.is_self_hosted and self.partitions is not None:
+            self.add_partitions(self.partitions)
+        self._create_default_search_params()
+        self._load()
+
+    def _create_collection(self, data: dict) -> None:
```
xiaofan-luan, line 575: We gonna to simply this part whenever the dynamic schema is ready. This is awful

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+        if self._get_index() is not None:
+            # Check if the collection is loaded or in progress of loading
+            if (
+                utility.load_state(
+                    self.collection_name, using=self.alias, timeout=self.timeout
+                )
+                != LoadState.NotLoad
+            ):
+                # IF the collection is loaded/loading, check the replica count
+                if len(self.collection.get_replicas().groups) == self.replica_number:
```
xiaofan-luan, line 761: pls open a issue in Milvus.
We have to make load idempotent.
It should dynamic add or remove one replica when ever you call load

xiaofan-luan: I though there is a refresh parameter which did this, pleas double check with Alex @soothing-rain

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+from pymilvus.orm import utility
+from pymilvus.orm.collection import Collection, CollectionSchema, FieldSchema
+from pymilvus.orm.connections import connections
+from pymilvus.orm.types import DataType, infer_dtype_bydata
+
+logger = logging.getLogger()
+logger.setLevel(logging.DEBUG)
+
+
+class MilvusClient:
```
xiaofan-luan, line 22: I think you will need a add a close method to:
close the connection
cleanup all the collections, index and release the data if necessary.

pymilvus/milvus_client/milvus_client.py
xiaofan-luan: I love this code BTW, this is the best we can do right now.
Just hate the inference part because it is error prone

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+            for field_dict in fields.values():
+                fieldschemas.append(FieldSchema(**field_dict))
+                self.fields[field_dict["name"]] = field_dict["dtype"]
+            # Create the schema for the collection
+            schema = CollectionSchema(fieldschemas)
+            # Create the collection
+            self.collection = Collection(
+                name=self.collection_name,
+                schema=schema,
+                consistency_level=self.consistency_level,
```
xiaofan-luan, line 661: ignore consistency and shard thing

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+        pks = []
+
+
+        for k in data:
+            for key, value in k.items():
+                if key in self.fields:
+                    insert_dict.setdefault(key, []).append(value)
+
+        # Insert the data in batches
+        for i in tqdm(range(0, len(data), batch_size), disable=not progress_bar):
```
xiaofan-luan, line 147: I love the batch design but we need to test the corner case.
For ex. batch_size = 0

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+                )
+                pks.extend(res.primary_keys)
+            except MilvusException as ex:
+                logger.error(
+                    "Failed to insert batch starting at entity: %s/%s", str(i), str(len(data))
+                )
+                raise ex
+        return pks
+
+    def upsert_data(
```
xiaofan-luan, line 171: change to upsert implementation in the future.
comment it out the performance might be really bad due to extra delete

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+            logger.error("Collection does not exist: %s", self.collection_name)
+            raise ValueError(
+                "Missing collection. Make sure data inserted or intialized on existing collection."
+            )
+
+        if not isinstance(data[0], list):
+            data = [data]
+
+        return_fields = list(self.fields.keys())
+        return_fields.remove(self.vector_field)
```
xiaofan-luan, line 258: gonna to support return fields soon

xiaofan-luan: for now if user has output field requirement for vector field, I guess we need to do a get_embedding_by_PK here.

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+        #     self.concurrent_counter += 1
+
+        if self.collection is None:
+            logger.error("Collection does not exist: %s", self.collection_name)
+            raise ValueError(
+                "Missing collection. Make sure data inserted or intialized on existing collection."
+            )
+
+        return_fields = list(self.fields.keys())
+        return_fields.remove(self.vector_field)
```
xiaofan-luan, line 314: query support vector field, is there a specific reason we remove it?

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+
+        if self.vector_field is None:
+            logger.error(
+                "Missing vector_field, cannot infer schema for collection: %s",
+                self.collection_name,
+            )
+            raise ValueError("Missing vector_field, cannot autoinfer schema.")
+
+        try:
+            self.dim = len(data[self.vector_field])
```
longjiquan, line 611: The dimension calculation is correct here for float vectors, but wrong for binary vectors. In fact, we already got the field type by `infer_dtype_bydata`, which can help us to determine the dimension calculation more clear.

longjiquan: Btw, `self.vector_field_type` should be also added. Or should we forbid the binary vector type?

pymilvus/milvus_client/milvus_client.py
longjiquan: agree, atomicity can't be promised if we implement the `upsert` on the client side.

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+        if self.collection is None:
+            logger.error("Collection does not exist: %s", self.collection_name)
+            raise ValueError(
+                "Missing collection. Make sure data inserted or intialized on existing collection."
+            )
+
+        if not isinstance(data[0], list):
+            data = [data]
+
+        return_fields = list(self.fields.keys())
```
longjiquan, line 257: quick question: can the `return_fields` be choosen by the users? If the users didn't specify it, then we return all fields by default.

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+            partition_names=partitions,
+            output_fields=return_fields,
+            timeout=timeout or self.timeout,
+        )
+
+        ret = []
+        for hits in res:
+            for hit in hits:
+                ret_dict = {x: hit.entity.get(x) for x in return_fields}
+                ret.append({"score": hit.score, "data": ret_dict})
```
longjiquan, line 275: We should have an identifier here which can let users know which entity belongs to which query.
For example, if the query is [q1, q2] with top4, but the server returns [entity1, entity2, entity3, entity4], which entity1 and entity2 belongs to q1 and entity3 and entity4 belongs to q2, we can't handle this case here.

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+            timeout=timeout or self.timeout,
+        )
+
+        # TODO: Figure out thread safety
+        # with self.concurrent_lock:
+        #     self.concurrent_counter -= 1
+
+        return res
+
+    def get_embeddings_by_pk(
```
longjiquan, line 329: Great wrapper for the `get_embeddings_by_pk` and `delete_by_pk`. Without this, I noticed that there are many issues in the community related to how to get vectors by pk and delete by pk. I believe this will make these developers' life easier.

pymilvus/milvus_client/milvus_client.py
```diff
@@ -0,0 +1,801 @@
...
+                current_partitions = {
+                    partition.name for partition in self.collection.partitions
+                }
+                new_partitions = input_partitions.difference(current_partitions)
+                # If partitions need to be added, add them
+                if len(new_partitions) != 0:
+                    # Try to unload the collection, if exception already released
+                    reload = False
+                    try:
+                        self.collection.release(timeout=self.timeout)
```
longjiquan, line 454: Refactor this after https://github.com/milvus-io/milvus/pull/22655 has been released. Then there is no necessary to release the collection when the users tried to add a partition.

longjiquan (changes requested):

pymilvus/milvus_client/defaults.py
filip-halt: Yes, for now I was putting L2 as we cant enforce that the data is normalized which will create very poor results. If we end up normalizing at the segment level it may be good enough to default to.

pymilvus/milvus_client/milvus_client.py
filip-halt: They still need to name the partitions correct? Should I instead do the partition check on each insert?

pymilvus/milvus_client/milvus_client.py
filip-halt: The user is able to set an index at init if they want to use their own index params.

pymilvus/milvus_client/milvus_client.py
filip-halt: On close do you mean have a init arg drop_on_close and have the __delete__ also drop the collection? Or do we want dropping to only be called when the user calls it specifically. The reason I put this in so for the use case where you are doing a lot of testing and need a fresh collection each time.

pymilvus/milvus_client/milvus_client.py
filip-halt: We have this for refreshing. The description didnt sound like that logic but I could be wrong:
image

pymilvus/milvus_client/milvus_client.py
filip-halt: good catch

pymilvus/milvus_client/milvus_client.py
filip-halt: Yes, this is something that I want to replace when upsert is supported. For now most integrations and users need some kind of upsert and this is the route we are recommending, so this is just a temporary solution. Once 2.3 stable is released I will replace it. I will add a comment in docstring that emphasizes that this operation is slow and not guaranteed.

pymilvus/milvus_client/milvus_client.py
filip-halt: Ah I thought it only supports query based on pk.

pymilvus/milvus_client/milvus_client.py
xiaofan-luan: Agreed, by default should be only score and id, but user should be able to specify it

pymilvus/milvus_client/milvus_client.py
xiaofan-luan: we should support for binary vector, but by default we assume it's float vector

pymilvus/milvus_client/milvus_client.py
xiaofan-luan: I think if load is done. calling another load should just return success if collection is already loaded.

pymilvus/milvus_client/milvus_client.py
filip-halt: I think by default it might be better to do all fields. If a user connects to an instance and doesnt know the fields, they might want to grab all of them and parse. If they want to optimize they can say which specific ones.

pymilvus/milvus_client/milvus_client.py
filip-halt: I believe that it is idempotent, it was more for the dynamic replica count that I did this logic.

pymilvus/milvus_client/milvus_client.py
filip-halt: pymilvus checks for bytestrings so thats what I did here. If its binary vector the dim is 8 * the bystring length.

longjiquan:
@filip-halt could you please solve the code style issues, otherwise this can't be merged automatically.

pymilvus/milvus_client/milvus_client.py
filip-halt: ah, woops refactored too early.

Issue #1369, opened 2023-04-19
[FEATURE]: Support set username and password in system config


xiaofan-luan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

something like
user= os.getenv("MILVUS_USER")
password= os.getenv("MILVUS_PASSWORD")
so user don't need to hardcode their user and password somewhere

filip-halt:
In regards to this, do we have any plans/ideas for supporting a unified connection that takes username and password into account for milvus and zilliz? 


Something along the lines:
`connection_string = "username:password@localhost:1930"`
or
`connection_string = "localhost:19530"`
or 
`connection_string = "username:password@in01-random123.aws-us-west-2.vectordb.zillizcloud.com:19538"`

Some way to get away all the different options: address vs host/port vs uri/user/password/secure.

xiaofan-luan:
that actually make sense to me also.

something like that
username:password@in01-random123.aws-us-west-2.vectordb.zillizcloud.com:19538

XuanYang-cn:
Done

PR #1384, opened 2023-04-21
Add support for setting uri from environment

XuanYang-cn:
See also: #1369


czs007:
We need to actually test it to see if we can connect to zilliz cloud.

Issue #1361, opened 2023-04-14
[FEATURE]: Check if an index exists


filip-halt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

At the current moment there is not fast way to check if an index exists on the vector in a collection. 

In order to check if an index exists you first need to extract the embedding field name from the schema by iterating over all its fields looking for a dataype of FLOAT_VECTOR or BINARY_VECTOR. Then you must go through the indexes of the collection and check if the index.field_name matches the embedding field name. 

The other route is to try to load() and catch the exception but that seems a bit hacky. 

### Describe the solution you'd like

collection.vector_indexed() returns true or false based on having an index built on the vector field.

xiaofan-luan:
collection.has_index should be the API you are looking for

filip-halt:
> collection.has_index should be the API you are looking for

Unfortunately, just having an attribute index causes it to respond with True. I think it grabs the first index it can without checking if its on the embedding field.

xiaofan-luan:
> 

This one should check for vector field and I guess it only cares about vector index unless you pass in indexname.
@czs007 correct me if I'm worng

filip-halt:
```python
connections.connect(host='localhost', port=19530)
fields = [
    FieldSchema(name="id", dtype=DataType.INT64, is_primary=True, auto_id=True),
    FieldSchema(name="string1", dtype=DataType.VARCHAR, max_length=1000),
    FieldSchema(name="embedding", dtype=DataType.FLOAT_VECTOR, dim=DIMENSION),
]
schema = CollectionSchema(fields=fields)
if utility.has_collection("lol"):
    utility.drop_collection("lol")
collection = Collection("lol", schema, consistency_level="Strong")
index_params = {
    'metric_type':'L2',
    'index_type':"HNSW",
    'params':{'M': 8, 'efConstruction': 64}
}
collection.create_index('id')
print("has_index() call", collection.has_index())
try:
    collection.load()
except MilvusException as e:
    print(e.message)
```

will print:
```
has_index() call True
RPC error: [load_collection], <MilvusException: (code=1, message=there is no vector index on collection: lol, please create index firstly)>, <Time:{'RPC start': '2023-04-19 18:02:20.819621', 'RPC error': '2023-04-19 18:02:20.831111'}>
there is no vector index on collection: lol, please create index firstly
```

xiaofan-luan:
[...]

PR #1386, opened 2023-04-23
List indexes by field name

longjiquan:
issue: #1361


filip-halt:
I know this is closed but why is the field_name under kwargs? It might be better to expose it as a parameter.

Issue #1393, opened 2023-04-25
[Bug]: Incorrectly sized float vector has error message for binary vector


filip-halt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

<img width="1006" alt="image" src="https://user-images.githubusercontent.com/81822489/234362299-a76467e5-1f06-4b84-99dd-994e9181bd44.png">

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

PR #1394, opened 2023-04-25
Fix incorrect dim error msg

filip-halt:
Float and binary vector error message shared which results in incorrect message


filip-halt:
fixes https://github.com/milvus-io/pymilvus/issues/1393

PR #1407, opened 2023-04-28
Multi cherry picks from master

XuanYang-cn:
- Fix incorrect dim error msg (#1394)
- Include MilvusClient (#1374)


filip-halt:
If this code could be included in the cherrypick that would be awesome. I also have the release notes example included in the PR description.

PR #1412, opened 2023-05-05
Unify num_shards and mark shards_num as deprecated

XuanYang-cn:
The user can still use shards_num to create collection. But shards_num are the recommanded way to create collection and the only way to get from a collection's property

See also: milvus-io/milvus#23853


longjiquan (approved):
/lgtm

PR #1434, opened 2023-05-16
Add support of insert by rows

czs007:
issue #1454


pymilvus/orm/collection.py
```diff
@@ -430,15 +431,21 @@ def insert(self, data: Union[List, pandas.DataFrame], partition_name: str = None
...
-            return MutationFuture(res)
+        if not row_based:
+            check_insert_or_upsert_data_schema(self._schema, data)
+            entities = Prepare.prepare_insert_or_upsert_data(data, self._schema)
+            res = conn.batch_insert(self._name, entities, partition_name,
+                                    timeout=timeout, schema=self._schema_dict, **kwargs)
+            if kwargs.get("_async", False):
+                return MutationFuture(res)
+        else:
+            res = conn.insert_rows(self._name, data, partition_name,
```
filip-halt, line 447: Wouldn't it make sense to do the conversion that is found in Prepare. row_insert_or_upsert_param() here and instead of having both insert_batch and insert_rows()?

examples/dynamic_field.py
```diff
@@ -0,0 +1,95 @@
...
+
+rng = np.random.default_rng(seed=19530)
+vectors_to_search = rng.random((1, dim))
+search_params = {
+    "metric_type": "L2",
+    "params": {"nprobe": 10},
+}
+
+start_time = time.time()
+result = hello_milvus.search(vectors_to_search, "embeddings", search_params, limit=3, output_fields=["pk", "embeddings"])
```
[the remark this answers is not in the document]
czs007: If you wish to access a specific field within the meta field, simply include the field's name within the output_fields. If you desire to access the entire content of the meta field, include "$meta" within the output_fields.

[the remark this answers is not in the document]
czs007: The meaning of "$meta" will be specified within our documentation.

pymilvus/orm/collection.py
czs007: Yes, initially I had contemplated such an approach. However, upon closer inspection, I noticed that even after converting the data into the format required for batch_insert, batch_insert would still proceed to perform another conversion to generate the insertRequest. In light of performance considerations, I therefore opted to provide an insert_rows function which directly transforms into insertRequest.

czs007: In the future, if we decide to abandon batch_insert, it might be better to provide a unified entry point for inserts in milvus_client that can handle both row and column format.

PR #1436, opened 2023-05-17
Update only one field message

xige-16:


xige-16:
link https://github.com/milvus-io/milvus/issues/24149

xige-16:

PR #1437, opened 2023-05-17
Fix num_partitions type

xige-16:
issue: https://github.com/milvus-io/milvus/issues/24140

Signed-off-by: xige-16 <xi.ge@zilliz.com>


xige-16:

PR #1446, opened 2023-05-18
reuse grpc channel and clean connections

filip-halt:
This PR:
- adds in grpc reuse
- corrects user and db_name alias checks
- cleans up and reduces helper functions
- centralizes connection args logic
- enforces correct URI, does not attempt to attach default host and port to miss structed uri

### DIFF
--- a/pymilvus/orm/connections.py
+++ b/pymilvus/orm/connections.py
@@ -17,7 +17,6 @@ from typing import Tuple
 
 from ..client.check import is_legal_host, is_legal_port, is_legal_address
 from ..client.grpc_handler import GrpcHandler
-from ..client.utils import get_server_type, ZILLIZ
 
 from ..settings import Config
 from ..exceptions import ExceptionsMessage, ConnectionConfigException, ConnectionNotExistException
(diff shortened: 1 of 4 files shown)


pymilvus/orm/connections.py
```diff
@@ -199,9 +201,14 @@ def disconnect(self, alias: str):
...
 
-        if alias in self._connected_alias:
-            self._connected_alias.pop(alias).close()
-
+        with self._con_lock:
+            if alias in self._connected_alias:
+                gh = self._connected_alias.pop(alias)
+                self._connection_references[id(gh)] -= 1
+                if self._connection_references[id(gh)] <= 0:
+                    gh.close()
```
longjiquan, line 209: It seems that `close` is not necessary to be called, see also https://github.com/grpc/grpc/issues/19235.

longjiquan (changes requested):

pymilvus/orm/connections.py
filip-halt: It seems like its not worth reference tracking then. Instead of tracking we just reuse if available and keep them open until the program terminates. This could result in OOM in a long running program though. What do you think? Reuse channels but risk OOM or use new channels but have clean close? @xiaofan-luan @longjiquan

filip-halt: Figured out my error, I called wait_for_channel_ready on an already made channel. Guessing waiting for ready only works when unready.

pymilvus/orm/connections.py
```diff
@@ -266,96 +219,74 @@ def connect(self, alias=Config.MILVUS_CONN_ALIAS, user="", password="", db_name=
...
+                        and connection_details["db_name"] == kwargs["db_name"]
+                    ):
+                        gh = self._connected_alias[key]
+                        break
 
-            t = kwargs.get("timeout")
-            timeout = t if isinstance(t, (int, float)) else Config.MILVUS_CONN_TIMEOUT
+                if gh is None:
+                    gh = GrpcHandler(**kwargs)
+                    t = kwargs.get("timeout")
```
longjiquan, line 240: maybe `kwargs.get("timeout", None)` is better to avoid crash.

pymilvus/orm/connections.py
```diff
@@ -266,96 +219,74 @@ def connect(self, alias=Config.MILVUS_CONN_ALIAS, user="", password="", db_name=
...
         def connect_milvus(**kwargs):
-            gh = GrpcHandler(**kwargs)
+            with self._con_lock:
+                gh = None
+                for key, connection_details in self._alias.items():
+
+                    if (
+                        key in self._connected_alias
+                        and connection_details["address"] == kwargs["address"]
+                        and connection_details["user"] == kwargs["user"]
```
longjiquan, line 232: why we don't check the password?

pymilvus/orm/connections.py
```diff
@@ -266,96 +219,74 @@ def connect(self, alias=Config.MILVUS_CONN_ALIAS, user="", password="", db_name=
...
-                kwargs["secure"] = True
+        # No params, env, and cached configs for the alias
+        else:
+            raise ConnectionConfigException(message=ExceptionsMessage.ConnLackConf % alias)
 
-            connect_milvus(**kwargs, user=user, password=password, db_name=db_name)
-            return
+         # Set secure=True if username and password are provided
+        if len(user) > 0 and len(password) > 0:
+            kwargs["secure"] = True
```
longjiquan, line 286: Any reason we relate the `tls` (`secure`) with the `user` and `password`? In fact, authentication with user/password can be done without `tls`, see also https://wiki.lfaidata.foundation/display/MIL/MEP+27+--+Support+Basic+Authentication.

longjiquan (changes requested):

pymilvus/orm/connections.py
XuanYang-cn: Please remove this. We don't need to set secure=True for user and password authentication. secure=True is for tls secure connections.

XuanYang-cn (changes requested):

pymilvus/orm/connections.py
XuanYang-cn: password won't be cached, so unable to check

pymilvus/orm/connections.py
XuanYang-cn: kwargs.get("timeout") and kwargs.get("timeout", None) are identical

pymilvus/orm/connections.py
```diff
@@ -272,107 +234,97 @@ def connect(self, alias=Config.MILVUS_CONN_ALIAS, user="", password="", token=""
...
+                if gh is None:
+                    gh = GrpcHandler(**kwargs)
+                    t = kwargs.get("timeout", None)
+                    timeout = t if isinstance(t, (int, float)) else Config.MILVUS_CONN_TIMEOUT
+                    gh._wait_for_channel_ready(timeout=timeout)
+
+                kwargs.pop('password', None)
+                kwargs.pop('token', None)
+
+                self._connected_alias[alias] = gh
```
czs007, line 263: The current action involves overwriting the previous object without reducing the count by one in the _connection_references.

XuanYang-cn:
closing it for stale

PR #1472, opened 2023-05-29
Support api key (#1460)

bigsheeper:


pymilvus/client/grpc_handler.py
```diff
@@ -131,15 +133,22 @@ def reset_db_name(self, db_name):
...
+            authorization = base64.b64encode(f"{token}".encode('utf-8'))
+            keys.append("authorization")
+            values.append(authorization)
+        elif user and password:
             authorization = base64.b64encode(f"{user}:{password}".encode('utf-8'))
-            key = "authorization"
-            self._authorization_interceptor = interceptor.header_adder_interceptor(key, authorization)
+            keys.append("authorization")
+            values.append(authorization)
+        self._authorization_interceptor = interceptor.header_adder_interceptor(keys, values)
```
XuanYang-cn, line 147: Please set the interceptor if len(keys) > 0 and len(values) > 0, so that ut won't stuck forever

XuanYang-cn (changes requested):

PR #1523, opened 2023-06-08
[WIP] Simple API

filip-halt:
An update to the milvus client that is at the database level, takes advantage of partition keys, and dynamic schema.


filip-halt:
@czs007 I feel like this covers more of our features while staying easy to use. What do you think?

XuanYang-cn:
closing it for stale

Issue #1525, opened 2023-06-08
[Bug]: [High level] Shall we not set  "None" as default value for Parameter "filter" in milvus_client.query 


binbinlv:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Parameter "filter" in milvus_client.query should not have default value:

when typing milvus_client.query("binbin_milvus"), it reports
```
TypeError: query() missing 1 required positional argument: 'collection_name'
```
and when typing "milvus_client.query(collection_name="binbin_milvus")", it reports:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/pymilvus/milvus_client/milvus_client.py", line 277, in query
    raise DataTypeNotMatchException(message=ExceptionsMessage.ExprType % type(filter))
pymilvus.exceptions.DataTypeNotMatchException: <DataTypeNotMatchException: (code=1, message=The type of expr must be string ,but <class 'NoneType'> is given.)>
```

Why it will report error as None is the default value.

### Expected Behavior

1. did not set default value for filter, then it could tell users there are 2 parameters are needed

OR

2. set a excutable default value for filter

### Steps/Code To Reproduce behavior

```markdown
Only query with collection name, do not input value for parameter "filter"

milvus_client.query("binbin_milvus")
```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

### Anything else?

pymilvus: 2.2.10.dev1
milvus: 2.2.0-20230607-02e3b06b

binbinlv:
Now, will report errror, but the error message seems not very meaningful:

```
>>> milvus_client.query("milvus_client_api_alias_kxmKVlJg")
RPC error: [query], <MilvusException: (code=65535, message=empty expression should be used with limit)>, <Time:{'RPC start': '2024-01-25 11:54:36.360605', 'RPC error': '2024-01-25 11:54:36.447364'}>
Failed to query collection: milvus_client_api_alias_kxmKVlJg
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/binbin/milvus_latest/lib/python3.8/site-packages/pymilvus/milvus_client/milvus_client.py", line 393, in query
    raise ex from ex
[...]

Issue #1522, opened 2023-06-07
[Bug]: [milvus_client] The api "milvus_client.describe_collection" has no "enable_dynamic_field" values when setting enable_dynamic_field as False


binbinlv:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

The api "milvus_client.describe_collection" has no "enable_dynamic_field" values when setting enable_dynamic_field as False
```
>>>
>>> milvus_client.create_collection(collection_name, dim, enable_dynamic_field=False)
>>>
>>> milvus_client.describe_collection("binbin_milvus")
{'auto_id': False, 'description': '', 'fields': [{'name': 'id', 'description': '', 'type': <DataType.INT64: 5>, 'is_primary': True, 'auto_id': False}, {'name': 'vector', 'description': '', 'type': <DataType.FLOAT_VECTOR: 101>, 'params': {'dim': 128}}]}
>>>
```

### Expected Behavior

```
>>> milvus_client.describe_collection("binbin_milvus")
{'auto_id': False, 'description': '', 'fields': [{'name': 'id', 'description': '', 'type': <DataType.INT64: 5>, 'is_primary': True, 'auto_id': False}, {'name': 'vector', 'description': '', 'type': <DataType.FLOAT_VECTOR: 101>, 'params': {'dim': 128}}], 'enable_dynamic_field': False}
```

### Steps/Code To Reproduce behavior

```python
from pymilvus import MilvusClient
milvus_client = MilvusClient()
dim = 256
collection_name = "binbin_milvus"
milvus_client.create_collection(collection_name, dim, enable_dynamic_field=False)
milvus_client.describe_collection("binbin_milvus")
```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

binbinlv:
In low level api:

>>> collection.describe()
[...]

PR #1526, opened 2023-06-09
Fix bugs: describe_collection and query filter parameter

czs007:
relate #1522  #1525


pymilvus/client/abstract.py
```diff
@@ -187,6 +192,16 @@ def __pack(self, raw):
 
         self.properties = raw.properties
 
+    def _rewrite_schema_dict(self, schema_dict):
```
longjiquan, line 195: Make this class method?

longjiquan (changes requested):

PR #1527, opened 2023-06-09
Fix bugs: describe_collection return dict and query filter parameter

czs007:
relate #1522 #1525


longjiquan:
cherry-pick of #1526

PR #1555, opened 2023-06-26
Lost precision when querying data of type 'FLOAT_VECTOR'

chendong0120:


ponponon:
In Milvus, FLOAT_VECTOR refers to single-precision floats (32-bit floats), not double-precision floats (64-bit floats).

Milvus is an open source vector similarity search engine for efficiently storing and searching large-scale vector data. When you create a collection and specify the field type as FLOAT_VECTOR, Milvus will use single-precision floating-point numbers to store vector data, thus saving storage space and improving search performance.

If you want to store double-precision floating-point vectors, you can use the field type DOUBLE_VECTOR, but in large-scale vector data scenarios, single-precision floating-point is usually sufficient and provides a better balance between storage and performance requirements.

XuanYang-cn:
closing for wontfix

Issue #1554, opened 2023-06-25
[Bug]: [high level api] error message is not meaningful when creating collections with invalid primary key type


binbinlv:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

error message is not meaningful when creating collections with invalid primary key type
```
>>> milvus_client.create_collection(collection_name, dim, id_type="invalid")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/binbin/towhee-test-env-3.8-0.9.0/lib/python3.8/site-packages/pymilvus/milvus_client/milvus_client.py", line 82, in create_collection
    raise PrimaryKeyException(message=ExceptionsMessage.PrimaryFieldType)
pymilvus.exceptions.PrimaryKeyException: <PrimaryKeyException: (code=1, message=Param primary_field must be str type.)>
```

### Expected Behavior

code=1, message=Param id_type must be int or string.

### Steps/Code To Reproduce behavior

```markdown
from pymilvus import MilvusClient
milvus_client = MilvusClient()
collection_name = "coll1"
dim = 8
milvus_client.create_collection(collection_name, dim, id_type="invalid")
```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

binbinlv:
Now int has been added in error message.

pymilvus: 2.2.13.dev7
```
>>> milvus_client.create_collection(collection_name, dim, id_type="invalid")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/binbin/towhee-test-env-3.8-0.9.0/lib/python3.8/site-packages/pymilvus/milvus_client/milvus_client.py", line 82, in create_collection
    raise PrimaryKeyException(message=ExceptionsMessage.PrimaryFieldType)
pymilvus.exceptions.PrimaryKeyException: <PrimaryKeyException: (code=1, message=Param primary_field must be int or str type.)>
```

PR #1561, opened 2023-06-28
Enable user to set max_length for pk field of string type

czs007:
issue #1554 
releated https://github.com/milvus-io/milvus/issues/25110


longjiquan:
cherry-pick of #1560

PR #1571, opened 2023-07-07
Create index brefore wait for completion

xiaocai2333:
/kind improvement


pymilvus/client/grpc_handler.py
```diff
@@ -854,6 +856,14 @@ def get_index_state(
...
         self, collection_name: str, index_name: str, timeout: Optional[float] = None, **kwargs
     ):
+        if kwargs.get("refresh", False):
+            index_info = self.describe_index(collection_name, index_name, timeout)
+            index_params = {
+                "index_type": index_info["index_type"],
+                "metric_type": index_info["metric_type"],
+                "params": index_info["params"],
+            }
+            self.create_index(collection_name, index_info["field_name"], index_params, timeout)
```
XuanYang-cn, line 866: Do we also wait inside create_index? maybe use sync = False here

XuanYang-cn (changes requested):

pymilvus/client/grpc_handler.py
xiaocai2333: yeah, you're right. I will fix it.

PR #1572, opened 2023-07-10
Append all json data with no dynamic fields

XuanYang-cn:


pymilvus/client/entity_helper.py
```diff
@@ -203,16 +203,14 @@ def check_append(field_data: Any):
             return
 
         if field_data.type == DataType.JSON and len(field_data.scalars.json_data.data) >= index:
-            json_value = field_data.scalars.json_data.data[index]
-            json_dict = ujson.loads(json_value)
+            json_dict = ujson.loads(field_data.scalars.json_data.data[index])
 
-            if not field_data.is_dynamic:
```
czs007, line 209: should keep this logic

PR #1574, opened 2023-07-10
Seperate insert with upsert

XuanYang-cn:
Using bool to separate insert and upsert parser function is a bad coding style. And this is really painful when sync codes from master to 2.2 branch.


czs007:
please pick to 2.2 ?

PR #1584, opened 2023-07-13
Fix describe_collection failed when max_length set to non-varchar field

longjiquan:
/kind bug
/kind improvement


czs007:
need to pick to 2.2

PR #1596, opened 2023-07-17
Fix import connections in db

XuanYang-cn:


XuanYang-cn:
https://github.com/milvus-io/milvus/issues/25667

PR #1598, opened 2023-07-18
Increase the number of retry times and decrease the duration of sleep

bigsheeper:
/kind improvement


bigsheeper:
related: https://github.com/milvus-io/pymilvus/issues/1599

PR #1638, opened 2023-08-06
Unmask dataframe field name mismatch error

jyc4617:
When names and fields from a dataframe(received) are compared against those of a collection(expected), the name match check  should be done first before type match check.

Otherwise, when the dataframe and collection have different number of fields, or different ordering of the fields, a type mismatch error message could be logged, which is more misleading than a name miss match error.

### DIFF
--- a/pymilvus/orm/schema.py
+++ b/pymilvus/orm/schema.py
@@ -558,16 +558,16 @@ def check_infer_fields_valid(
         )
 
     for x, y in zip(infer_fields, tmp_fields):
+        if is_data_frame and x.name != y.name:
+            raise DataNotMatchException(
+                message=f"The name of field don't match, expected: {y.name}, got {x.name}"
+            )
         if x.dtype != y.dtype:
             msg = (
                 f"The data type of field {y.name} doesn't match, "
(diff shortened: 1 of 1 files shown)


XuanYang-cn:
Great catch!
@jyc4617 Please sign off this PR following the instruction of DCO robots: https://github.com/milvus-io/pymilvus/pull/1638/checks?check_run_id=15658631609

XuanYang-cn:
@jyc4617 Could you please rebase the latest mastse codes to pass the Code checker actions. We've fixed it in our latest codes.

jyc4617:
> @jyc4617 Could you please rebase the latest mastse codes to pass the Code checker actions. We've fixed it in our latest codes.

Hi, it appears that my branch is already rebased to the master latest(7896d05).
I have q question though, there is a pylint error in the code checker run, but it does not seem to be related to my changes.

XuanYang-cn:
The pylint fix is at 1313bdb7273c3c84c6af88167be90c82d39eceee

rebase the latest code and the pylint error will be gone

jyc4617:
Thanks, rebased to master branch latest(9acbc7e).

PR #1648, opened 2023-08-16
refine iterator params to deprecate limit for clients

MrPresent-Han:
related: https://github.com/milvus-io/milvus/issues/26358
related: https://github.com/milvus-io/milvus/issues/26397


pymilvus/orm/constants.py
```diff
@@ -27,6 +27,7 @@
 
 OFFSET = "offset"
 LIMIT = "limit"
```
XuanYang-cn, line 29: Could remove the LIMIT relate constants since not inuse

pymilvus/orm/iterator.py
```diff
@@ -195,13 +210,15 @@ def __init__(
         **kwargs,
     ) -> SearchIterator:
         if len(data) > 1:
-            raise MilvusException(message="Not support multiple vector iterator at present")
+            raise ParamError(message="Not support multiple vector iterator at present")
```
XuanYang-cn, line 213: Not clear enough

"Not support search iteration over multiple vectors at present"

pymilvus/orm/iterator.py
```diff
@@ -214,11 +231,17 @@ def __init__(
...
         self.__check_metrics()
         self.__check_radius()
         self.__seek()
         self.__setup__pk_prop()
 
+    # as we use batch_size as the page size, so we remove LIMIT here
+    def __check_remove_limit(self):
+        if self._kwargs.get(LIMIT, 0) != 0:
+            self._kwargs.pop(LIMIT)
+
```
XuanYang-cn, line 244: Why delibretely remove "limit" key? would it be clearer just NOT USE IT?

pymilvus/orm/collection.py
```diff
@@ -798,7 +798,7 @@ def search_iterator(
         data: List,
         anns_field: str,
         param: Dict,
-        limit: int,
+        batch_size: Optional[int] = 1000,
```
XuanYang-cn, line 801: Why adding defaults? Comparing to the no defaults limit.

XuanYang-cn (changes requested):

pymilvus/orm/collection.py
XuanYang-cn: https://github.com/milvus-io/milvus/issues/26358#issuecomment-1678500383

pymilvus/orm/iterator.py
MrPresent-Han: to avoid clients provide limit parameter when calling the method, otherwise the function will failed with exception 'multivalve for limit'

pymilvus/orm/collection.py
MrPresent-Han: required by clients, in case they don't provide batch_size

pymilvus/orm/iterator.py
MrPresent-Han: changed

pymilvus/orm/iterator.py
```diff
@@ -280,7 +303,7 @@ def next(self):
             self._iterator_params["data"],
```
XuanYang-cn, line 303: Please use key-value call to prevent double assign from self._kwargs and better compatibility.

```python
search(
   data=xxx,
   anns_field=xxxx,
   limit=xxx,
   expr=xxx,
)
```

pymilvus/orm/constants.py
MrPresent-Han: currently, we still use query interface of milvus client, so internally, limit is still used, the 'batch_size' parameter is used to avoid ambiguity of meaning

pymilvus/orm/iterator.py
XuanYang-cn: see the comments below, key-value calls will avoid this exception, betther not use position calls on a funciton with so many parameters, could easy got wrong, and needed extra care for the kwargs.

PR #1651, opened 2023-08-19
refine iterator params to differentiate limit for clients(#26358 #26397)

MrPresent-Han:
related: https://github.com/milvus-io/milvus/issues/26397


longjiquan (commented):
LGTM

PR #1654, opened 2023-08-22
Add optimze to get best performance

XuanYang-cn:
VectorDBBench need this, so that we don't need to change VDB everytime when server breaks some rules.

This function makes sure VDB results are reproduceable.


XuanYang-cn:
still under tests

PR #1657, opened 2023-08-23
Fix semver not match pep440

XuanYang-cn:
"no-guess-dev" will add a post1.devN to the commit

```
25hWARNING  Error during upload. Retry with the --verbose option for more details. 
32
ERROR    HTTPError: 400 Bad Request from https://test.pypi.org/legacy/          
33
         '2.3.0b1.dev124+g71b52fc' is an invalid value for Version. Error: Can't
34
         use PEP 440 local versions. See                                        
35
         https://packaging.python.org/specifications/core-metadata for more     
36
         information.   
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,8 @@ dynamic = ["version"]
 "repository" = 'https://github.com/milvus-io/pymilvus'
 
 [tool.setuptools_scm]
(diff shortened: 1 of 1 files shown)


longjiquan (approved):
/lgtm

PR #1658, opened 2023-08-23
fix for new-version of iterator(#1660)

MrPresent-Han:
related: #1660  https://github.com/milvus-io/milvus/issues/26552

### DIFF
--- a/pymilvus/orm/iterator.py
+++ b/pymilvus/orm/iterator.py
@@ -12,8 +12,11 @@ from .connections import Connections
 from .constants import (
     BATCH_SIZE,
     CALC_DIST_COSINE,
+    CALC_DIST_HAMMING,
     CALC_DIST_IP,
+    CALC_DIST_JACCARD,
     CALC_DIST_L2,
+    CALC_DIST_TANIMOTO,
     DEFAULT_SEARCH_EXTENSION_RATE,
     FIELDS,
     INT64_MAX,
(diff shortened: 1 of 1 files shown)


MrPresent-Han:
/hold
local test is running, hold for a few minutes

pymilvus/orm/iterator.py
```diff
@@ -257,6 +247,20 @@ def merge(self, others: List[Hits]):
             for other in others:
                 self._results.append(other)
 
+    def ids(self):
+        ids = []
+        for res in self._results:
+            for hit in res:
+                ids.append(hit.id)
+        return ids
```
longjiquan, line 255: Maybe ids can be cached if it's immutable, and then it will be faster next time `ids` is called.

pymilvus/orm/iterator.py
```diff
@@ -257,6 +247,20 @@ def merge(self, others: List[Hits]):
...
+
+    def distances(self):
+        distances = []
+        for res in self._results:
+            for hit in res:
+                distances.append(hit.distance)
+        return distances
```
longjiquan, line 262: Ditto.

longjiquan (changes requested):

pymilvus/orm/iterator.py
MrPresent-Han: True, it can be faster, but SearchPage is not immutable as it supports merging other SearchPages. Once changed, cached ids and distances need to be calculated once more. And currently, ids and distances methods are only called by e2e test. I suppose most users would only use get__item method to get the whole entity so caching these two list inside SearchPage would consume double-size memory and may not be very cost-effective

pymilvus/orm/iterator.py
MrPresent-Han: same reasons

pymilvus/orm/iterator.py
```diff
@@ -192,31 +195,18 @@ def close(self) -> None:
 
 
 def metrics_positive_related(metrics: str) -> bool:
-    if metrics is CALC_DIST_L2:
+    if (
+        metrics is CALC_DIST_L2
```
czs007, line 199: how about 
` return metrics in [CALC_DIST_L2, CALC_DIST_JACCARD, CALC_DIST_HAMMING, CALC_DIST_TANIMOTO]`

MrPresent-Han: fine, changed. I just know there is 'in' keyword in python

pymilvus/orm/iterator.py
```diff
@@ -192,31 +195,13 @@ def close(self) -> None:
 
 
 def metrics_positive_related(metrics: str) -> bool:
-    if metrics is CALC_DIST_L2:
+    if metrics in [CALC_DIST_L2, CALC_DIST_JACCARD, CALC_DIST_HAMMING, CALC_DIST_TANIMOTO]:
         return True
```
MrPresent-Han, line 199: changed to 'in' form

longjiquan:
Please change the title and the commit message, since github won't link issues or prs across repositories. @MrPresent-Han

longjiquan:
LGTM

PR #1673, opened 2023-09-07
bulkwriter sample read csv

yhmo:


examples/example_bulkwriter.py
```diff
@@ -41,65 +44,119 @@
...
-    local_writer.commit()
-    print("test local writer done!")
-    print(local_writer.data_path)
-    return local_writer.data_path
+def test_local_writer_npy(schema: CollectionSchema):
+    print(f"\n===================== test local npy writer ====================")
+    with LocalBulkWriter(
+            schema=schema,
+            local_path="/tmp/bulk_writer",
+            segment_size=4*1024*1024,
```
yiwen92, line 133: any reason for hardcode 4 here?

examples/example_bulkwriter.py
```diff
@@ -41,65 +44,119 @@
...
+    ) as local_writer:
+        read_sample_data("./data/train_embeddings.csv", local_writer)
```
yiwen92, line 135: Is it possible that csv is too big(100 GB) that cannot load into the processing?

examples/example_bulkwriter.py
```diff
@@ -41,65 +44,119 @@
...
+    local_writer = LocalBulkWriter(
+        schema=schema,
+        local_path="/tmp/bulk_writer",
+        segment_size=1000 * 1024 * 1024,
+        file_type=BulkFileType.JSON_RB,
+    )
     threads = []
     thread_count = 100
-    rows_per_thread = 1000
+    rows_per_thread = 100
```
yiwen92, line 159: Any limitation for size per row here?

pymilvus/bulk_writer/local_bulk_writer.py
```diff
@@ -76,20 +95,27 @@ def commit(self, **kwargs):
         x = Thread(target=self._flush, args=(call_back,))
```
xiaofan-luan, line 95: we don't need to flush every append right?
only if buffer is large than expect size? then we change to a new buffer and flush the old buffer async

PR #1675, opened 2023-09-07
Use flush_ts and collection_name when getting flush state

bigsheeper:
issue: https://github.com/milvus-io/milvus/issues/26820

/kind bug


longjiquan (approved):
/lgtm

PR #1716, opened 2023-09-22
fix conflict for ef and batch size

MrPresent-Han:
related: https://github.com/milvus-io/milvus/issues/27284


XuanYang-cn:
link #1717

Issue #1663, opened 2023-08-30
[Bug]: query_iterator gives KeyError: 'is_primary' exception


hasansustcse13:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

__setup__pk_prop function in iterator.py file produce the exception

`    
 File "E:\Projects\CloudApper\AI-ML\CloudApper Sense\sense_api\milvus_store.py", line 55, in get_all_data
    query_iterator = self.collection.query_iterator(expr=expression, **kwargs)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Projects\CloudApper\AI-ML\CloudApper Sense\venv\Lib\site-packages\pymilvus\orm\collection.py", line 934, in query_iterator
    return QueryIterator(
           ^^^^^^^^^^^^^^
  File "E:\Projects\CloudApper\AI-ML\CloudApper Sense\venv\Lib\site-packages\pymilvus\orm\iterator.py", line 70, in __init__
    self.__setup__pk_prop()
  File "E:\Projects\CloudApper\AI-ML\CloudApper Sense\venv\Lib\site-packages\pymilvus\orm\iterator.py", line 164, in __setup__pk_prop
    if field["is_primary"]:
       ~~~~~^^^^^^^^^^^^^^
KeyError: 'is_primary' `

The code in this function:
```
def __setup__pk_prop(self):
        fields = self._schema[FIELDS]
        for field in fields:
            if field["is_primary"]:
                if field["type"] == DataType.VARCHAR:
                    self._pk_str = True
                else:
                    self._pk_str = False
                self._pk_field_name = field["name"]
                break
        if self._pk_field_name is None or self._pk_field_name == "":
            raise MilvusException(message="schema must contain pk field, broke")
```
**Observation**: This will give error when ` fields = self._schema[FIELDS]` some of the fields don't have the `is_primary` key.
 **Possible Solution/Fix:** This is a common dictionary error. Instead of writing this `if field["is_primary"]` we can write this code as `if field.get("is_primary")`. It will fix the issue. 
**Version**: pymilvus==2.3.0
            


### Expected Behavior

_No response_

### Steps/Code To Reproduce behavior

_No response_

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
[...]

PR #1745, opened 2023-10-19
add judgement for isPrimary

MrPresent-Han:
related: #1663

### DIFF
--- a/pymilvus/orm/constants.py
+++ b/pymilvus/orm/constants.py
@@ -36,6 +36,7 @@ RADIUS = "radius"
 RANGE_FILTER = "range_filter"
 FIELDS = "fields"
 EF = "ef"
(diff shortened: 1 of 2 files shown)


hasansustcse13:
@MrPresent-Han 

This [issue ](https://github.com/milvus-io/pymilvus/issues/1663) is still persist as you don't check the key before access in query iterator but checked in search iterator. 

**QueryIterator:__setup__pk_prop**
`if field[IS_PRIMARY]:`

**SearchIterator:__setup__pk_prop**
`if IS_PRIMARY in field and field[IS_PRIMARY]:`

**Question**: Can anyone tell me `self.collection.query(expr=expression, output_fields=output_fields, **kwargs)` return all the rows from DB if the collection has so many docs(more than 16384, assume 100000)  if don't set any limit.

PR #1748, opened 2023-10-23
Fix search result

XuanYang-cn:
1. Print id instead of pk
2. Return the correct data type for future.result()


XuanYang-cn:
See also: milvus-io/milvus#25504

longjiquan (approved):
/lgtm

PR #1750, opened 2023-10-24
Fix has_collection result with Milvus2.2

XuanYang-cn:
For Milvus2.2.x, PyMilvus needs to check for `error_code` before checking codes, cuz code will alwasy returns 0, which indicates rpc success


NicoYuan1986:
lgtm

longjiquan (approved):
/lgtm

PR #1776, opened 2023-11-09
Revert "Add float16 Vector Approve"

XuanYang-cn:
Reverts milvus-io/pymilvus#1772

@Writer-X We're reverting this PR because the next version of pymilvus & Milvus is not ready for float16, and we don't want to give false signal to our user that it's ready. After releasing the next pymilvus==2.3.3, we'll revert this revert


longjiquan (approved):
/lgtm

PR #1777, opened 2023-11-10
Modify highlevel delete to not return any value

czs007:
Starting from version 2.3.2, Milvus no longer includes the primary keys in the result when processing the delete operation with expressions. This change is due to the large amount of data involved. The delete interface no longer returns any results. If no exceptions are thrown, it indicates a successful deletion. However, for backward compatibility, if the primary_keys returned from Milvus is not empty the list of primary keys is still returned.

### DIFF
--- a/pymilvus/milvus_client/milvus_client.py
+++ b/pymilvus/milvus_client/milvus_client.py
@@ -379,6 +379,15 @@ class MilvusClient:
         Delete all the entries based on the pk. If unsure of pk you can first query the collection
         to grab the corresponding data. Then you can delete using the pk_field.
 
+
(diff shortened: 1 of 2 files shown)


longjiquan (approved):
/lgtm

pymilvus/milvus_client/milvus_client.py
```diff
@@ -379,6 +379,15 @@ def delete(
         Delete all the entries based on the pk. If unsure of pk you can first query the collection
         to grab the corresponding data. Then you can delete using the pk_field.
 
+
```
XuanYang-cn, line 382: The return type should be changed to Optional[List[Union[str, int]]]

XuanYang-cn (changes requested):

PR #1786, opened 2023-11-15
Infer ARRAY datatype from list like non numeric objects

### LINKED ISSUE #1785: [Bug]: The pymilvus data type inference fails to identify the ARRAY data type for lists.

### Describe the bug

The pymilvus data type inference fails to identify the ARRAY data type for lists.
This is the issue with `infer_dtype_bydata` function. I have observed it while using langchain vectorstore connector which uses this function to infer metadata types. The inference doesn't work for metadata with list of strings or booleans. 

### Expected Behavior

`infer_dtype_bydata` should return Datatype.ARRAY for list like objects.

```markdown
Call `infer_dtype_bydata(["A", "B"])` returns DataType.UNKNOWN instead of DataType.ARRAY
```

### Environment details

_No response_

### Anything else?

_No response_

abratnap:
Currently, when inferring the data type for arrays that contain non-numeric objects, the result is set to UNKNOWN data type. We need to adjust this behavior to correctly infer it as an ARRAY data type.

Resolves
#1785


axiangcoding:
hi, any progress on this PR? i have face the same issue.

PR #1818, opened 2023-12-15
use custmized version

XuanYang-cn:


XuanYang-cn:
image

XuanYang-cn:
image
https://devguide.python.org/versions/

_version_helper.py
```diff
@@ -0,0 +1,97 @@
...
+    hg.parse_archival,
+]
+
+
+def parse(root: str, config: Configuration) -> ScmVersion | None:
+    for maybe_parse in try_parse:
+        try:
+            parsed = maybe_parse(root, config)
+        except OSError as e:
+            log.warning("parse with %s failed with: %s", maybe_parse, e)
```
longjiquan, line 46: should be `maybe_parse.__name__`?

_version_helper.py
```diff
@@ -0,0 +1,97 @@
...
+        else:
+            if parsed is not None:
+                return parsed
+    else:
```
longjiquan, line 50: No `if` statement for this else branch.

longjiquan (changes requested):

_version_helper.py
XuanYang-cn: for `git.parse` maybe parse will give more info like
```
<function parse at 0x7fa03a40e5f0>
```

but __name__ would be just 
```
parse
```

PR #1881, opened 2024-01-22
fix: change confusing error message

smellthemoon:
related with https://github.com/milvus-io/milvus/issues/30094


pymilvus/exceptions.py
```diff
@@ -191,7 +191,7 @@ class ExceptionsMessage:
     PartitionKeyFieldType = "Param partition_key_field must be str type."
     PartitionKeyFieldNotExist = "the specified partition key field {%s} not exist"
     IsPartitionKeyType = "Param is_partition_key must be bool type."
-    DataTypeInconsistent = "The data in the same column must be of the same type."
+    DataTypeInconsistent = "Data type is inconsistent with input data, please check it."
```
XuanYang-cn, line 194: It's for insert data check, better to change to  "The Input data type is inconsistent with defined schema ...."

XuanYang-cn (changes requested):

PR #1904, opened 2024-01-30
BulkWriter supports Azure blob storage

yhmo:


pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies=[
     "requests",
     "minio>=7.0.0",
     "pyarrow>=12.0.0",
+    "azure.storage.blob",
```
XuanYang-cn, line 30: Please use azure-storage-blob instead of

XuanYang-cn (changes requested):

Issue #1952, opened 2024-03-05
[Bug]: [milvus_client] Failed to enable dynamic field for an existing non-dynamic field when not specifying it in the scheme before


binbinlv:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Failed to enable dynamic field for an existing non-dynamic field when not specifying it in the scheme before
```
{'collection_name': 'customized_setup_binbin_new_1', 'auto_id': False, 'num_shards': 1, 'description': '', 'fields': [{'field_id': 100, 'name': 'my_id', 'description': '', 'type': 5, 'params': {}, 'element_type': 0, 'is_primary': True}, {'field_id': 101, 'name': 'my_vector', 'description': '', 'type': 101, 'params': {'dim': 5}, 'element_type': 0}], 'aliases': [], 'collection_id': 448169809257578564, 'consistency_level': 2, 'properties': {}, 'num_partitions': 1, 'enable_dynamic_field': False}
```

### Expected Behavior

'enable_dynamic_field': True

### Steps/Code To Reproduce behavior

```python
from pymilvus import MilvusClient, DataType
import numpy as np

client = MilvusClient()


schema = MilvusClient.create_schema()
schema.add_field(field_name="my_id", datatype=DataType.INT64, is_primary=True)
schema.add_field(field_name="my_vector", datatype=DataType.FLOAT_VECTOR, dim=5)

dim = 5
collection_name = "customized_setup_binbin_new_1"
if collection_name in client.list_collections():
    client.drop_collection(collection_name)
    print("dropped collection")
client.create_collection(collection_name=collection_name, schema=schema)
client.create_collection(collection_name=collection_name, schema=schema, enable_dynamic_field=True)
```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0): master-20240305-1936aa4c
- pymilvus: 2.4.0rc47
- Milvus configuration (Settings you made in `server_config.yaml`):
```

yiwen92:
I think 2 command for create_collection, but with different params - enable_dynamic_field=True, server will ignore the second command, so this behavior is acceptable.

<img width="999" alt="image" src="https://github.com/milvus-io/pymilvus/assets/34636520/8f685cce-f80c-41e7-89b6-0a52cf6b530e">

yiwen92:
[...]

PR #1954, opened 2024-03-08
Avoiding parameter conflicts and add compatibility logic

czs007:
issue #1952

### DIFF
--- a/pymilvus/milvus_client/index.py
+++ b/pymilvus/milvus_client/index.py
@@ -37,8 +37,10 @@ class IndexParam:
 
 
 class IndexParams:
-    def __init__(self):
+    def __init__(self, field_name: str = "", **kwargs):
         self._indexes = {}
+        if field_name:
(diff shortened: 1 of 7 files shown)


pymilvus/milvus_client/index.py
```diff
@@ -37,8 +40,10 @@ def __eq__(self, other: None):
 
 
 class IndexParams:
-    def __init__(self):
+    def __init__(self, field_name: Optional[str] = "", **kwargs):
```
XuanYang-cn, line 43: Optional mean None or str, if it has a default value "", just use `filed_name: str = ""`

XuanYang-cn (changes requested):

pymilvus/milvus_client/index.py
XuanYang-cn: Or `field_name: Optional[str] = None`

PR #1958, opened 2024-03-11
enhance: update proto version to latest

zhengbuqian:
mainly to make sure pymilvus can have access to sparse related protos 

issue: https://github.com/milvus-io/milvus/issues/29419


zhengbuqian:
Note that other than proto submodule commit update and generated files change, this PR also updates `pymilvus/grpc_gen/python_gen.sh`, as we introduced a new `rg.proto` file to milvus-proto repo

PR #2008, opened 2024-03-28
fix:release partition before drop it

wildBigPanda:
when call  Collection drop_partition func，has error:

pymilvus.decorators | RPC error: [drop_partition], <MilvusException: (code=1, message=partition cannot be dropped, partition is loaded, please release it first)>, <Time:{'RPC start': '2024-03-28 16:13:04.921025', 'RPC error': '2024-03-28 16:13:04.923609'}>

fix Collection drop_partition func call，release partition before drop it

### DIFF
--- a/pymilvus/orm/collection.py
+++ b/pymilvus/orm/collection.py
@@ -1242,6 +1242,7 @@ class Collection:
             False
         """
         conn = self._get_connection()
+        conn.release_partitions(self._name, [partition_name])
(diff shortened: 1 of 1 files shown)


wildBigPanda:
@XuanYang-cn Is the design unreasonable? To drop partition, you have to call the release interface again separately, and the related interface for releasing partitions cannot be found in [collection.py]. How can you release it and drop a partitions?

PR #2022, opened 2024-04-09
covert params' value from string to int

PowderLi:
to avoid such param error:
```
pymilvus.exceptions.ParamError: <ParamError: (code=1, message=Collection field dim is 512, but entities field dim is 512)>
```


XuanYang-cn:
@PowderLi Please cherry-pick this pr to branch 2.4. THX.

PR #2023, opened 2024-04-09
Support UDS

junjiejiangjjj:


pymilvus/orm/connections.py
```diff
@@ -254,6 +254,8 @@ def __get_full_address(
             return address, None
 
         if uri != "":
+            if isinstance(uri, str) and uri.startswith("unix:"):
+                return uri, None
```
XuanYang-cn, line 258: __get_full_address should return a str with format  `{host}:{port}`, uri startswiths "unix:" doesn't seems to be correct.

And why do you need a uri starts with "unix:" scheme? I don't know how is this possible for self-deployed Milvus. Could you please file an issue to describe in detail why is it needed?

junjiejiangjjj: https://grpc.github.io/grpc/cpp/md_doc_naming.html
Using UDS (unix:path, unix://absolute_path – Unix domain sockets (Unix systems only)
) instead of ports resolves possible port conflicts

XuanYang-cn:
@junjiejiangjjj Please help cherry-pick this pr to branch 2.4, THX

Issue #2131, opened 2024-06-14
[Bug]: Can not use poetry to install pymilvus >= 2.4


Raysilience:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

create a new environment using poetry, execute in terminal `poetry add pymilvus=^2.4.3`, it will raise the following error,
![image](https://github.com/milvus-io/pymilvus/assets/45241093/0331b760-54b3-4c8d-b074-7dfa0ab2a57e)

### Expected Behavior

successfully install pymilvus with poetry 

### Steps/Code To Reproduce behavior

```markdown
1. create a new environment using poetry
2. execute in terminal `poetry add pymilvus=^2.4.3`
```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

zhuwenxing:
@Raysilience 
what is your os system? It works well in my Mac
![image](https://github.com/milvus-io/pymilvus/assets/12268675/7a589570-ba34-4a24-b7cf-c69afd1782ae)

pymilvus 2.3.4 includes a milvus-lite lib, and it is not supported in windows.

Raysilience:
Mine is windows. Then how to exclude the milvus-lite while installing pymilvus.

zhuwenxing:
> @zhuwenxing Mine is windows. Then how to exclude the milvus-lite while installing pymilvus.

@XuanYang-cn @junjiejiangjjj 
any advice?

Raysilience:
@zhuwenxing @XuanYang-cn @junjiejiangjjj  I check the pyproject.toml file. Maintainers list the milvus-lite inside the dependencies that leads to mandatory installation. This will definitely affect the developers operating on windows who are also using poetry to manage their projects. Would you consider moving the milvus-lite to optional dependencies such that windows developers still have access to pymilvus while other users are still able to install milvus-lite via poetry add pymilvus[milvus-lite] i.e. 
[project.optional-dependencies]
milvus-lite = [
  "milvus-lite>=2.4.0",
]
If so, I could create a pr to solve it.

XuanYang-cn:
@Raysilience we want "milvus-lite" to be the default dependency of pymilvus. So "No" for change it to optional dependencies for all platform.

[...]

PR #2134, opened 2024-06-16
feat: disable installation of milvus-lite on windows platform (#2131)

Raysilience:
@XuanYang-cn 
Since the project is managed by setuptools, and according to the [setuptools](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html), we can disable the automatic installation of milvus-lite on windows platform by adding environment markers. 
Following [PEP508](https://peps.python.org/pep-0508/), I pick the environment marker "sys_platform != 'win32'" to prevent the installation. 
I build packages after the modification. Then test the .tar.gz and the .whl files both on my windows laptop and a linux docker. The issue mentioned in #2131 seems gone. 
Due the lack of macOS machine, I could not perform testing on that particular platform. Please check validity before merging.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies=[
     "ujson>=2.0.0",
     "pandas>=1.2.4",
     "numpy<1.25.0;python_version<='3.8'",
-    "milvus-lite>=2.4.0",
+    "milvus-lite>=2.4.0;sys_platform!='win32'",
 ]
 
 classifiers=[


XuanYang-cn:
https://github.com/milvus-io/pymilvus/actions/runs/9545655624/job/26306874213?pr=2136

@Raysilience I tested it on windows platform by github actions, and it works perfectly. THX for the contribution!

/lgtm
/approve

PR #2141, opened 2024-06-18
enhance: update milvus-proto and correct index.drop()

XuanYang-cn:
Index name and field name is already there for Index object, No need to get them from kwargs when drop

```python
ivf_flat = Index(c, "embeddings", params, index_name="ivf_flat",
    construct_only=True)
ivf_flat.drop()
```


XuanYang-cn:
@Mergifyio refresh

PR #2144, opened 2024-06-20
return array data type in  infer_dtype_bydata

rgupta2508:
Now milvus supporting ARRAY data types so we can return DataType.ARRAY in case of if it is not VECTOR.

This will be helpful to create ARRAY data type field seamlessly  in milvus


rgupta2508:
@wangting0128 @longjiquan Please review , if this can be merged to return Milvus.ARRAY data type for infer_dtype_bydata function.

Or any thing we can do.

rgupta2508:
This PR will resolve this issue https://github.com/milvus-io/pymilvus/issues/2165
idea is to take kwargs which accept datatype along with all the parameter, based on kwargs we can decide return type.
As per current implementation it is difficult to find which one is vector array or float array

longjiquan:
Please help to review this. @xiaocai2333 
/assign @xiaocai2333

xiaocai2333:
It is recommended to change the default behavior to infer the vector, as pymilvus itself relies on the interface.

pymilvus/orm/types.py
```diff
@@ -121,7 +129,17 @@ def infer_dtype_bydata(data: Any):
             failed = True
         if not failed:
             d_type = dtype_str_map.get(type_str, DataType.UNKNOWN)
-            return DataType.FLOAT_VECTOR if is_numeric_datatype(d_type) else DataType.UNKNOWN
+            if is_varchar_datatype(d_type) or is_bool_datatype(d_type):
+                return DataType.ARRAY
+            if kwargs is None or len(kargs) == 0:
```
xiaocai2333, line 134: typo， kargs --> kwargs

longjiquan:
@rgupta2508 thanks for the contributions. I agree that we should infer data as array when the data is list-like and the infered type is bool or varchar, but it's also not enough to infer the schema since element type is required for array.

longjiquan:
@rgupta2508 Take array type as considerations, I prefer that this function should infer a field schema instead of a data type (since data type don't contain other informations).

longjiquan:
If so, I think we should do much more refactors.

rgupta2508:
> @rgupta2508 thanks for the contributions. I agree that we should infer data as array when the data is list-like and the infered type is bool or varchar, but it's also not enough to infer the schema since element type is required for array.

Yes correct , we cant return field schema with this, this we can use for data type only. directly we cant use it for schema. even max_capacity for array and max_length for varchar array is also mandatory to create schema, not sure if we can use this method for this this will just return DataType only.

used should handle this if they will get datatyepe Array.

rgupta2508:
> @rgupta2508 Take array type as considerations, I prefer that this function should infer a field schema instead of a data type (since data type don't contain other informations).

Yes correct, but even max_capacity for array and max_length for varchar array is also mandatory to create schema, not sure if we can add some default values on this.

rgupta2508:
Any suggestion on this PR . if this can be merged with this idea.
Similar issue previously https://github.com/milvus-io/pymilvus/issues/1785

rgupta2508:
Hi @longjiquan  Any suggestion on this PR . if this can be merged with this idea.
Similar issue previously https://github.com/milvus-io/pymilvus/issues/1785

PR #2210, opened 2024-08-02
fix: select a single column consisting of a list of column names

smellthemoon:
qWmIH3Y6zJ
fix it


smellthemoon:
An error will be reported, if 1. autoid==true; 2. data format is dataframe; 3. has pk name in dataframe, but has no data.
could you plz help to check and add some tests? @NicoYuan1986

Issue #2221, opened 2024-08-08
[Bug]: gRPC handler is reraising Exceptions in a way that breaks low-level Python Exception tracking


bcalvert-graft:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Hello Milvus maintainers,

Thank you in advance for reading my bug-report.

There are three spots in `grpc_handler.py` ([first spot](https://github.com/milvus-io/pymilvus/blob/66c23629cca718f4dc685d9a9efa31a207ea2442/pymilvus/client/grpc_handler.py#L582), [second spot](https://github.com/milvus-io/pymilvus/blob/66c23629cca718f4dc685d9a9efa31a207ea2442/pymilvus/client/grpc_handler.py#L619) and [third spot](https://github.com/milvus-io/pymilvus/blob/66c23629cca718f4dc685d9a9efa31a207ea2442/pymilvus/client/grpc_handler.py#L679)) that use the following pattern to reraise an `Exception`
```python
try:
    <some code>
except Exception as err:
    <do some other stuff>
    raise err from err
...
```
Respectfully, this `raise err from err` is not a great way to reraise exceptions; instead, as [mentioned in the Python docs](https://docs.python.org/3/tutorial/errors.html#handling-exceptions) it should be
```python
try:
    <some code>
except Exception as err:
    <do some other stuff>
    raise
...
```
Why is the first one not great? It attaches the `err` `Exception` as the `__cause__` to itself. As I understand it, this is not the desired semantics for `__cause__` and this breaks any tools that recursively unroll chained `Exception`s (e.g. for processing stack traces). As a concrete example of such a tool, we actually uncovered this seeming bug in `grpc_handler.py` when using Milvus as part of a larger [Dask](https://docs.dask.org/en/stable/) computation; specifically, as part of Dask's error-handling, the library uses a utility called [`collect_causes`](https://github.com/dask/distributed/blob/3d47d3a/distributed/core.py#L1624-L1629),
```python
def collect_causes(e: BaseException) -> list[BaseException]:
    causes = []
    while e.__cause__ is not None:
        causes.append(e.__cause__)
        e = e.__cause__
    return causes
```
When `collect_causes` encounters an `Exception` raised from one of the linked spots in `grpc_handler.py` it triggers an infinite loop of unrolling `e.__cause__`, consequently leaking memory until the process is killed.

[...]

PR #2225, opened 2024-08-09
Change `raise e from e` calls to be just `raise` and remove unneeded ones

bcalvert-graft:
# Description
As per the discussion from #2221, the pattern,
```python
try:
    <do something that might raise Exception>
except ExceptionType as exc
    <maybe do some additional stuff here>
    raise exc from exc
```
can break downstream tooling that seeks to unroll `exc.__cause__` (e.g. for processing the stack trace).

This PR tweaks all spots in the `pymilvus` codebase that were using that pattern to instead either just use this pattern
```python
try:
    <do something that might raise Exception>
except ExceptionType as exc
    <maybe do some additional stuff here>
    raise
```
or, if the `try/except` didn't have any additional logic in the `except` block, I removed the `try/except` entirely as a direct `reraise` would be redundant in those cases.

**NB** As mentioned in #2221, I originally only noticed three spots where the `raise err from err` pattern was used. While working on this PR I found a number of others and I ended up changing all of them. As per @XuanYang-cn's [comment](https://github.com/milvus-io/pymilvus/issues/2221#issuecomment-2277062379), if there are important reasons for one or more of these spots to stay as-is, please let me know and I can revert the changes to that section accordingly.
### Finding all the spots
To find all the spots, I ran this `grep`
```bash
$ grep "raise \(.*\) from \1" pymilvus/**/*.py 
```
On `master` branch, we get a non-zero number of hits with the expected pattern. On the branch for this PR, the `grep` returns no values
### Testing the fix
In one process, run
```python
In [1]: from milvus_lite.server import Server

In [2]: s = Server('test.db', '127.0.0.1:9999')

In [3]: s.start()
Out[3]: True

In [4]: s._p.wait()
```
In another process, run
```python
In [1]: import pymilvus

In [2]: client = pymilvus.MilvusClient("http://localhost:9999")
   ...: collection_name = "fake_embeddings"
   ...: if client.has_collection(collection_name=collection_name):
   ...:     client.drop_collection(collection_name=collection_name)
   ...: 

In [3]: vector_dim = 512
   ...: client.create_collection(
   ...:     collection_name=collection_name,
   ...:     vector_field_name="vector",
   ...:     dimension=vector_dim,
   ...:     auto_id=True,
   ...:     enable_dynamic_field=True,
   ...:     metric_type="COSINE",
   ...: )

In [4]: import numpy as np

In [5]: vectors = np.random.normal(loc=0, scale=1, size=(vector_dim + 2,))

In [6]: try:
   ...:     client.insert(collection_name, {"vector": vectors})
   ...: except pymilvus.MilvusException as exc:
   ...:     assert exc.__cause__ is not exc, "Exception is its own cause"
```
The `In [6]:` step will not through an `AssertionError`

### DIFF
--- a/pymilvus/bulk_writer/local_bulk_writer.py
+++ b/pymilvus/bulk_writer/local_bulk_writer.py
@@ -139,7 +139,7 @@ class LocalBulkWriter(BulkWriter):
                     call_back(file_list)
         except Exception as e:
             logger.error(f"Failed to fulsh, error: {e}")
-            raise e from e
+            raise
         finally:
             del self._working_thread[threading.current_thread().name]
             logger.info(f"Flush thread finished, name: {threading.current_thread().name}")
--- a/pymilvus/client/grpc_handler.py
+++ b/pymilvus/client/grpc_handler.py
@@ -151,8 +151,6 @@ class GrpcHandler:
                 code=Status.CONNECT_FAILED,
                 message=f"Fail connecting to server on {self._address}, illegal connection params or server unavailable",
             ) from e
-        except Exception as e:
-            raise e from e
 
     def close(self):
         self.deregister_state_change_callbacks()
@@ -579,7 +577,7 @@ class GrpcHandler:
         except Exception as err:
             if kwargs.get("_async", False):
                 return MutationFuture(None, None, err)
-            raise err from err
+            raise
         else:
             return m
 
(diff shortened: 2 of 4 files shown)


yunmanger1:
Can we get this fix anytime soon? This breaks our exception handling forcing to make some workarounds

PR #2244, opened 2024-08-30
enhance: [2.4] Make load parameter naming normal (#2243)

congqixia:
Previously added parameters for `load` API have leading underscore, which is misused and copied from `_async` parameter. The leading "_" is to avoid use python keyword `async` and other parameters do not have to follow this pattern.

This PR make `load` API support parameter with or without leading underscore and update related comment. Note that the ones without leading "_" have higher priority than the ones with it.


XuanYang-cn (approved):
/lgtm
/approve

PR #2249, opened 2024-09-03
fix: move page_retain_order to the same level as radius

PwzXxm:
issue: https://github.com/milvus-io/milvus/issues/35464


XuanYang-cn (approved):
/lgtm
/approve

PR #2275, opened 2024-09-26
fix not check None Type in json and support no need to pass in None when insert row

smellthemoon:
https://github.com/milvus-io/milvus/issues/36354
https://github.com/milvus-io/milvus/issues/36486


XuanYang-cn (approved):
/lgtm

Issue #2276, opened 2024-09-26
[Bug]: When using the bulk writer to format CSV files, if there are bf16 and float16 vector types present, a JSON dump error will occur.


zhuwenxing:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When using the bulk writer to format CSV files, if bf16 and float16 vector types are present, a JSON dump error will occur.

Failed to convert field float16_vec_field value to string, error: Object of type ndarray is not JSON serializable 

### Expected Behavior

works well

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

zhuwenxing:
/assign @yhmo 
Can you help me to assign the issue?

zhuwenxing:
https://github.com/milvus-io/pymilvus/blob/master/examples/example_bulkinsert_csv.py
@OxalisCu 
This example lacks many data types and can be supplemented with all data types. At the same time, this example also lacks the usage of bulk writers, please add that as well.

zhuwenxing:
https://github.com/milvus-io/pymilvus/blob/da51ba14edde3c67b682f2d0ae1b85a095815692/examples/example_bulkinsert_csv.py#L100-L117

@OxalisCu  @yhmo 

At the same time, is the usage of Dynamic SCHEMA correct in this example? I understand it should be written like this, right?

```
def gen_csv_rowbased(num, path, partition_name, sep=","):
    global id_start
    header = [_ID_FIELD_NAME, _JSON_FIELD_NAME, _VECTOR_FIELD_NAME, _VARCHAR_FIELD_NAME, "$meta"]
    rows = []
    for i in range(num):
        rows.append([
            id_start, # id field
            json.dumps({"Number": id_start, "Name": "book_"+str(id_start)}), # json field
            json.dumps([round(random.random(), 6) for _ in range(_DIM)]), # vector field
            "{}_{}".format(partition_name, id_start) if partition_name is not None else "description_{}".format(id_start), # varchar field
             json.dumps({"dynameic_fiield_1": 1, "dynamic_field_2":2}), # no field matches this value, this value will be put into dynamic field
        ])
        id_start = id_start + 1
    data = [header] + rows
    with open(path, "w") as f:
[...]

PR #2281, opened 2024-09-30
fix: Fix the f16 and bf16 dump error & add all type test for csv & csv supports null

OxalisCu:
https://github.com/milvus-io/pymilvus/issues/2276

### DIFF
--- a/pymilvus/bulk_writer/buffer.py
+++ b/pymilvus/bulk_writer/buffer.py
@@ -10,7 +10,6 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 
-import csv
 import json
 import logging
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


pymilvus/bulk_writer/buffer.py
```diff
@@ -290,49 +289,62 @@ def _persist_parquet(self, local_path: str, **kwargs):
 
     def _persist_csv(self, local_path: str, **kwargs):
         sep = self._config.get("sep", ",")
+        # nullkey is not supported in csv now
```
tedxu, line 292: Since CSV bulk insert has already supported nulls, can we add null support here?

pymilvus/bulk_writer/buffer.py
```diff
@@ -290,49 +289,62 @@ def _persist_parquet(self, local_path: str, **kwargs):
...
+                data[k] = pd.Series(v, dtype=dt)
+            else:
+                data[k] = pd.Series(v)
 
         file_path = Path(local_path + ".csv")
         try:
-            with file_path.open("w", encoding="utf-8") as f:
-                writer = csv.writer(f, delimiter=sep)
-                writer.writerows(rows)
+            data.to_csv(file_path, sep=sep, index=False)
```
tedxu, line 345: We should redefine the null value representation here, e.g., "NULL", configurable variable preferred.

PR #2285, opened 2024-10-09
Add database name to alias dictionary for retrieval in list_connections

### LINKED ISSUE #2161: [Bug]: db_name should return when list_connections called

### Describe the bug

I have a requirement where I need to connect to different db on same address (host + port) at the same time. And I have to use existing connection if it's alive for this is the code has been written (basically written in langchain codebase)

 
```
  if given_address is not None:
            for con in connections.list_connections():
                addr = connections.get_connection_addr(con[0])
                if (
                    con[1]
                    and ("address" in addr)
                    and (addr["address"] == given_address)
                    and ("user" in addr)
                    and (addr["user"] == tmp_user)
                ):
                    logger.debug("Using previous connection: %s", con[0])
                    return con[0]

        # Generate a new connection if one doesn't exist
        alias = uuid4().hex
        try:
            connections.connect(alias=alias, **connection_args)
            logger.debug("Created new connection using: %s", alias)
            return alias
        except MilvusException as e:
            logger.error("Failed to create new connection using: %s", alias)
            raise e

```

Let's say collection_1 is connected to db_1 and Now I want collection_2 will be connect to db_2.  As `addr = connections.get_connection_addr(con[0])` don't have the db_name value the if block will be executed and provide db_1 connection which is wrong.

I believe this is single line fix when connect to milvus we should store the db_name instead of popping. 
**Code from Milvus:**

```
def connect_milvus(**kwargs):
            # rest of the code
            kwargs.pop("password")
            kwargs.pop("token", None)
            kwargs.pop("db_name", "")  # the bug

            self._connected_alias[alias] = gh
            self._alias[alias] = copy.deepcopy(kwargs)
```

### Expected Behavior

_No response_

_No response_

### Environment details

```markdown
```

### Anything else?

_No response_

hasansustcse13:
Fix: https://github.com/milvus-io/pymilvus/issues/2161

### DIFF
--- a/pymilvus/orm/connections.py
+++ b/pymilvus/orm/connections.py
@@ -406,7 +406,6 @@ class Connections(metaclass=SingleInstanceMetaClass):
                 )
             kwargs.pop("password")
             kwargs.pop("token", None)
-            kwargs.pop("db_name", "")
 
             self._connected_alias[alias] = gh
             self._alias[alias] = copy.deepcopy(kwargs)


hasansustcse13:
@XuanYang-cn @longjiquan Could you please see this.

hasansustcse13:
@XuanYang-cn @longjiquan @czs007

XuanYang-cn:
closing as wontfix

PR #2286, opened 2024-10-09
fix: upsert rows when set autoid==true fail

smellthemoon:
https://github.com/milvus-io/milvus/issues/36710


XuanYang-cn (approved):
/lgtm
/approve

PR #2291, opened 2024-10-11
feat: allow empty sparse row

zhengbuqian:
currently pymilvus supports the following sparse vector formats:

* scipy sparse classes
* list of tuples: `[(dim, val), (dim, val), ...]`
* dict: `{dim: val, dim: val, ...}`

When dealing with input data, pymilvus doesn't check the schema to infer the data type, instead it tries to infer the data type based on the shape. that's why we have `entity_is_sparse_matrix` method.

Now we want to support empty sparse vector row, it is easy for scipy sparse classes, but for the other 2 format, it means we must infer empty list/dict `[]/{}` as sparse vector. 

Will this be problematic? what could be the other options?


zhengbuqian:
> Is this a subtask of None support？

no, this is to allow empty sparse vector(no valid dim), not null support

PR #2257, opened 2024-09-08
support new Function feature

zhengbuqian:
currently only BM25 Function is supported.

issue: https://github.com/milvus-io/milvus/issues/35853

PR #2294, opened 2024-10-12
fix: remove duplicate field name check in pymilvus

zhengbuqian:
this check was added in https://github.com/milvus-io/pymilvus/pull/2257. 

this caused an test in milvus to fail. the test expects the error to be thrown on server side instead of sdk side. it is ok to check on sdk side but I choose to remove it to keep the behavior consistent and comfort the ut without modifying it. 

Failed test: https://github.com/milvus-io/milvus/blob/5b2658327d676453e8a251ffc3273f4a218fb278/tests/python_client/testcases/test_collection.py#L318


XuanYang-cn (approved):
/lgtm
/approve

PR #2309, opened 2024-10-23
fix: entity_is_sparse_matrix: each row should be dict or list

zhengbuqian:
issue: https://github.com/milvus-io/milvus/issues/37022 and https://github.com/milvus-io/milvus/issues/37021

list of empty string is erroneously treated as sparse vector

adding a check to explicitly require that, if the input data will be seen as sparse vector, its data type, if not scipy format, should be list of list or list of dict.


XuanYang-cn (approved):
/lgtm

PR #2319, opened 2024-11-02
fix: bulkwriter to skip function output fields

zhengbuqian:
https://github.com/milvus-io/milvus/issues/37196


XuanYang-cn (approved):
/lgtm

PR #2347, opened 2024-11-14
fix insert to avoid describe collection on every insert call

zhengbuqian:
issue: https://github.com/milvus-io/milvus/issues/37622


XuanYang-cn (approved):
/lgtm

Issue #2327, opened 2024-11-06
[Bug]: `TypeError` if `consistency_level` is passed to `Collection.delete`


ofir-ov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When I do

```python
my_collection.delete(expr='...', consistency_level='Strong')  # Or any `consistency_level` for that matter
```

I get the following error:

```python
    @retry_on_rpc_failure()
    def delete(
        self,
        collection_name: str,
        expression: str,
        partition_name: Optional[str] = None,
        timeout: Optional[float] = None,
        **kwargs,
    ):
        check_pass_param(collection_name=collection_name, timeout=timeout)
        try:
>           req = Prepare.delete_request(
                collection_name,
                partition_name,
                expression,
                consistency_level=kwargs.get("consistency_level", 0),
                param_name=kwargs.pop("param_name", None),
                **kwargs,
            )
E           TypeError: pymilvus.client.prepare.Prepare.delete_request() got multiple values for keyword argument 'consistency_level'

/usr/local/lib/python3.12/site-packages/pymilvus/client/grpc_handler.py:597: TypeError
```

### Expected Behavior

To not fail

### Steps/Code To Reproduce behavior

All you need to do is pass the `consistency_level` to the `delete` method. This bug is directly due to the use of `kwargs.get` in the function in order to populate a "new" keyword argument called `consistency_level` in addition to the existing one. It should happen no matter the other circumstances.

### Environment details

```markdown
- Hardware/Softward conditions: (OS: `6.10.0-linuxkit` Docker Desktop on Mac running Linux using the image `python:3.12-slim`, CPU: Apple M2 Pro, Memory: 8GB)
- Method of installation (Docker, or from source): A Docker image, as mentioned above. I'm using LangChain, so technically not calling the collection directly. `langchain-milvus==0.1.4`
- Milvus version (v0.3.1, or v0.4.0): Milvus is running using the Docker image `milvusdb/milvus:v2.4.6`
- Milvus configuration (Settings you made in `server_config.yaml`): Didn't change anything.
```

### Anything else?

The bug is inherit in that call in the code, it basically will never work if `consistency_level` is provided. It looks like a direct result of this change: https://github.com/milvus-io/pymilvus/commit/cb4cbc6e8672f11c40bfe3dd41ec6c08408ecfc5

The solution could be to change the `.get` to `.pop` but it's worth making sure it's not going to be a problem with the line 

```python
[...]

PR #2350, opened 2024-11-15
fix: cannot pass consistency level for delete

XuanYang-cn:
Also optinal variables default value should always be None.
And empty str is always invalid variable

See also: #2327

### DIFF
--- a/pymilvus/client/grpc_handler.py
+++ b/pymilvus/client/grpc_handler.py
@@ -596,17 +596,15 @@ class GrpcHandler:
         check_pass_param(collection_name=collection_name, timeout=timeout)
         try:
             req = Prepare.delete_request(
-                collection_name,
-                partition_name,
-                expression,
-                consistency_level=kwargs.get("consistency_level", 0),
-                param_name=kwargs.pop("param_name", None),
+                collection_name=collection_name,
+                expression=expression,
+                partition_name=partition_name,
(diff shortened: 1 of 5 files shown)


pymilvus/milvus_client/milvus_client.py
```diff
@@ -543,8 +543,8 @@ def delete(
         collection_name: str,
         ids: Optional[Union[list, str, int]] = None,
         timeout: Optional[float] = None,
-        filter: Optional[str] = "",
```
czs007, line 546: As you suggested, move the checks for IDs and filters to this layer.

pymilvus/client/grpc_handler.py
```diff
@@ -596,17 +596,15 @@ def delete(
...
+                filter=expression,
+                partition_name=partition_name,
+                consistency_level=kwargs.pop("consistency_level", 0),
                 **kwargs,
             )
             future = self._stub.Delete.future(req, timeout=timeout)
-
             if kwargs.get("_async", False):
-                cb = kwargs.get("_callback")
+                cb = kwargs.pop("_callback")
```
czs007, line 607: if "_callback" does not exist, it will raise a Exception.

czs007: cb = kwargs.pop("_callback", None) may be better.

czs007: I will change this in next pr.

PR #2387, opened 2024-11-28
fix: type annotation in Hits and docs in Collection

jiangyinzuo:


pymilvus/client/abstract.py
```diff
@@ -652,7 +652,7 @@ class Hits(list):
     def __init__(
         self,
         topk: int,
-        pks: Union[int, str],
+        pks: List[Union[int, str]],
```
jiangyinzuo, line 655: self.ids is `List[Union[str, int]]`, so pks should be `List[Union[str, int]]` too.

pymilvus/orm/collection.py
```diff
@@ -700,7 +700,7 @@ def search(
 
             partition_names (``List[str]``, optional): The names of partitions to search on.
             output_fields (``List[str]``, optional):
-                The name of fields to return in the search result.  Can only get scalar fields.
```
jiangyinzuo, line 703: `output_fields` can be vector fields too, see `examples/orm/hello_milvus.py`

PR #2396, opened 2024-11-29
fix: grant ManualCompact api doesn't work

weiliu1031:
issue: milvus-io/milvus#38086
cause RBAC require to check collection name, but ManualCompact rpc pass collection id in request, so grant ManualCompact api doesn't work.

This PR refine compact api impl to pass collection name in request.


XuanYang-cn (approved):
/lgtm

Issue #2338, opened 2024-11-08
[Enhancement]: alterindex & altercollection supports altering properties


JsDove:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What would you like to be added?

alterindex & altercollection supports altering properties

XuanYang-cn:
should be done

PR #2406, opened 2024-12-04
enhance: alterindex & altercollection supports altering properties

JsDove:
enhance :
alterindex delete properties
We have introduced a new parameter deleteKeys to the alterindex functionality, which allows for the deletion of properties within an index. This enhancement provides users with the flexibility to manage index properties more effectively by removing specific keys as needed.
altercollection delete properties
We have introduced a new parameter deleteKeys to the altercollection functionality, which allows for the deletion of properties within an collection. This enhancement provides users with the flexibility to manage collection properties more effectively by removing specific keys as needed.
3.support altercollectionfield
We currently support modifying the fieldparams of a field in a collection using altercollectionfield, which only allows changes to the max-length attribute.
Key Points:
New Parameter - deleteKeys: This new parameter enables the deletion of specified properties from an index. By passing a list of keys to deleteKeys, users can remove the corresponding properties from the index.
Mutual Exclusivity: The deleteKeys parameter cannot be used in conjunction with the extraParams parameter. Users must choose one parameter to pass based on their requirement. If deleteKeys is provided, it indicates an intent to delete properties; if extraParams is provided, it signifies the addition or update of properties.
issue :[#2338](https://github.com/milvus-io/pymilvus/issues/2338)


JsDove:
@czs007 help review

Issue #2418, opened 2024-12-09
[Enhancement]: Extend offset for QueryIterator


MrPresent-Han:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What would you like to be added?

Currently, input offset for query iterator can not be larger than 16384 which restricts users a lot. This issue is to cancel this limit

PR #2419, opened 2024-12-09
extend unlimted offset for query iterator(#2418)

MrPresent-Han:
related: #2418
pr: https://github.com/milvus-io/pymilvus/pull/2420


pymilvus/orm/iterator.py
```diff
@@ -37,7 +38,6 @@
     MILVUS_LIMIT,
     OFFSET,
     PARAMS,
-    PRINT_ITERATOR_CURSOR,
```
XuanYang-cn, line 40: Please remove log=logging.getLogger(__name__) in L55

and  LOGGER.setLevel(logging.INFO) in L51

MrPresent-Han: it's a separate file and I suppose I need to set up log as started?
log = logging.getLogger("pymilvus")
log.setLevel(logging.INFO)
changed at line 50~51

XuanYang-cn: `pymilvus/orm/iterator.py` is part of the source codes, so don't need to getLogger from pymilvus, just use getLogger(__name__) is enough, and donot setLevel in source code, this'll break settings outside of PyMilvus(such as log settings in examples)

Loggers for examples needs to getLogger("pymilvus")

XuanYang-cn (approved):
/lgtm

XuanYang-cn (approved):

PR #2438, opened 2024-12-12
fix pkg_resources api deprecated warning

zhuwenxing:
/kind improvement

### DIFF
--- a/pymilvus/client/__init__.py
+++ b/pymilvus/client/__init__.py
@@ -2,8 +2,7 @@ import logging
 import re
 import subprocess
 from contextlib import suppress
-
-from pkg_resources import DistributionNotFound, get_distribution
+from importlib.metadata import PackageNotFoundError, version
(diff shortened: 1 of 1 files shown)


zhuwenxing:
got warnings when using pymilvus
```
.venv/lib/python3.12/site-packages/pymilvus/client/__init__.py:6
  /Users/zilliz/workspace/milvus_beir/.venv/lib/python3.12/site-packages/pymilvus/client/__init__.py:6: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    from pkg_resources import DistributionNotFound, get_distribution

```

this pr is to fix this warning

djwessel:
@XuanYang-cn When do we plan for this to find it's way into 2.x? we have now had at least 3 releases of 2.5.x since this fix

PR #2461, opened 2024-12-17
enhance: support milvus-client iterator

MrPresent-Han:
related: #2464


pymilvus/milvus_client/milvus_client.py
```diff
@@ -480,6 +482,84 @@ def query(
...
+        batch_size: Optional[int] = 1000,
+        limit: Optional[int] = UNLIMITED,
+        filter: Optional[str] = "",
+        output_fields: Optional[List[str]] = None,
+        partition_names: Optional[List[str]] = None,
+        timeout: Optional[float] = None,
+        **kwargs,
+    ):
+        if filter is not None and not isinstance(filter, str):
+            raise DataTypeNotMatchException(message=ExceptionsMessage.ExprType % type(filter))
```
MrPresent-Han, line 497: can we directly depend on orm.iterator, which is the simpliest way to implement

PR #2487, opened 2024-12-20
fix: remove async function in AsyncClient connection initialization

brcarry:
- Move check for channel ready and identifier interceptor setup before each query call instead of during the AsyncClient connection initialization

### DIFF
--- a/pymilvus/client/async_grpc_handler.py
+++ b/pymilvus/client/async_grpc_handler.py
@@ -10,7 +10,7 @@ from urllib import parse
 import grpc
 from grpc._cython import cygrpc
 
-from pymilvus.decorators import ignore_unimplemented, retry_on_rpc_failure, upgrade_reminder
+from pymilvus.decorators import ignore_unimplemented, retry_on_rpc_failure
 from pymilvus.exceptions import (
     AmbiguousIndexName,
     DescribeCollectionException,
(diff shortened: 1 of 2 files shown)


XuanYang-cn (changes requested):
Please try a class decorator to reduce repeating codes like this:
```python
import asyncio

class AsyncPreMethodDecoratorMeta(type):
    async def __new__(mcs, name, bases, attrs):
        for attr_name, attr_value in attrs.items():
            if asyncio.iscoroutinefunction(attr_value):
                async def wrapper(*args, **kwargs):
                    # Your asynchronous pre-method logic here
                    print(f"Before {attr_name}")
                    return await attr_value(*args, **kwargs)
                attrs[attr_name] = wrapper
        return type.__new__(mcs, name, bases, attrs)

class AsyncMyClass(metaclass=AsyncPreMethodDecoratorMeta):
    async def async_method1(self):
        print("Async Method 1")

    async def async_method2(self):
        print("Async Method 2")

async def main():
    obj = AsyncMyClass()
    await obj.async_method1()
    await obj.async_method2()

asyncio.run(main())
```

XuanYang-cn:
This is not a blocking request change
/lgtm
/approve

XuanYang-cn (approved):

PR #2513, opened 2024-12-25
enhance: Enable resource group api in milvus client

weiliu1031:
issue: milvus-io/milvus#38739
this PR enable those API in milvus client:
- create_resource_group
- update_resource_groups
- drop_resource_group
- describe_resource_group
- list_resource_groups
- transfer_replica

cause the resource group prefer to use declarative api, so the transfer_node API is deprecated in milvus client, it's recommand to use update_resource_group to manage node num in resource group.

PR #2514, opened 2024-12-26
enhance: Enable resource group api in milvus client (#2513)

weiliu1031:
issue: milvus-io/milvus#38739
pr: #2513
this PR enable those API in milvus client:
- create_resource_group
- update_resource_groups
- drop_resource_group
- describe_resource_group
- list_resource_groups
- transfer_replica

cause the resource group prefer to use declarative api, so the transfer_node API is deprecated in milvus client, it's recommand to use update_resource_group to manage node num in resource group.

### DIFF
--- a/examples/resource_group.py
+++ b/examples/resource_group.py
@@ -1,0 +1,105 @@
+from pymilvus import (
+    MilvusClient,
+    DataType,
+)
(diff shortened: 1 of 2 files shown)


XuanYang-cn (approved):
/lgtm
/approve

PR #2551, opened 2025-01-08
enhance: Add schema update time verification to insert and upsert to use cache

JsDove:
enhance: Add schema update time verification to insert and upsert to use cache
issue: https://github.com/milvus-io/milvus/issues/39093


pymilvus/client/abstract.py
```diff
@@ -247,6 +247,7 @@ def dict(self):
             "properties": self.properties,
             "num_partitions": self.num_partitions,
             "enable_dynamic_field": self.enable_dynamic_field,
+            "update_timestamp": self.update_timestamp,
```
czs007, line 250: remove this

Issue #2555, opened 2025-01-10
[Bug]: Fix iterator context change


MrPresent-Han:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

when milvus alias change, iterator should switch context accordingly

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

PR #2556, opened 2025-01-10
fix: check collection id match for possible alias switch

MrPresent-Han:
related: #2555


XuanYang-cn:
@MrPresent-Han Please rebase the latest master to fix the code checker

Issue #2558, opened 2025-01-13
[Bug]: Pymilvus erroneously presumes the primary key to be named "id" when deal with SearchResults


czs007:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

 Pymilvus erroneously presumes the primary key to be named "id" when deal with SearchResults

### Expected Behavior

Determine the name of the primary field based on the primary_field_name in the returned results.

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

PR #2559, opened 2025-01-13
fix: use correct primary field name in Hits

czs007:
issue: https://github.com/milvus-io/pymilvus/issues/2558


XuanYang-cn (approved):
/lgtm

PR #2569, opened 2025-01-15
fix: use dummy call for search iterator v2 compatibility

PwzXxm:
issue: https://github.com/milvus-io/milvus/issues/39117


XuanYang-cn (approved):
/lgtm

Issue #2601, opened 2025-01-24
[Bug]: Not providing auto ID field in upsert request raises an error "Insert missed an field `id` to collection without set nullable==true or set default_value"


issamemariold:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Not providing auto ID field in upsert request raises an error "Insert missed an field `id` to collection without set nullable==true or set default_value"

### Expected Behavior

The upsert request should not fail because the ID field is configured to be `auto_id`

### Steps/Code To Reproduce behavior

```python
from pymilvus import CollectionSchema, DataType, FieldSchema
from pymilvus.milvus_client.index import IndexParams

client = MilvusClient(
    uri=<SECRET>,
    token=<SECRET>,
)

dimension = 4
collection_name = "testing_collection"
client.create_collection(
    collection_name,
    schema=CollectionSchema(
        fields=[
            FieldSchema(
                name="id",
                dtype=DataType.INT64,
                is_primary=True,
                auto_id=True,
                nullable=True,
                description="Auto increment ID",
            ),
            FieldSchema(
                name="vector",
                dtype=DataType.FLOAT_VECTOR,
                dim=dimension,
                description="Embedding vector",
            ),
        ],
    ),
)

client.create_index(
    collection_name,
    index_params=IndexParams(
        field_name="vector",
        index_type="IVF_FLAT",
        metric_type="COSINE",
    ),
)

client.load_collection(collection_name)

client.upsert(
    collection_name,
    data=[{"vector": [i, i, i, i]} for i in range(10)],
)
```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory): MacOS
- Method of installation (Docker, or from source): pip install
- Milvus version (v0.3.1, or v0.4.0): pymilvus==2.5.3
```

### Anything else?

The issue seems to be coming from this line

https://github.com/milvus-io/pymilvus/blob/c5a3c18be2fb4dde06167f05af2761456af0e2d4/pymilvus/client/prepare.py#L562

Here it should be checked whether the field is `auto_id` and allow it not to be provided

issamemari:
I'm proposing a fix in this PR https://github.com/milvus-io/pymilvus/pull/2602

issamemari:
@yanliang567 hmm. I would have expected this to run fine. To upsert is to update or insert. If the ID is not provided it should be assumed that this is an insert request.

[...]

PR #2602, opened 2025-01-24
Do not raise error if auto_id field is not provided

issamemari:
Fixes issue https://github.com/milvus-io/pymilvus/issues/2601


XuanYang-cn:
This is by design according to the issue,  closing this PR now.

PR #2603, opened 2025-01-29
Make milvus-lite a optional dependency

SuperSecureHuman:
Right now, milvus lite is a required dependency. When trying to use it under docker, milvus-lite alone uses ~150MB.

This space is mostly unwanted if we dont use local DB (using localdb is unlikely in docker deployments)

### DIFF
--- a/pymilvus/orm/connections.py
+++ b/pymilvus/orm/connections.py
@@ -385,9 +385,14 @@ class Connections(metaclass=SingleInstanceMetaClass):
                     message=f"Open local milvus failed, dir: {parent_path} not exists"
                 )
 
-            from milvus_lite.server_manager import (
-                server_manager_instance,
-            )
+            try:
+                from milvus_lite.server_manager import (
+                    server_manager_instance,
+                )
+            except ImportError:
+                raise ConnectionConfigException(
+                    message="milvus-lite is not installed. Please install it using 'pip install pymilvus[milvus_lite]' or pip install milvus-lite"
+                )
 
             local_uri = server_manager_instance.start_and_get_uri(kwargs["uri"])
             if local_uri is None:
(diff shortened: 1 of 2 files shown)


XuanYang-cn:
@SuperSecureHuman Usually a system can run Milvus cluster or Milvus standalone is able to host ~200MB of PyMilvus, cuz 200MB is nothing comparing to Milvus. Also, if resources are limited, Milvus-lite is the best choice. 

So why are you using Milvus but concerning PyMilvus disk usage? Also mounting volumes might be a solution to use milvus-lite inside a docker.

SuperSecureHuman:
It's when using pymilvus connect to an existing cluster..

The node does not run milvus, rather connects to milvus.

XuanYang-cn:
@SuperSecureHuman Hi, THX for the reply. We want milvus-lite to be part of PyMilvus by default so I labelled this PR `wontfix`. In this case I suggest you to clone pymilvus repo and use your own build of it if you insist, it's quiet easy for python:

1. fork and remove the milvus-lite requirement
2. git clone your pymilvus repo in docker
3. just run `pip install -e .` or `make install` at the project root

**OR, a even simpler choice:**
1. fork and remove the milvus-lite requirement in your own repo
2. install from remote resource with this command: `pip install git+ssh://github.com/{YOUR-REPO}/pymilvus.git@master`

XuanYang-cn:
BTW you could try this in your Dockerfile: `pip install git+ssh://github.com/SuperSecureHuman/pymilvus.git@master`

SuperSecureHuman:
Hi,

So the current workaround that I am doing is, installing pymilvus without deps, and installing the needed deps manually.. In our case, this alone gives us the required result (and ~20% reduction in image size)


You may close the issue if needed, this can be a suggested way if someone wants to setup their minimal docker images.

XuanYang-cn:
Thanks for the contribution anyway!

PR #2639, opened 2025-02-17
feat: add external filter func for search iterator v2

PwzXxm:
issue: https://github.com/milvus-io/milvus/issues/39914


XuanYang-cn:
#2640

PR #2653, opened 2025-02-18
Fix bug for BulkWriter with nullable/default_value fields

yhmo:
https://github.com/milvus-io/pymilvus/issues/2654


XuanYang-cn (approved):
/lgtm

Issue #2626, opened 2025-02-10
[Bug]: EF inconsistency after enabling search params flatten


MrPresent-Han:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Steps/Code To Reproduce behavior

```markdown

```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

XuanYang-cn:
should be fixed

PR #2669, opened 2025-02-21
fix: remove set batchsize to ef for param flat(#2626)

MrPresent-Han:
related: #2626
pr: https://github.com/milvus-io/pymilvus/pull/2627


XuanYang-cn (approved):
/lgtm
/approve

Issue #2676, opened 2025-03-03
[FEATURE]: Enable to create index on dynamic field


XuanYang-cn:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

See also: https://github.com/milvus-io/milvus/issues/35528
This check may not needed at all, just use checks from proxy. https://github.com/milvus-io/pymilvus/blob/af2f6df35aba379e2af5e78ecfd13f67738a0c4c/pymilvus/client/grpc_handler.py#L1038

XuanYang-cn:
See also: #2685

PR #2677, opened 2025-03-04
fix: Remove field check in create_index

sunby:


XuanYang-cn:
See also: #2676

XuanYang-cn:
@sunby Please fix the coding style

PR #2680, opened 2025-03-05
fix: hybrid search, SequenceIterator is not iterable

Benjacho:
This fixes the issue [#40391](https://github.com/milvus-io/pymilvus/issues/2681)

By just doing the same as the search method, instead of using the list comprehension syntax.


QMJ602:
I also encountered this problem. How did you solve it?

PR #2685, opened 2025-03-11
fix: Remove field check in create_index (#2677)

sunby:
pr: https://github.com/milvus-io/pymilvus/pull/2677


XuanYang-cn (approved):
/lgtm

PR #2725, opened 2025-04-14
doc: add time unit for constructor

proost:
It would be more helpful what is unit of timeout in constructor


XuanYang-cn (approved):
/lgtm

Issue #2727, opened 2025-04-15
[Bug]: if db_name is passed as a path in the uri it is overwritten with empty string


mihailyanchev:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

This is related to this [issue](https://github.com/milvus-io/pymilvus/issues/2670) and its fix, which I implemented.

The original issue is fixed, but the fix introduces another bug - if you pass a uri to the `MilvusClient` or `AsyncMilvusClient` that contains the db_name as a path (`http://localhost:19530/test_db`) and you do not explicitly pass the db_name argument, it will get overwritten with an empty string.

The issue is here in this line: https://github.com/milvus-io/pymilvus/blob/master/pymilvus/orm/connections.py#L458 
we should have `group[0]` instead of `group[1]`.

Sorry for overlooking this.

### Expected Behavior

When you pass the db_name as a path in the uri and not as an explicit argument, the client should connect to the provided db_name.

### Steps/Code To Reproduce behavior

```
"""
Put a break point here and check the value of group: https://github.com/milvus-io/pymilvus/blob/master/pymilvus/orm/connections.py#L458
"""

import asyncio

from pymilvus import AsyncMilvusClient


async def main():
    _ = AsyncMilvusClient(
        uri="http://localhost:19530/test_db",
    )


asyncio.run(main())
```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0): 2.5.6
- Milvus configuration (Settings you made in `server_config.yaml`):
```

XuanYang-cn:
Welcome for any fixes.

mihailyanchev:
@XuanYang-cn here is a PR for it: https://github.com/milvus-io/pymilvus/pull/2734

XuanYang-cn:
Shoud be fixed in PyMilvus 2.5.7, closing it

XuanYang-cn:
should be fixed already

PR #2734, opened 2025-04-17
fix: db_name will be overwritten by mistake

mihailyanchev:
Fix for [2727](https://github.com/milvus-io/pymilvus/issues/2727). Ensures that both passing a db_name as an explicit argument and as a path to the connection uri is used and not overwritten with an empty string. Added a couple of test cases.


XuanYang-cn (approved):
/lgtm

PR #2738, opened 2025-04-21
fix: [25]db_name will be overwritten by mistake (#2734)

XuanYang-cn:
Fix for [2727](https://github.com/milvus-io/pymilvus/issues/2727). Ensures that both passing a db_name as an explicit argument and as a path to the connection uri is used and not overwritten with an empty string. Added a couple of test cases.


XuanYang-cn:
Group[0] might be address, holding for tests
/hold

XuanYang-cn:
See #2954

PR #2740, opened 2025-04-21
enhance: [25]Use SearchResult in MilvusClient (#2735)

XuanYang-cn:
SearchResult's type is SearchResult[Hits[hit]], which is List[List[Hit]].

This PR changes `Hit` into `Dict`. So the SearchResult type becomes List[List[Dict]] to match MilvusClient's return type.

This avoids extra copy in the MilvusClient search to enhance performance.
The usage of MilvusClient.search and orm.Collection.search keeps unchange.


pymilvus/client/search_reasult.py
```diff
@@ -0,0 +1,480 @@
...
+
+        >>> res.extra
+        {"cost": 1}
+
+    Attributes:
+        recalls(List[float], optional): The recalls of the search result, one for each query.
+        extra(Dict, optional): The extra information of the search result.
+    """
+
+    recalls: Optional[List[float]] = None
```
czs007, line 39: Why was it changed to class member?

czs007 (approved):

PR #2762, opened 2025-04-27
Fix hybrid search params check

junjiejiangjjj:
- Solve the problem that the `expr_params` parameter of `async hybrid_search` is not effective
- Solve the problem of rerank parameter verification


XuanYang-cn (approved):
/lgtm

PR #2766, opened 2025-04-28
fix: Pass request-id in metadata per API

XuanYang-cn:
client-request-id is used to trace each API, and unique for each API. The previous code resets the channel for every API call to set it into global channel interceptor, which has great performance down and also will rewrite each other.

This PR adds client-request-id to per API metadata, and remove some of
the future calls.

See also: milvus-io#2761, milvus-io#2765


XuanYang-cn:
@mergifyio refresh

czs007 (approved):

Issue #2781, opened 2025-05-08
[Bug]: num_replica is set to 1 by default


XuanYang-cn:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

The default values should be defined by Milvus server side, not by client side.

### Steps/Code To Reproduce behavior

```markdown

```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

XuanYang-cn:
See also milvus-io/milvus#41673

PR #2782, opened 2025-05-08
fix: remove default replica_number in load_collection

XuanYang-cn:
The default value should be decided in the server side

See also: #2781, milvus-io/milvus#41673


XuanYang-cn:
/hold
for load_partitions

czs007 (approved):

PR #2716, opened 2025-03-31
avoid collision on aliases in MilvusClient _create_connection

jperez999:
This PR handles the #2715 

It ensures that only one alias is used when calling connections.connect within the milvus client. This will allow users to create milvus connections to multiple milvus instances in the same context without worrying about collisions on aliases.

Issue #2715, opened 2025-03-31
[Bug]: Cannot create connections to  different milvus instances


jperez999:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

When I try to create two connections to separate instances of milvus, in the same execution, I get errors.

### Expected Behavior

I expect to be able to make both connections successfully.

### Steps/Code To Reproduce behavior

When I execute the following code:
```
import uuid
from pymilvus import (
    MilvusClient,
    connections
)

a_name = str(uuid.uuid4())[:8]
collection_name = "test_collection"
milvus_uri = "http://milvus:19530"
connections.connect(alias=a_name, uri=milvus_uri)
client_1 = MilvusClient(milvus_uri, alias=a_name)

a_name = str(uuid.uuid4())[:8]
collection_name = "test_collection"
milvus_uri = "http://10.57.202.36:19530"
connections.connect(alias=a_name, uri=milvus_uri)
client_2 = MilvusClient(milvus_uri, alias=a_name)
```
I get this error:
```
Traceback (most recent call last):
  File "/opt/conda/envs/nv_ingest_runtime/lib/python3.10/pdb.py", line 1723, in main
    pdb._runscript(mainpyfile)
  File "/opt/conda/envs/nv_ingest_runtime/lib/python3.10/pdb.py", line 1583, in _runscript
    self.run(statement)
  File "/opt/conda/envs/nv_ingest_runtime/lib/python3.10/bdb.py", line 598, in run
    exec(cmd, globals, locals)
  File "<string>", line 1, in <module>
  File "/raid/workspace/nv-ingest/client/client_examples/examples/index_test.py", line 12, in <module>
    client_1 = MilvusClient(milvus_uri, alias=a_name)
  File "/opt/conda/envs/nv_ingest_runtime/lib/python3.10/site-packages/pymilvus/milvus_client/milvus_client.py", line 66, in __init__
    self._using = self._create_connection(
  File "/opt/conda/envs/nv_ingest_runtime/lib/python3.10/site-packages/pymilvus/milvus_client/milvus_client.py", line 921, in _create_connection
    raise ex from ex
  File "/opt/conda/envs/nv_ingest_runtime/lib/python3.10/site-packages/pymilvus/milvus_client/milvus_client.py", line 918, in _create_connection
    connections.connect(using, user, password, db_name, token, uri=uri, **kwargs)
TypeError: Connections.connect() got multiple values for argument 'alias'

```

[...]

PR #2799, opened 2025-05-21
fix: [cp25]avoid collision on aliases in MilvusClient (#2716)

XuanYang-cn:
This PR handles the #2715

It ensures that only one alias is used when calling connections.connect within the milvus client. This will allow users to create milvus connections to multiple milvus instances in the same context without worrying about collisions on aliases.


czs007 (approved):
/lgtm

Issue #2794, opened 2025-05-15
[Bug]: Strange KeyError('__pydantic_validator__') with pymilvus 2.5.7 and up


stronk7:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

I'm using pymilvus within a langgraph graph (I don't use langchain own milvus client but pymilvus).

Basically what I've is a simple graph with retrieve, rank and generate nodes (simple RAG).

As part of the retrieve node, I execute an hybrid search and return the list of retrieved docs (instances of `Hit`:

```
...
...
retrieved_docs = milvus.hybrid_search( ... )
milvus.close()
return {"vector_search": retrieved_docs[0]}
```

Then, the graph continues its execution with the rest of nodes till generation.

Up to pymilvus 2.5.6 everything was looking ok, but with 2.5.7 and 2.5.8 I've started getting some "strange" warnings in logs from some code within Langgraph.

```
WARNING [langchain_core.callbacks.manager] Error in StreamMessagesHandler.on_chain_end callback: KeyError('__pydantic_validator__')
```

Over the paper, those `Hit` objects are ok, but I bet that there is something with them that causes some validation problem at some point. In previous versions, those retrieved docs where plain dictionaries, not `Hit` instances.

I've tried enabling debug logging, but that's all the information I've been able to collect, I'm afraid.

That said, everything seems to continue working ok, I've traced here and all the results are being passed and used for generation as expected, but I think that those warnings may be an indication of something not being 100% ok, specially when integrating with other systems.

Hope it helps, ciao :-)

### Expected Behavior

the retrieved docs in a search can be passed to other systems (langgraph) without any problem, not causing the problems / warnings described above.

### Steps/Code To Reproduce behavior

```markdown
In any basic langgraph RAG graph (retrieve + generate), use pymilvus to perform a search and return the retrieved docs in the state to the generation node.
```

### Environment details

```markdown
- Hardware/Softward conditions (MacOS 15.4, 64GB):
- Method of installation (pip installed pymilvus):
- Milvus version (v2.5.10, running on Docker, with docker compose default file):
- Milvus configuration (Settings you made in `server_config.yaml`): Nothing.
```

XuanYang-cn:
Looks weird, I'll look into it.

XuanYang-cn:
releases with 2.5.11

stronk7:
Cool, thanks, will try soon!

stronk7:
Uhm,

[...]

PR #2806, opened 2025-05-21
Fixes hasattr checks against Hit for missing attributes

michaelkl:
## The problem
`hasattr` calls against `Hit` instances fail with `KeyError`.

### How to replicate
```python
>>> res = {
...   "my_id": 1,
...   "distance": 0.3
... }
>>> h = Hit(res, pk_name="my_id")

>>> hasattr(h, 'my_id')
True

>>> hasattr(h, '__pydantic_decorators__')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.12/site-packages/pymilvus/client/search_result.py", line 389, in __getattr__
    return self.__getitem__(item)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/pymilvus/client/search_result.py", line 425, in __getitem__
    return self.data["entity"][key]
           ~~~~~~~~~^^^^^^^^^^
KeyError: 'entity'
```

## Root cause

The current implementation of `Hit.__getattr__` raises `KeyError` when a requested key does not exist. But the only exception that is expected is `AttributeError` (see [`hasattr` docs](https://docs.python.org/3/library/functions.html#getattr)).

This is especially crucial because `pydantic` library uses `hasattr(instance, '__pydantic_decorators__')` call for its needs. And I was “lucky” to get there when I tried to use `llama-index` instrumentation with data originating from Milvus.

### DIFF
--- a/pymilvus/client/search_result.py
+++ b/pymilvus/client/search_result.py
@@ -392,7 +392,10 @@ class Hit(UserDict):
         if item == "entity":
             return self
 
-        return self.__getitem__(item)
+        try:
+            return self.__getitem__(item)
+        except KeyError as exc:
+            raise AttributeError from exc
 
     def to_dict(self) -> Dict[str, Any]:
         """Patch for orm, will be deprecated soon"""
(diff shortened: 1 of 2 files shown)


XuanYang-cn:
Fixes: #2794

XuanYang-cn (approved):
/lgtm

PR #2790, opened 2025-05-13
fix: wrong marshal field schema prarams case alter collection failed

aoiasd:
relate: https://github.com/milvus-io/milvus/issues/41804

PR #2811, opened 2025-05-23
fix:[2.5] wrong marshal field schema prarams case alter collection failed

aoiasd:
pr: https://github.com/milvus-io/pymilvus/pull/2790
relate: https://github.com/milvus-io/milvus/issues/41804


XuanYang-cn (approved):
/lgtm

PR #2816, opened 2025-05-28
fix: MilvusClient misusage of disconnect

XuanYang-cn:
To remove a one-time connection, we need to call
remove_connection instead of disconnect


czs007 (approved):
/lgtm

Issue #2839, opened 2025-06-06
[Enhancement]: reduce describecollection calls in compact method


JsDove:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

reduce describecollection calls in compact method

PR #2840, opened 2025-06-06
fix: add retry  compact for for compatibility with old Milvus server

JsDove:
add retry  compact for for compatibility with old Milvus server in order to reduce describecollection
issue: https://github.com/milvus-io/pymilvus/issues/2839


XuanYang-cn (approved):
/lgtm

Issue #2848, opened 2025-06-10
[Enhancement]: Optimize perf with large topk


MrPresent-Han:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Under the case where users' topk is very large, the bottleneck introduced by sdk is very obvious

PR #2866, opened 2025-06-19
fix: hybridExtraList miss iter method and dynamic fields value overrite fixed field value(#2848)

MrPresent-Han:
related: #2848
pr: https://github.com/milvus-io/pymilvus/pull/2862


XuanYang-cn (approved):
/lgtm

PR #2872, opened 2025-06-23
fix: Recursively retry dml op if schema mismatch detected

congqixia:
Related to milvus-io/milvus#41858

`SchemaMismatch` could be returned if multiple schema change happens in a very short period of time. The retry attempt could fail when a new schema change happened.

This PR make retry for schema mismatch case happens in a recursive way handling previouly described case.

### DIFF
--- a/pymilvus/client/grpc_handler.py
+++ b/pymilvus/client/grpc_handler.py
@@ -573,11 +573,14 @@ class GrpcHandler:
         resp = self._stub.Insert(request=request, timeout=timeout, metadata=_api_level_md(**kwargs))
         if resp.status.error_code == common_pb2.SchemaMismatch:
             schema = self.update_schema(collection_name, timeout, **kwargs)
-            request = self._prepare_row_insert_request(
-                collection_name, entities, partition_name, schema, timeout, **kwargs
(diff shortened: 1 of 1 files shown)


pymilvus/client/grpc_handler.py
```diff
@@ -573,11 +573,14 @@ def insert_rows(
...
-                request=request, timeout=timeout, metadata=_api_level_md(**kwargs)
+            # recursively calling `insert_rows` handling another schema change happens during retry
+            return self.insert_rows(
+                collection_name=collection_name,
+                entities=entities,
+                partition_name=partition_name,
+                schema=schema,
+                timeout=timeout,
+                **kwargs,
             )
```
XuanYang-cn, line 584: Maybe create a new exception like "class SchemaMisMatch(MilvusException)" and raise it here, and then catch it at decorators.py and retry there

XuanYang-cn (approved):
/lgtm

pymilvus/client/grpc_handler.py
congqixia: will do in next pr

PR #2873, opened 2025-06-24
enhance: reuse channel by default

tedxu:


tedxu:
My local test shows that only 1 connection is used when creating 100 clients, after this PR applied:

```
============================================================
MILVUS SERVER PERFORMANCE BENCHMARK REPORT
============================================================
Total Duration: 38.09 seconds
Total Operations: 10000
Throughput: 262.51 ops/sec
Active Connections: 1
Clients: 100
Timestamp: 2025-06-24T14:21:30.032776

Operation: has_collection
  Count: 10000
  Success Rate: 100.00%
  Error Count: 0
  Timing (ms):
    Mean: 334.99
    Median: 287.82
    Min: 192.38
    Max: 992.72
    Std Dev: 123.36
    95th Percentile: 576.57
    99th Percentile: 713.30
```

PR #2878, opened 2025-06-27
[fix]: Fix compaction params error in async grpc handler & [Add]: Previledge async test & resource group test

Shawnzheng011019:


XuanYang-cn (approved):
/lgtm

PR #2893, opened 2025-07-15
support stage operation

lentitude2tk:


pymilvus/stage/file_utils.py
```diff
@@ -0,0 +1,42 @@
...
+        result = []
+        total_size = 0
+
+        for root, _, files in os.walk(folder_path):
+            for file in files:
+                full_path = Path(root) / file
+                result.append(str(full_path.resolve()))
+                total_size += full_path.stat().st_size
+
+        return result, total_size
```
XuanYang-cn: It's recommanded to take full advantage of pathlib to support windows platform. There're two things to be cared of:
1. local path should always use pathlib.Path instead of str
2. Remote path(endpoint or url) should always use PosixPath.

For example In your Code `Path(root) / file` only fits in posix-like system. Windows users are going to complain about it.

If you don't intend to support windows platform, Please doc or comment it properly.

pymilvus/stage/stage_operation.py
```diff
@@ -0,0 +1,148 @@
...
+
+import urllib3
+from minio import Minio
+from minio.error import S3Error
+
+from pymilvus.stage.file_utils import FileUtils
+from pymilvus.stage.stage_restful import apply_stage
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger("stage_operation")
```
XuanYang-cn: Please use `logging.getLogger(__name__)` to inherientent the logging configs from pymilvus. 

__name__ in here is equal to "pymilvus.stage.stage_operation"

pymilvus/stage/stage_restful.py
```diff
@@ -0,0 +1,117 @@
+import json
+import logging
+from typing import Optional, Union
+
+import requests
+
+from pymilvus.exceptions import MilvusException
+
+logger = logging.getLogger("stage_restful")
```
XuanYang-cn: ditto

XuanYang-cn:
Please add stages's requirements in pyproject.toml.

pymilvus/stage/stage_restful.py
lentitude2tk: ok, i use logging.getLogger(__name__) replace

pymilvus/stage/stage_operation.py
lentitude2tk: ok, i use logging.getLogger(name) replace

pymilvus/stage/file_utils.py
lentitude2tk: Adjusted as required

lentitude2tk:
> Please add stages's requirements in pyproject.toml.
ok, i will do it

XuanYang-cn (approved):
/lgtm

Issue #2886, opened 2025-07-08
[Bug]: JSON data type checking contains misleading errors.


zhuwenxing:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

```
        for i in range(nb):
            # Generate diverse numeric values
            if i % 5 == 0:
                # Some large numbers
                numeric_value = rng.integers(100000, 999999)
            elif i % 3 == 0:
                # Some medium numbers
                numeric_value = rng.integers(1000, 9999)
            else:
                # Regular sequential numbers
                numeric_value = i
            
            nested_value = i * 7 + (i % 13)  # Different pattern for nested
            
            all_numbers.append(numeric_value)
            all_nested_numbers.append(nested_value)
            
            rows.append({
                default_primary_key_field_name: i,
                default_vector_field_name: list(rng.random((1, default_dim))[0]),
                "json_field": {
                    "number": numeric_value,
                    "index": i,
                    "data": {"nested_number": nested_value}
                }
            })
        
        self.insert(client, collection_name, rows)
```

[2025-07-08 11:58:21 - ERROR - ci_test]: (api_response) : <DataNotMatchException: (code=1, message=The Input data type is inconsistent with defined schema, {json_field} field should be a json, but got a {<class 'dict'>} instead.)> (api_request.py:36)

The root cause of the insert error is likely due to the presence of numpy data types in the JSON data, which cannot be serialized into JSON. However, it triggered a misleading error instead. Our JSON datatype supports the dict class, but this error can cause confusion.

### Steps/Code To Reproduce behavior

```markdown

```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

XuanYang-cn:
Should be fixed

Issue #2838, opened 2025-06-05
[Bug]: After Milvus recovers from standalone pod kill chaos, the connection still keeps reporting errors.


zhuwenxing:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

pymilvus version: pymilvus-2.6.0rc136

```
[2025-06-05T04:14:37.417Z] >>, message=[query] Retry timeout: 10s, message=failed to connect to all addresses; last error: UNKNOWN: ipv4:10.255.117.119:19530: Failed to connect to remote host: Timeout occurred: FD Shutdown)>, <Time:{'RPC start': '2025-06-05 04:14:24.430398', 'RPC error': '2025-06-05 04:14:37.102152'}> (decorators.py:140)
[2025-06-05T04:14:37.417Z] [2025-06-05 04:14:37 - ERROR - ci_test]: Traceback (most recent call last):
[2025-06-05T04:14:37.417Z]   File "/home/jenkins/agent/workspace/tests/python_client/utils/api_request.py", line 22, in inner_wrapper
[2025-06-05T04:14:37.417Z]     res = func(*args, **_kwargs)
[2025-06-05T04:14:37.417Z]   File "/home/jenkins/agent/workspace/tests/python_client/utils/api_request.py", line 53, in api_request
[2025-06-05T04:14:37.417Z]     return func(*arg, **kwargs)
[2025-06-05T04:14:37.417Z]   File "/usr/local/lib/python3.10/dist-packages/pymilvus/orm/collection.py", line 1083, in query
[2025-06-05T04:14:37.417Z]     return conn.query(
[2025-06-05T04:14:37.417Z]   File "/usr/local/lib/python3.10/dist-packages/pymilvus/decorators.py", line 141, in handler
[2025-06-05T04:14:37.417Z]     raise e from e
[2025-06-05T04:14:37.417Z]   File "/usr/local/lib/python3.10/dist-packages/pymilvus/decorators.py", line 137, in handler
[2025-06-05T04:14:37.417Z]     return func(*args, **kwargs)
[2025-06-05T04:14:37.417Z]   File "/usr/local/lib/python3.10/dist-packages/pymilvus/decorators.py", line 173, in handler
[2025-06-05T04:14:37.417Z]     return func(self, *args, **kwargs)
[2025-06-05T04:14:37.417Z]   File "/usr/local/lib/python3.10/dist-packages/pymilvus/decorators.py", line 92, in handler
[...]

Issue #2824, opened 2025-05-29
[FEATURE]: Please upgrade grpcio compatibility to support >=1.71.0


liubaiand:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Description

Currently, `pymilvus` (e.g., version 2.5.10) restricts `grpcio` to versions <=1.67.1.

This causes dependency conflicts when integrating with libraries that require `grpcio>=1.71.0`, such as:

- `grpcio-status==1.71.0`
- `opentelemetry-exporter-otlp-proto-grpc`
- Other Google Cloud components like `google-cloud-pubsub`

These libraries are widely used in production and rely on newer gRPC features and security updates.

## Suggested Fix

Please consider updating the `grpcio` dependency in your `setup.py` or `pyproject.toml` to support:
grpcio>=1.49.1,<2.0.0
Or even better (if compatible):
grpcio>=1.71.0,<2.0.0

This would help avoid dependency resolution failures in modern Python environments and Docker builds using pip's 2020+ resolver.

## Environment Example

pymilvus==2.5.10
grpcio-status==1.71.0
grpcio==1.67.1  # forced, causes ResolutionImpossible

Thanks 🙏
Thanks for the great work on pymilvus! Let me know if you need help testing with the newer grpcio versions.

XuanYang-cn:
related to #2358 
Thanks for the advise.

I want to but it didn't work out well. This issue is still on the way to be fixed on grpc version >= 1.68.0. So I setup a upper limit to minimize the impact on our users on production. https://github.com/milvus-io/pymilvus/issues/2358

Any advise to solve this issue would be warmly welcomed. It's infecting our Milvus tests as well.

Olimoyo:
Any updates or work arounds for this? I also have other packages that require `grpcio>=1.71.0`.

XuanYang-cn:
@Olimoyo We finally made some progress on this issue. It's planed to be fixed in the next pymilvus release v2.5.13

XuanYang-cn:
Should be fixed

Issue #2358, opened 2024-11-19
[Bug]: After grpcio is upgraded to 1.68.0, multi-threaded concurrent dql request will report error `failed to connect to all address`


wangting0128:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

grpcio==1.68.0

![img_v3_02gp_ef3905a1-c54c-4ad1-8043-5d6e06f34deg](https://github.com/user-attachments/assets/572e6852-b9e0-4d74-abab-5fa0e4f22d4e)

grpcio==1.60.0 not report connection error

fouram benchmark test case `test_concurrent_locust_ivf_sq8_search_cluster` can reproduce this issue

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

XuanYang-cn:
upgrading grpc version of Milvus now

doncat99:
how is going on this issue? grpcio is now release to 1.70.0 already [newest version](https://pypi.org/project/grpcio/)

can I override to 1.68.1 without report error?

```
dependencies=[
    "setuptools>69",
    "setuptools<70.1;python_version<='3.8'",
    "grpcio>=1.49.1,<=1.67.1", #issue https://github.com/milvus-io/pymilvus/issues/2358
    "protobuf>=3.20.0",
    "python-dotenv>=1.0.1, <2.0.0",
    "ujson>=2.0.0",
    "pandas>=1.2.4",
    "numpy<1.25.0;python_version<='3.8'",
    "milvus-lite>=2.4.0;sys_platform!='win32'",
]
```

XuanYang-cn:
Might related: https://github.com/milvus-io/pymilvus/pull/2684

XuanYang-cn:
I tested on Milvus with grpc-go==1.68.1 and it's not working. From issues on grpc repo, this's a bug introduced in grpc-cpp and grpc-python[grpcio]. I'm still investigating in how to avoid this issue when using grpcio>=1.68.0.

The issue in grpc repo was idled since March. The worst temp solution would be to never use grpcio greater than 1.68.0 just like PyMilvus 2.5.x.

Ref: grpc/grpc#39021, grpc/grpc#38282, grpc/grpc#38290

XuanYang-cn:
[...]

PR #2906, opened 2025-07-22
fix: Multiple cherry-picks from master

XuanYang-cn:
- [fix]: Fix compaction params error in async grpc handler
- [Add]: Previledge async test & resource group test (#2878)
- fix: refine error message (#2887)
- fix: loose grpcio upper version range (#2891)

See also: #2358, #2838, #2824
issue: https://github.com/milvus-io/pymilvus/issues/2886

### DIFF
--- a/examples/async_DDL_test.py
+++ b/examples/async_DDL_test.py
@@ -1,0 +1,256 @@
+import asyncio
+import time
+import numpy as np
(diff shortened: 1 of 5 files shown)


czs007 (approved):
/lgtm

PR #2914, opened 2025-07-30
fix: Return new pk value for upsert when autoid=true

yhmo:
issue: https://github.com/milvus-io/milvus/issues/25397


XuanYang-cn (approved):
/lgtm

PR #2923, opened 2025-08-06
feat: update sdk for ngram

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/42053

### DIFF
--- a/pymilvus/orm/schema.py
+++ b/pymilvus/orm/schema.py
@@ -412,6 +412,9 @@ class FieldSchema:
         self.element_type = kwargs.get("element_type")
         if "mmap_enabled" in kwargs:
             self._type_params["mmap_enabled"] = kwargs["mmap_enabled"]
+        if "enable_ngram_index" in kwargs:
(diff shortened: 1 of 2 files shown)


pymilvus/orm/schema.py
```diff
@@ -412,6 +412,9 @@ def __init__(self, name: str, dtype: DataType, description: str = "", **kwargs)
         self.element_type = kwargs.get("element_type")
         if "mmap_enabled" in kwargs:
             self._type_params["mmap_enabled"] = kwargs["mmap_enabled"]
+        if "enable_ngram_index" in kwargs:
+            self._type_params["min_gram"] = kwargs["min_gram"]
```
zhengbuqian, line 416: do we have default value for those if not provided? 

do we check for valid min/max_gram in server side?

XuanYang-cn, line 416: If user passes "enable_ngram_index" and not "min_gram", "max_gram", this line will raise KeyError

PR #2938, opened 2025-08-20
feat: Add partial update support for upsert operations

weiliu1031:
- Add partial_update parameter to MilvusClient and AsyncMilvusClient
- Enable selective field updates without requiring all fields
- Add example demonstrating partial vs full upsert usage
- Update protobuf definitions and handlers for partial update support

This allows efficient metadata updates without modifying vector data.


XuanYang-cn (approved):
/lgtm

PR #2955, opened 2025-08-27
fix: Refine partial upsert to preserve unspecified fields

weiliu1031:
issue: milvus-io/milvus#43980

This commit enhances the partial update functionality to prevent the unintentional clearing of existing data. The core change ensures that when `partial_update=True`, fields not provided by the user are omitted from the request payload, preserving their current values on the server.

The key modifications are in `pymilvus/client/prepare.py`:

- In `_parse_upsert_row_request`, the logic now explicitly skips processing for any fields that are missing from an entity when `partial_update` is enabled.

- This prevents the client from automatically assigning `None` to missing nullable fields or injecting default values, which would otherwise cause the server to overwrite existing data.

### DIFF
--- a/pymilvus/client/prepare.py
+++ b/pymilvus/client/prepare.py
@@ -553,10 +553,17 @@ class Prepare:
             field_len[DYNAMIC_FIELD_NAME] = 0
 
         try:
+            fields_num = 0
             for entity in entities:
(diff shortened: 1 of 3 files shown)


XuanYang-cn:
@weiliu1031 Please cherry-pick this pr to 2.6 branch THX

PR #2957, opened 2025-08-27
enhance: Add unixmsec in every RPC call

XuanYang-cn:
See also: milvus-io/milvus#44058

### DIFF
--- a/pymilvus/client/interceptor.py
+++ b/pymilvus/client/interceptor.py
@@ -13,16 +13,21 @@
 # limitations under the License.
 """Base class for interceptors that operate on all RPC types."""
 
+import time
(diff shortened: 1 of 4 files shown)


pymilvus/client/interceptor.py
```diff
@@ -13,16 +13,17 @@
...
+import time
 from typing import Any, Callable, List, NamedTuple, Optional, Tuple
 
 import grpc
 
 
 def _api_level_md(**kwargs) -> Optional[List[Tuple]]:
-    metadata = None
+    metadata = []
+    metadata.append(("client-request-unixmsec", str(time.time_ns())))
```
congqixia: IMHO, time.time_ns() returns nano seconds, this header value shall be in millisecond

PR #2974, opened 2025-09-08
fix: Correct validation logic for partial upsert requests

weiliu1031:
This commit fixes a validation bug in the `_parse_upsert_row_request` function that incorrectly enforced field count consistency for partial updates.


XuanYang-cn (approved):
/lgtm

PR #2999, opened 2025-09-19
enhance: update get_cost_from_status

chasingegg:
https://github.com/milvus-io/milvus/issues/44212

We will add more fields in extra_info and need to handle the case that extra_info is not nil but report_value is not in it.


XuanYang-cn (approved):
/lgtm

PR #3028, opened 2025-10-17
Fix hybrid_search to support EmbeddingList in request data

zhuwenxing:
- Add EmbeddingList import to grpc_handler.py
- Add automatic conversion of EmbeddingList to flat array in hybrid_search
- Set is_embedding_list flag when EmbeddingList is detected
- This enables hybrid search to work with struct array vector fields

This fix aligns hybrid_search behavior with the regular search method, which already supports EmbeddingList conversion.


XuanYang-cn (approved):
/lgtm

Issue #3033, opened 2025-10-20
[Enhancement]: Async Upsert: Support response with Primary Key.


Wulala-T:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

<img width="2136" height="998" alt="Image" src="https://github.com/user-attachments/assets/afabbdd7-6447-4689-b484-2bb637fe7ddb" />

iamjoker021:
@XuanYang-cn , 
I raised a PR for this issue
https://github.com/milvus-io/pymilvus/pull/3035

I am a first time contributor and this issue seems good first issue hence I tried and raised above PR.
Please feel free to review and if any issue highlight
Thanks!

PR #3034, opened 2025-10-20
Add async upsert response with primary key (Issue: #3033)

iamjoker021:
Hi, I am a first time contributer and saw this issue: https://github.com/milvus-io/pymilvus/issues/3033, which seems beginner friendly.
So I made the change and raised a PR.

Changes Made:
- I added "primary_keys" key and value in return statement of Upsert method in "async_milvus_client.py" module.
- Also, I added the "ids" key as default when no data is there referring to Insert method

### DIFF
--- a/pymilvus/client/entity_helper.py
+++ b/pymilvus/client/entity_helper.py
@@ -197,6 +197,27 @@ def convert_to_json(obj: object):
             return bool(obj)
         return obj
 
+    # Handle JSON string input
+    if isinstance(obj, str):
+        try:
+            # Validate JSON string by parsing it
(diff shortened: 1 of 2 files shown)


pymilvus/milvus_client/async_milvus_client.py
```diff
@@ -338,7 +338,7 @@ async def upsert(
             raise TypeError(msg)
 
         if len(data) == 0:
-            return {"upsert_count": 0}
+            return {"upsert_count": 0, "ids": []}
```
XuanYang-cn: Please align with MilvusClient.upsert, which contains only {"upsert_count": 0}

PR #3035, opened 2025-10-20
Add PK in the upsert response (Issue: #3033)

iamjoker021:
Hi, 
I am a first time contributer and saw issue: https://github.com/milvus-io/pymilvus/issues/3033, which seem easy.

Change I made in following file "async_milvus_client" are below
- Added the "ids" key which return the "primary_keys" from upsert_rows
- Added the "ids" key as empty list when no data present in upsert method

### DIFF
--- a/pymilvus/milvus_client/milvus_client.py
+++ b/pymilvus/milvus_client/milvus_client.py
@@ -272,7 +272,7 @@ class MilvusClient:
             raise TypeError(msg)
 
         if len(data) == 0:
-            return {"upsert_count": 0}
+            return {"upsert_count": 0, "ids": []}
 
         conn = self._get_connection()
         # Upsert into the collection.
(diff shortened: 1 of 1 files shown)


XuanYang-cn:
@iamjoker021 The changes seems all good. But I'd like you to help fix the upsert() return in milvus_client.py MilvusClient() too. It's wrongly returned as "primary_keys", and the empty results needs to add "ids" too.

iamjoker021:
Hi @XuanYang-cn , 
Thanks for the feedback. I have updated the key from "primary_keys" to "ids" as you mentioned and also kept "ids" as empty list on case of no data present.
Just a small doubt, what will happen if people are currently using the "primary_keys" as key-name in their code?
Thanks!

XuanYang-cn:
> Hi @XuanYang-cn , Thanks for the feedback. I have updated the key from "primary_keys" to "ids" as you mentioned and also kept "ids" as empty list on case of no data present. Just a small doubt, what will happen if people are currently using the "primary_keys" as key-name in their code? Thanks!

This is a really good question. Actually this is a tough decision, whether to break the API return or keep it for compatibility. 

After I check our doc and notice the doc is outdated for a long time, I decide to break the return. And treat the previous "primary_key" as a bug, not feature.

I'll put proper release notes to describe this.

Thanks again for this PR.

PR #3047, opened 2025-10-27
Add example for STRUCT import

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/42148

Add example for STRUCT import for JSON/Parquet

### DIFF
--- a/examples/bulk_import/example_bulkinsert_json.py
+++ b/examples/bulk_import/example_bulkinsert_json.py
@@ -9,8 +9,6 @@ from minio.error import S3Error
 
 from pymilvus import (
     DataType,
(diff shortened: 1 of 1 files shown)


examples/bulk_import/example_bulkinsert_json.py
```diff
@@ -0,0 +1,356 @@
...
+import time
+import os
+from typing import List
+
+from minio import Minio
+from minio.error import S3Error
+
+from pymilvus import (
+    DataType,
+    utility,
```
XuanYang-cn: Please avoid use utility, this is orm-based API.

PR #3054, opened 2025-10-30
Support TypeParams in STRUCT

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/42148

### DIFF
--- a/pymilvus/client/abstract.py
+++ b/pymilvus/client/abstract.py
@@ -1,5 +1,4 @@
 import abc
-import json
 import logging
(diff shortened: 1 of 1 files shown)


pymilvus/client/abstract.py
```diff
@@ -1,4 +1,5 @@
 import abc
+import json
```
XuanYang-cn: Please use orjson instead of json.

XuanYang-cn (changes requested):

XuanYang-cn (approved):
/lgtm

PR #3076, opened 2025-11-14
Fix: MilvusClient.insert() does not pass **kwargs to underlying insert_rows() call

### LINKED ISSUE #3075: [Bug]: MilvusClient.insert() does not pass **kwargs to underlying insert_rows() call

### Describe the bug

The `MilvusClient.insert()` method accepts `**kwargs` parameter but does not pass it to the underlying `conn.insert_rows()` call. This prevents users from passing important parameters like `retry_times`, `retry_on_rate_limit`, etc. to control retry behavior during insert operations.

This is inconsistent with:
1. `MilvusClient.upsert()` method, which correctly passes `**kwargs`
2. `Collection.insert()` method in ORM, which also passes `**kwargs`

### Expected Behavior

The `MilvusClient.insert()` method should pass `**kwargs` to `conn.insert_rows()` so that users can control retry behavior and other advanced options, just like `upsert()` method does.

```markdown

client = MilvusClient("http://localhost:19530")

# This should work but currently doesn't pass retry_times to the underlying call
client.insert(
    collection_name="test_collection",
    data=[{"id": 1, "vector": [0.1, 0.2]}],
    retry_times=10  # This parameter is silently ignored
)The `retry_times` parameter is accepted by the method signature but not forwarded to the actual RPC call, so it has no effect.
```

### Environment details

```markdown
- Hardware/Software conditions (OS, CPU, GPU, Memory): Any
- Method of installation (Docker, or from source): Any
- Milvus version: v2.4.0 (and likely earlier versions)
- Milvus configuration: N/A
```

### Anything else?

_No response_

thangTang:
Fix #3075

### DIFF
--- a/examples/bulk_import/bulk_writer_all_types.py
+++ b/examples/bulk_import/bulk_writer_all_types.py
@@ -216,6 +216,7 @@ def remote_writer(schema: CollectionSchema, file_type: BulkFileType):
     with RemoteBulkWriter(
             schema=schema,
             remote_path="bulk_data",
+            local_path="/tmp/PARQUET",
             connect_param=RemoteBulkWriter.S3ConnectParam(
                 endpoint=MINIO_ADDRESS,
                 access_key=MINIO_ACCESS_KEY,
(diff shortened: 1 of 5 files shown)


XuanYang-cn:
@thangTang Please sign off this PR to pass the DCO robot check https://github.com/milvus-io/pymilvus/pull/3076/checks?check_run_id=55377453235

XuanYang-cn (approved):
/lgtm

PR #3077, opened 2025-11-14
[Enhancement] allows to specufy temp local path for RemoteBulkWriter for temporarily persisting data files

yhmo:
issue: https://github.com/milvus-io/pymilvus/issues/3015

Issue #3015, opened 2025-10-02
bulk_writer tries to use /usr/local/lib/python3.11/site-packages/bulk_writer and runs into permission error


wolfganghuse:
Hello!
When using bulk_writer in a Jupyter Notebook it fails with permissions error.

> PermissionError: [Errno 13] Permission denied: '/usr/local/lib/python3.11/site-packages/bulk_writer'

`        local_path = Path(sys.argv[0]).resolve().parent.joinpath("bulk_writer")
        super().__init__(schema, str(local_path), chunk_size, file_type, config, **kwargs)`

Found here: https://github.com/milvus-io/pymilvus/blob/43f04470bc3f3488957cf0c58877fdb317da9c09/pymilvus/bulk_writer/remote_bulk_writer.py#L115C9-L116C1

Can we ensure that local writes will not happen in this location ?

Regards,
Wolfgang

yhmo:
RemoteBulkWriter requires a local path to temporarily persist data files and then upload to s3.
The local path is automatically detected from the path where to run the script. The Jupyter Notebook runs the script from /user/local/lib so that it throws this error.
Proposal: allows users to specify a local path for temporarily persisting data files.

yhmo:
This pr will fix the issue:
https://github.com/milvus-io/pymilvus/pull/3077

We offer a parameter for RemoteBulkWriter to specify the local temp path:
```
with RemoteBulkWriter(
            schema=schema,
            remote_path="bulk_data",
            local_path="/tmp/PARQUET",
            connect_param=RemoteBulkWriter.S3ConnectParam(
                endpoint=MINIO_ADDRESS,
                access_key=MINIO_ACCESS_KEY,
                secret_key=MINIO_SECRET_KEY,
                bucket_name="a-bucket",
            ),
            segment_size=512 * 1024 * 1024,
            file_type=file_type,
    ) as writer:
```

RemoteBulkWriter persists data files into the temp local path and then uploads files to S3, and eventually deletes the local path after the data files are successfully uploaded.

If the "local_path" is not specified, it will use the execution path as the default.

wolfganghuse:
Thanks @yhmo for fixing this. Really appreciate!

XuanYang-cn:
Should be ready in the next 2.6.x release

XuanYang-cn:
[...]

PR #3078, opened 2025-11-14
[Enhancement] allows to specufy temp local path for RemoteBulkWriter for temporarily persisting data files

yhmo:


XuanYang-cn:
See also: #3077, #3015

Issue #3060, opened 2025-11-05
[Bug]: async flush() does not wait for completed


yanliang567:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

run async def test_async_client_with_schema(), if calling async flush() it fails with assertion in num_entities for actually the flush() call is not completed.

![Image](https://github.com/user-attachments/assets/634e7589-da26-4d30-a33b-cbfb6232fbb0)

### Steps/Code To Reproduce behavior

```markdown

```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

XuanYang-cn:
/assign @yanliang567 
please help verify this fix
/unassign @XuanYang-cn @jac0626

yanliang567:
verified fix on pymilvus-2.6.4rc12 and pymilvus-2.7.0rc68

PR #3082, opened 2025-11-19
fix: Fix async flush() not waiting for segments to be flushed (#3060)

jac0626:
The async flush() method was returning immediately after the RPC call without waiting for segments to be actually flushed, causing num_entities to return incorrect values (usually 0) right after flush().

This fix:
- Adds async get_flush_state() method to check flush status
- Adds async _wait_for_flushed() method to wait for flush completion
- Updates flush() to wait for all segments to be flushed in parallel using asyncio.gather()
- Ensures async flush() behavior matches synchronous flush()

See also: #3060

### DIFF
--- a/pymilvus/client/async_grpc_handler.py
+++ b/pymilvus/client/async_grpc_handler.py
@@ -1885,12 +1885,76 @@ class AsyncGrpcHandler:
         check_status(resp)
 
     @retry_on_rpc_failure()
+    async def get_flush_state(
+        self,
+        segment_ids: List[int],
+        collection_name: str,
+        flush_ts: int,
+        timeout: Optional[float] = None,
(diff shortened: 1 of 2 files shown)


XuanYang-cn:
@jac0626 Plz dont use Fix or Resolve xx in commit message, This will tell github to close the issue when pr is merged, which we don't want to see.
image

XuanYang-cn (changes requested):
LGTM for the code, plz change the commit messages to aviod auto-close issue without verification

jac0626:
> LGTM for the code, plz change the commit messages to aviod auto-close issue without verification

done

XuanYang-cn (approved):
/lgtm

PR #3086, opened 2025-11-20
fix(async): include event loop ID in connection alias to prevent reusing closed connections

jac0626:
# Description

This PR fixes a critical issue where `AsyncMilvusClient` could reuse a connection associated with a closed event loop, causing `RuntimeError: Event loop is closed` when running multiple async tasks or tests sequentially.

## Problem

* The `create_connection` utility generates a connection alias based on URI, user, and other parameters but **ignores the running event loop**.
* In `AsyncMilvusClient`, connections are cached globally by this alias.
* If an event loop is closed (e.g., after `asyncio.run()` finishes) and a new loop is started, the client reuses the **cached connection** which is bound to the **old, closed** loop.
* This results in errors like: `RuntimeError: Event loop is closed` when trying to use the client in the new loop.

## Solution

* Updated `create_connection` in `pymilvus/milvus_client/_utils.py` to include the **current event loop ID** in the connection alias when `use_async=True`.
* **New Alias Format:** `async-{uri}-{hash}-loop{loop_id}`
* This ensures that each event loop gets its own dedicated connection instance.
* Added a check to raise `RuntimeError` if no running loop is found in an async context (fail-fast instead of silent failure).

## Verification

* Added a new unit test `tests/test_async_milvus_client_reuse.py` to reproduce the issue and verify the fix.
* The test confirms that different event loops generate different connection aliases, preventing invalid reuse.
* Existing tests pass.

## Related Issue

https://github.com/milvus-io/pymilvus/issues/3087

### DIFF
--- a/pymilvus/milvus_client/_utils.py
+++ b/pymilvus/milvus_client/_utils.py
@@ -1,3 +1,4 @@
+import asyncio
 import hashlib
 import logging
 
@@ -33,8 +34,15 @@ def create_connection(
             md5.update(token.encode())
             auth_fmt = f"{md5.hexdigest()}"
 
-        # different uri, auth, db_name cannot share the same connection
-        not_empty = [v for v in [use_async_fmt, uri, db_name, auth_fmt] if v]
+        # For async connections, include event loop ID in alias to prevent
+        # reusing connections from closed event loops
+        loop_id_fmt = ""
+        if use_async:
+            loop = asyncio.get_running_loop()
+            loop_id_fmt = f"loop{id(loop)}"
+
+        # different uri, auth, db_name, and event loop (for async) cannot share the same connection
+        not_empty = [v for v in [use_async_fmt, uri, db_name, auth_fmt, loop_id_fmt] if v]
         using = "-".join(not_empty)
 
     if connections.has_connection(using):
--- a/tests/test_async_milvus_client_reuse.py
+++ b/tests/test_async_milvus_client_reuse.py
@@ -1,0 +1,48 @@
+import asyncio
+import pytest
+from unittest.mock import patch, MagicMock
+from pymilvus import AsyncMilvusClient
+
+class TestAsyncConnectionReuse:
+    def test_async_client_alias_different_loops(self):
+        """
+        Test that AsyncMilvusClient generates different connection aliases
+        for different event loops to avoid reusing closed connections.
+        """
+        uri = "http://localhost:19530"
+        
+        async def create_client():
+            # Mock connections.connect to avoid real network connection
+            # Mock utility.get_server_type to avoid checking server version
+            with patch("pymilvus.orm.connections.Connections.connect") as mock_connect, \
+                 patch("pymilvus.orm.utility.get_server_type", return_value="milvus"):
+                
+                client = AsyncMilvusClient(uri=uri)
+                # Return the alias used and the current loop id
+                return client._using, id(asyncio.get_running_loop())
+
+        # Manually create loops to ensure we have distinct objects
+        loop1 = asyncio.new_event_loop()
+        loop2 = asyncio.new_event_loop()
+        
+        try:
+            # Run in loop 1
+            alias1, loop1_id = loop1.run_until_complete(create_client())
+            
+            # Run in loop 2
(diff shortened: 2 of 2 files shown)


pymilvus/milvus_client/_utils.py
```diff
@@ -38,7 +76,30 @@ def create_connection(
...
     if connections.has_connection(using):
-        return using
+        # For async connections, check if event loop matches
+        if use_async:
+            if not _check_async_connection_event_loop(using):
+                # Event loop mismatch or closed, remove old connection and create new one
+                logger.debug(f"Removing connection {using} due to event loop mismatch or closure")
+                try:
+                    connections._alias_handlers.pop(using, None)
+                    connections._alias_config.pop(using, None)
```
XuanYang-cn: I think we have a method in connections to remove an alias

pymilvus/milvus_client/_utils.py
```diff
@@ -38,7 +76,30 @@ def create_connection(
         using = "-".join(not_empty)
 
     if connections.has_connection(using):
-        return using
+        # For async connections, check if event loop matches
```
XuanYang-cn: Maybe move all these logic in has_connection? It seems we're trying to identify if a connections is valid by check deeper into the handler.

Maybe add a method for both handler for example "is_alive() -> bool" and write logics accordingly?

XuanYang-cn:
What's the issue, root cause, and what's this fix do? Please add them to the commit message. Can we add a ut to verify it's fixed?

pymilvus/milvus_client/_utils.py
```diff
@@ -38,7 +76,30 @@ def create_connection(
...
+                    connections._alias_config.pop(using, None)
+                except Exception as e:
+                    logger.debug(f"Error removing connection {using}: {e}")
+            else:
+                return using
+        else:
+            # For sync connections, check if channel is still valid
+            try:
+                handler = connections._fetch_handler(using)
+                if isinstance(handler, GrpcHandler) and handler._channel is None:
```
XuanYang-cn: This doesn't seems to be able to happen.

pymilvus/milvus_client/_utils.py
```diff
@@ -38,7 +76,30 @@ def create_connection(
...
+                    logger.debug(f"Removing sync connection {using} due to closed channel")
+                    connections._alias_handlers.pop(using, None)
+                    connections._alias_config.pop(using, None)
```
XuanYang-cn: ditto

pymilvus/milvus_client/_utils.py
```diff
@@ -38,7 +76,30 @@ def create_connection(
...
+                else:
+                    return using
+            except Exception:
```
XuanYang-cn: Why there will be exceptions? why are we ignorign them? what exceptions will be ignored?

pymilvus/milvus_client/_utils.py
```diff
@@ -1,11 +1,49 @@
...
+        except RuntimeError:
+            return False
+
+        # Check if channel's event loop matches current loop
+        channel = handler._async_channel
+        cython_ch = channel._cython_channel
+        if cython_ch is None:
+            return False
+
+        channel_loop = cython_ch._loop
```
XuanYang-cn: private vars may change  by grpc in anytime, it's not always safe. And why are we checking them?

pymilvus/client/async_grpc_handler.py
```diff
@@ -117,6 +123,15 @@ def __exit__(self: object, exc_type: object, exc_val: object, exc_tb: object):
...
         self._async_channel = None
+        # Automatically remove connection from pool when channel is closed
+        if self._connection_alias:
+            try:
+                connections._alias_handlers.pop(self._connection_alias, None)
+                connections._alias_config.pop(self._connection_alias, None)
+            except Exception as e:
+                logger.debug(
+                    f"Error removing connection {self._connection_alias} during cleanup: {e}"
+                )
```
XuanYang-cn: Why we need call back? I think AsyncMilvusClient close() already does the same thing. The call stack should be  AsyncClient.close -> connections.async_remove_connections -> asyncgrpchander.close() why reverse here?

Issue #3099, opened 2025-11-25
[Enhancement]: improve Async Connection Error Message


jac0626:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Problem Description

When users directly call `AsyncMilvusClient` in a synchronous function, the current error message is not user-friendly and does not clearly indicate that this might be a case of calling async code from a sync context.

## Current Behavior

When creating `AsyncMilvusClient` in a synchronous function, the following error is raised:

```
RuntimeError: no running event loop

```

## Issues

1. The error message does not clearly indicate that this is a common problem of **calling async code from sync context**
2. Missing specific solutions and code examples
3. Users may not know how to fix this issue

## Expected Behavior

The error message should:
1. **Clearly identify the problem**: Indicate that the user might be calling async client from a synchronous function
2. **Provide solutions**: Explain that it needs to be called inside an async function, or use `asyncio.run()`
3. **Provide code examples**: Show the correct usage pattern

## Suggested Error Message Format

```
ConnectionConfigException: Cannot create async connection: no running event loop. 
Async connections must be created within an async context (e.g., inside an async function).

Hint: This might be because you're calling an async client from a synchronous function.
Solution: Please call AsyncMilvusClient inside an async function (async def), 
or use asyncio.run() or an event loop to run async code.

Example:
  async def main():
      client = AsyncMilvusClient('http://localhost:19530')
      # Use client...
  
  # Then run:
  asyncio.run(main())
```

wzdavid:
Met the same "no running event loop" problem with pymilvus 2.6.4 when using llama-index-vector-stores-milvus. Had to downgrade pymilvus to 2.6.2.

jac0626:
> Met the same "no running event loop" problem with pymilvus 2.6.4 when using llama-index-vector-stores-milvus. Had to downgrade pymilvus to 2.6.2.

did you use async client in sync context? this is allowed before 2.6.4 , but now is banned.

PR #3102, opened 2025-11-25
enhance: improve AsyncMilvusClient connection error message

jac0626:
Provide user-friendly error message with solutions and examples when AsyncMilvusClient is called from sync context without event loop.

also see https://github.com/milvus-io/pymilvus/issues/3099

### DIFF
--- a/pymilvus/milvus_client/_utils.py
+++ b/pymilvus/milvus_client/_utils.py
@@ -2,6 +2,7 @@ import asyncio
 import hashlib
 import logging
 
+from pymilvus.exceptions import ConnectionConfigException
 from pymilvus.orm.connections import connections
 
 logger = logging.getLogger(__name__)
@@ -38,8 +39,12 @@ def create_connection(
         # reusing connections from closed event loops
         loop_id_fmt = ""
         if use_async:
-            loop = asyncio.get_running_loop()
-            loop_id_fmt = f"loop{id(loop)}"
+            try:
+                loop = asyncio.get_running_loop()
+                loop_id_fmt = f"loop{id(loop)}"
+            except RuntimeError as e:
+                error_msg = "Cannot create async connection: no running event loop. Please ensure you are running in an async context."
+                raise ConnectionConfigException(message=error_msg) from e
 
         # different uri, auth, db_name, and event loop (for async) cannot share the same connection
         not_empty = [v for v in [use_async_fmt, uri, db_name, auth_fmt, loop_id_fmt] if v]


pymilvus/milvus_client/_utils.py
```diff
@@ -38,8 +39,24 @@ def create_connection(
...
+                    "Solution: Please call AsyncMilvusClient inside an async function (async def), "
+                    "or use asyncio.run() or an event loop to run async code.\n\n"
+                    "Example:\n"
+                    "  async def main():\n"
+                    "      client = AsyncMilvusClient('http://localhost:19530')\n"
+                    "      # Use client...\n\n"
+                    "  # Then run:\n"
+                    "  asyncio.run(main())"
+                )
+                raise ConnectionConfigException(message=error_msg) from e
```
XuanYang-cn, line 47: I'm wondering this might be too strict for user to init AsyncMilvusClient class inside a event loop

``` python
>>> c = AsyncMilvusClient()
```

Fails when creating an object is not intuitive.

jac0626: > I'm wondering this might be too strict for user to init AsyncMilvusClient class inside a event loop
> 
> ```python
> >>> c = AsyncMilvusClient()
> ```
> 
> Fails when creating an object is not intuitive.

because now channel and event loop is binding during _init_, so we must require user to init in a running event loop, if we move the binding to when needed, we can  init AsyncMilvusClient  in sync code, but still require  an running event loop when needed(eg. first query).

XuanYang-cn: There's a workaround for python shell
```shell
python -m asyncio
```

```python
>>> from pymilvus import AsyncMilvusClient
>>> c = AsyncMilvusClient() # this wont raise exception
>>> await c.list_collections() # we can await directly
```

jac0626: > There's a workaround for python shell
> 
> ```shell
> python -m asyncio
> ```
> 
> ```python
> >>> from pymilvus import AsyncMilvusClient
> >>> c = AsyncMilvusClient() # this wont raise exception
> >>> await c.list_collections() # we can await directly
> ```

yes，this is ok because it runs in an async context

XuanYang-cn: https://github.com/grpc/grpc/blob/eaae0ff558694d5b0fff0ab439095d5920bde5b8/src/python/grpcio/grpc/_cython/_cygrpc/aio/common.pyx.pxi#L249

XuanYang-cn: Grpc's bahaivour on creating an async channel is very suprising.

jac0626: > https://github.com/grpc/grpc/blob/eaae0ff558694d5b0fff0ab439095d5920bde5b8/src/python/grpcio/grpc/_cython/_cygrpc/aio/common.pyx.pxi#L249

this function may allow user init asyncClient in sync context, but should we allow this usage?

Issue #3134, opened 2025-12-09
[Enhancement]: Refactor: Invasive Changes Required for Adding New Data Types


jac0626:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Problem

Adding a new data type requires modifying **5+ files** with scattered `elif` branches:

- `types.py` - `HybridExtraList._extract_lazy_fields()` (7+ elif branches)
- `entity_helper.py` - `entity_to_field_data()` (10+ elif branches)
- `search_result.py` - Add to type set
- `embedding_list.py` - Add to dtype_map
- `bulk_writer/constants.py` - Add to multiple dictionaries

This violates the Open/Closed Principle and makes the codebase hard to maintain.

## Current Code Pattern

```python
# Repeated in multiple files
elif field_data.type == DataType.FLOAT16_VECTOR:
    dim = field_data.vectors.dim
    bytes_per_vector = dim * 2
    start_pos = index * bytes_per_vector
    # ... 10+ lines of similar code
```

## Proposed Solution

Implement a **Type Handler Registry** using Strategy Pattern:

```python
class TypeHandler(ABC):
    @abstractmethod
    def extract_from_field_data(self, field_data, index, row_data):
        pass
    
    @abstractmethod
    def pack_to_field_data(self, entity_values, field_data, field_info):
        pass

# Usage: Replace all elif branches with
handler = get_type_handler(field_data.type)
handler.extract_from_field_data(field_data, index, row_data)
```

## Benefits

- **Before**: Modify 5+ files to add a type
- **After**: Create 1 handler class + 1 line registration
- Better testability, extensibility, and maintainability

## Migration

1. Create `TypeHandler` base class and registry
2. Migrate existing types incrementally
3. Replace `elif` branches with registry calls
4. Remove old code

XuanYang-cn:
https://github.com/milvus-io/pymilvus/pull/3507 This PR might need to take extra care

XuanYang-cn:
Closing — the refactor is done on master, backported to 3.0 and 2.6.
  
  Instead of the proposed TypeHandler registry, we went with a data-only TypeInfo fact registry
  plus consumer-owned dispatch (a universal handler would couple packing/decoding/validation and
  pull heavy imports like pyarrow into the search path). An early registry attempt (#3531) was
  reverted (#3545).

  Result: a DataType's static facts (slot, vector width, placeholder, dtype) live once in
[...]

PR #3135, opened 2025-12-09
refactor: refactor data process code

jac0626:
- Introduced TypeHandler strategy pattern to encapsulate type-specific logic in dedicated handlers
- Refactored `entity_helper.py` and `search_result.py`, eliminating extensive if/elif branches
- Added `type_handlers` module with scalar, vector, and complex handlers

also see #3134


jac0626:
this pr is too heavy to review, will be split

PR #3142, opened 2025-12-16
enhance: change highlight query keyword

aoiasd:
Use `highlight_query` instead of queries.
relate: https://github.com/milvus-io/milvus/issues/42589


XuanYang-cn (approved):
/lgtm

PR #3158, opened 2025-12-25
feat: Add namespace support

sunby:
issue: #3160 
pr: https://github.com/milvus-io/pymilvus/pull/3159
add enable_namespace in create_collection api and namespace in query/search api


pymilvus/grpc_gen/milvus-proto
XuanYang-cn, line 1: Please use the commit from 2.6 branch of milvus-proto

Issue #3163, opened 2025-12-26
[Bug]: Cannot create SCANN index with with_raw_data=False


smdsbz:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

When I create SCANN index with `with_raw_data=False` via pymilvus, the index parameter is wiped. However, I am able to properly create the index with Attu.

The steps to reproduce is shown as below:

<img width="2596" height="1426" alt="Image" src="https://github.com/user-attachments/assets/1665b8cf-989b-4747-a81f-4b044ebb65bc" />

<img width="2750" height="1796" alt="Image" src="https://github.com/user-attachments/assets/bf290a10-fdf4-4350-9f4e-926d1185d31a" />

Milvus cluster: reproduced on 2.5.12 and above
pymilvus: 2.6.5
Attu: 2.5.5

### Expected Behavior

Able to properly initialize the SCANN index params just as shown in the docs:

<img width="1738" height="1030" alt="Image" src="https://github.com/user-attachments/assets/5efc0a17-55bf-44a4-ace2-ab43e7ac47ce" />

### Steps/Code To Reproduce behavior

```markdown
See `Describe the bug`
```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0): reproduced on 2.5.12 and above
- Milvus configuration (Settings you made in `server_config.yaml`):

- pymilvus version: 2.6.5 via `pip install`
- Attu version: 2.5.5
```

XuanYang-cn:
/assign @jac0626 
PTAL

FeilongHou:
This issue is currently blocking e2e test cases. More specifically on tests/python_client/milvus_client/test_milvus_client_alter.py::TestMilvusClientAlterIndex::test_milvus_client_alter_index_default

![Image](https://github.com/user-attachments/assets/2e7936a2-4e88-4fb3-81e1-1ac9762a8c7f)
The current behavior is showing a lower case 'true' instead of an upper case 'True' \
This pymilvus bug exists in 2.7.0rc96 and 2.6.6rc12. \
Pymilvus version earlier than these are ok.

XuanYang-cn:
@FeilongHou Help verify plz

PR #3164, opened 2025-12-26
fix: preserve False boolean values in SCANN index parameters

jac0626:
use if is not none to check the parameter
see https://github.com/milvus-io/pymilvus/issues/3163


XuanYang-cn (approved):
/lgtm

Issue #3019, opened 2025-10-11
[Bug]: insert 10000 int8_vector(dim=768), run out of memory in client-side


yhmo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Insert 10000 int8_vector(dim=768), it takes long time(more than 10 seconds) and eventually run out of memory in client-side

### Steps/Code To Reproduce behavior

Insert 10000 int8_vector(dim=768) in one batch
```markdown
import random
import numpy as np

from pymilvus import (
    DataType,
    MilvusClient,
)


def gen_int8_vectors(num: int, dim: int):
    raw_vectors = []
    int8_vectors = []
    for _ in range(num):
        raw_vector = [random.randint(-128, 127) for _ in range(dim)]
        raw_vectors.append(raw_vector)
        int8_vector = np.array(raw_vector, dtype=np.int8)
        int8_vectors.append(int8_vector)
    return raw_vectors, int8_vectors


if __name__ == "__main__":
    c = MilvusClient()

    dim = 768
    nb = 10000
    collection_name = "hello_milvus_int8"

    schema = MilvusClient.create_schema()
    schema.add_field("int64", DataType.INT64, is_primary=True, auto_id=True)
    schema.add_field("int8_vector", DataType.INT8_VECTOR, dim=dim)

    if c.has_collection(collection_name):
        c.drop_collection(collection_name)
    c.create_collection(collection_name, schema=schema)

    _, vectors = gen_int8_vectors(nb, dim)
    rows = [{"int8_vector": vectors[i]} for i in range(nb)]
    c.insert(collection_name, rows)
```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

yhmo:
In entity_helper.py, the pack_field_value_to_field_data() method use "+=" to append bytes to `FieldData.vectors.int8_vector`
https://github.com/milvus-io/pymilvus/blob/c6459680ea29da06da04aa5b1bd66d5ad1e95431/pymilvus/client/entity_helper.py#L439

`FieldData.vectors.int8_vector` is bytes. Normally, a bytes object can be appended without memory problem. But the `FieldData.vectors.int8_vector` is wrapped by grpc and there could be some special process for "+=" operator.
[...]

PR #3181, opened 2026-01-08
fix:use list join instead of += to operate bytes to prevent oom

jac0626:
use list join instead of += to operate bytes to prevent oom
see #3019


pymilvus/client/entity_helper.py
```diff
@@ -28,6 +28,10 @@
 
 CHECK_STR_ARRAY = True
 
+# Cache for temporary byte lists for bytes vector fields (int8, binary, float16, bfloat16)
+# key: field_data object id, value: list of bytes
+_vector_bytes_cache: Dict[int, List[bytes]] = {}
```
XuanYang-cn: This is global cache, does it able to handle concurrent search?

jac0626: you are right，i will fix it，try to use a local variable

Issue #3186, opened 2026-01-09
[Bug]: Implement Global Thread-Safe Schema Cache with Eviction Policy


jac0626:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

**Description:**
Currently, `schema_cache` is an instance attribute of `GrpcHandler`. This has two main drawbacks:
1.  **Lifecycle Tying:** When `GrpcHandler` is reset (e.g. during `reset_db_name`), the cache is wiped, leading to unnecessary `DescribeCollection` calls.
2.  **Redundancy:** Multiple clients connecting to the same cluster but different DBs (or using different connection aliases) maintain duplicate caches.

**Proposed Changes:**
1.  **Global Cache:** Extract `schema_cache` into a global singleton or module-level component.
2.  **Key Strategy:** The cache key must identify the unique collection resource. Proposed key: `Hash(endpoint, db_name, collection_name)`.
3.  **Thread Safety:** The cache must be thread-safe (using `RWMutex` or `Lock`) to support concurrent access from multiple threads/coroutines.
4.  **Eviction Policy:** Implement an eviction policy (e.g., LRU) to prevent unbound memory growth, as the cache is now long-lived.
5.  **Invalidation:** Ensure `load_collection` or schema change operations update or invalidate the appropriate cache entries.

**Impact:**
- Reduced metadata RPC calls (`DescribeCollection`).
- Lower memory footprint by removing duplicate schema entries.
- Better stability for long-running applications.

### Steps/Code To Reproduce behavior

```markdown

```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

PR #3188, opened 2026-01-09
Refactor: Introduce thread-safe GlobalSchemaCache for gRPC handlers

jac0626:
- Global Singleton: Replaced instance-level caches with a thread-safe GlobalSchemaCache
- LRU Eviction: Implemented LRU logic with a 4096 capacity.
- Efficiency: Reduced redundant describe_collection calls across client instances.
see  #3186


jac0626:
> I think a popular library cachetools can help alot.



> cachetools

done

pyproject.toml
```diff
@@ -29,6 +29,8 @@ dependencies=[
     "python-dotenv>=1.0.1, <2.0.0",
     "pandas>=1.2.4",
     "numpy<1.25.0;python_version<='3.8'",
+    "cachetools>=5.0.0,<6.0.0;python_version<='3.8'",  # Python 3.8 compatibility
+    "cachetools>=5.0.0;python_version>='3.9'",  # Python 3.9+ can use latest
```
XuanYang-cn: The second is redundent, see numpy for reference.

I think the first might be unnecessary too, because packages often maintain supported python ranges, and pip should  be able to resolve it to the correct highest-version for python3.8.

jac0626: done

pymilvus/client/grpc_handler.py
```diff
@@ -701,16 +705,23 @@ def insert_rows(
...
         schema = self.describe_collection(
             collection_name, timeout=timeout, metadata=_api_level_md(**kwargs)
         )
         schema_timestamp = schema.get("update_timestamp", 0)
 
-        self.schema_cache[collection_name] = {
-            "schema": schema,
-            "schema_timestamp": schema_timestamp,
-        }
+        GlobalSchemaCache().set(
```
XuanYang-cn: No need to invalidate before update, just set the new one.

pymilvus/client/grpc_handler.py
```diff
@@ -745,27 +756,36 @@ def _prepare_row_insert_request(
...
+
+        # Cache the fetched schema and timestamp
+        cache.set(
+            cache_key,
+            {
                 "schema": schema,
                 "schema_timestamp": schema_timestamp,
-            }
+            },
+        )
```
XuanYang-cn: schema_timestamp is redundant, it's already in the schema, no need to get it and store in a different key.

pymilvus/client/schema_cache.py
```diff
@@ -0,0 +1,84 @@
...
+        Args:
+            endpoint: Server address (e.g., "localhost:19530")
+            db_name: Database name (empty string defaults to "default")
+            collection_name: Collection name
+
+        Returns:
+            Cache key in format "{endpoint}/{db_name}/{collection_name}"
+        """
+        db = db_name if db_name else "default"
+        return f"{endpoint}/{db}/{collection_name}"
```
XuanYang-cn: Better use a tuple for the key, much simpler than a customized format.

pymilvus/client/grpc_handler.py
XuanYang-cn: for later work: we should do the time-consuming work, and cache it when there's no cache, and invalidate it if it's out of date or there's no need, It's such a confusing way for someone directly calls update_cache, which is conflicting with get_from_cache_or_remote method.
Also we need to organize better how schema is updated.

pymilvus/client/schema_cache.py
```diff
@@ -0,0 +1,84 @@
...
+        return cls._instances[cls]
+
+    @classmethod
+    def _reset_for_testing(cls):
+        """Reset all singleton instances. Only for testing."""
+        with cls._lock:
+            cls._instances.clear()
+
+
+class GlobalSchemaCache(metaclass=Singleton):
```
XuanYang-cn: This class is more suitable for a universal  cache, it's not schema specific.

here's a more expandable proposal

```python
class CacheRegion:
# thread safe, universal caches for everything just like this one

class SchemaCache(CacheRegion):
# schema caches use specific keys, and stores one types of value.

class GlobalCache:
    schema = SchemaCache() # as a class variable, will be init once only.

# usage
GlobalCache.schema.get(endpoint, db_name, collection_name)
```

In GrpcHandler, we need three helper funcs only
```python
_get_schema(collection)
_invalidate_schema(collection) # when drop collection, alter a collection/collection field, insert/upsert exception etc
# and when drop a database.
_invalidate_db_schemas(db_name)
```

pymilvus/client/grpc_handler.py
XuanYang-cn: See the comments on global schema cache

pymilvus/client/schema_cache.py
```diff
@@ -0,0 +1,122 @@
...
+                # Double-check locking pattern
+                if cls not in cls._instances:
+                    cls._instances[cls] = super().__call__(*args, **kwargs)
+        return cls._instances[cls]
+
+    @classmethod
+    def _reset_for_testing(cls):
+        """Reset all singleton instances. Only for testing."""
+        with cls._lock:
+            cls._instances.clear()
```
XuanYang-cn: Singleton is nolonger needed

pymilvus/decorators.py
```diff
@@ -8,13 +8,86 @@
...
+class SchemaMismatchRetryable(Exception):
+    """Exception to signal schema mismatch that should trigger retry.
+
+    This exception is raised when the server returns a SchemaMismatch error,
+    indicating that the cached schema is stale and needs to be refreshed.
+    """
+
+    def __init__(self, collection_name: str, message: str = ""):
+        self.collection_name = collection_name
+        super().__init__(message or f"Schema mismatch for collection '{collection_name}'")
```
XuanYang-cn: Put this in exceptions.py and inherit from MilvusException plz

pymilvus/client/async_grpc_handler.py
```diff
@@ -596,27 +617,13 @@ async def insert_rows(
...
-        if resp.status.error_code == common_pb2.SchemaMismatch:
-            schema = await self.update_schema(collection_name, timeout, **kwargs)
-            request = await self._prepare_row_insert_request(
-                collection_name, entities, partition_name, schema, timeout, **kwargs
-            )
-            resp = await self._async_stub.Insert(
-                request=request, timeout=timeout, metadata=_api_level_md(**kwargs)
-            )
-        check_status(resp.status)
+        self._check_schema_mismatch(resp.status, collection_name)
```
XuanYang-cn: I dont think omit check_status is a good idea, people may copy this code forget to check_status in there codes.

(even though check_status on every method is a bad design, lets make it a must for every method.)

pymilvus/client/grpc_handler.py
```diff
@@ -671,47 +678,18 @@ def insert_rows(
...
-                schema=schema,
-                timeout=timeout,
-                **kwargs,
-            )
-        check_status(resp.status)
+        self._check_schema_mismatch(resp.status, collection_name)
         ts_utils.update_collection_ts(collection_name, resp.timestamp)
         return MutationResult(resp)
 
     def update_schema(self, collection_name: str, timeout: Optional[float] = None, **kwargs):
```
XuanYang-cn: I think after the refactory, no one is using update_schema now?

pymilvus/decorators.py
```diff
@@ -8,13 +8,86 @@
...
 
-from .exceptions import ErrorCode, MilvusException
+from .exceptions import DataNotMatchException, ErrorCode, MilvusException
 from .grpc_gen import common_pb2
 
 LOGGER = logging.getLogger(__name__)
 WARNING_COLOR = "\033[93m{}\033[0m"
 
 
+class SchemaMismatchRetryable(Exception):
```
XuanYang-cn: Also, we don't need to reorginize Milvus's exception message, just present as is. so collection_name is also not needed here, help to simplify the check_schema_mismatch into check_status.

pymilvus/client/grpc_handler.py
```diff
@@ -745,29 +716,46 @@ def _prepare_row_insert_request(
...
 
-        return schema, schema_timestamp
+    def _check_schema_mismatch(self, status: common_pb2.Status) -> None:
+        """Check response status and raise SchemaMismatchRetryableException if schema mismatch.
+
+        This method should be called after gRPC calls that may return SchemaMismatch.
+        It allows the @retry_on_schema_mismatch decorator to catch and retry.
+        """
+        if status.error_code == common_pb2.SchemaMismatch:
+            raise SchemaMismatchRetryableException(status.reason)
```
XuanYang-cn: better to let _check_schema_mismatch outside as an util method beside check_status for both grpchandler and async handler, avoid dup methods

pymilvus/client/cache.py
```diff
@@ -0,0 +1,101 @@
+import logging
```
XuanYang-cn, line 1: better to change module name `schema_cache` into more general name like cache or global_cache

pymilvus/client/async_grpc_handler.py
```diff
@@ -532,46 +533,56 @@ async def rename_collection(
...
+        GlobalCache.schema.invalidate_db(self.server_address, db_name)
+
+    def _check_schema_mismatch(self, status: common_pb2.Status) -> None:
+        """Check response status and raise SchemaMismatchRetryableException if schema mismatch.
+
+        This method should be called after gRPC calls that may return SchemaMismatch.
+        It allows the @retry_on_schema_mismatch decorator to catch and retry.
+        """
+        if status.error_code == common_pb2.SchemaMismatch:
+            raise SchemaMismatchRetryableException(status.reason)
```
XuanYang-cn: ditto

pymilvus/client/grpc_handler.py
XuanYang-cn: Or maybe just merge this logic into check_status

XuanYang-cn (changes requested):

PR #3192, opened 2026-01-13
Fix a bug of bulkwriter to support all-empty struct list

yhmo:
issue: https://github.com/milvus-io/pymilvus/issues/3199


XuanYang-cn (approved):
/lgtm

Issue #3189, opened 2026-01-09
[Bug]: GTsDict (collection timestamp cache) lacks multi-cluster/multi-database isolation


jac0626:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Problem

The `GTsDict` singleton in `pymilvus/client/ts_utils.py` caches the last write timestamp for each collection using only `collection_name` as the key.

This causes incorrect behavior when:
1. **Multiple clusters**: A single process connects to Cluster A and Cluster B, both having a collection named `"users"`. Writes to A pollute the timestamp cache used by B.
2. **Multiple databases**: Within one cluster, `db1.users` and `db2.users` share the same timestamp entry.

## Current Implementation

```python
class GTsDict(metaclass=Singleton):
    def update(self, collection: int, ts: int):
        self._last_write_ts_dict[collection] = ts  # key = collection_name only
```

## Impact

- **Session Consistency** relies on this timestamp. If the wrong timestamp is used, queries may return stale data or fail unexpectedly.
- This becomes a real issue in multi-tenant applications or microservices sharing a single process.

## Proposed Solution

Refactor `GTsDict` to use composite keys similar to `GlobalSchemaCache`:
```python
key = f"{endpoint}/{db_name}/{collection_name}"
```

## Related

- PR introducing `GlobalSchemaCache` with proper key isolation. The same pattern should be applied here.

### Steps/Code To Reproduce behavior

```markdown

```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

PR #3203, opened 2026-01-14
refactor:refactor collection ts cache

jac0626:
This PR refactors the collection timestamp cache to address isolation issues and potential memory leaks. see https://github.com/milvus-io/pymilvus/issues/3189
### Modifications
- **New Cache Implementation**: Replaced the unbound `GTsDict` with `CollectionTsCache`, which inherits from `CacheRegion` (LRU support).
- **Isolation Keys**: Changed the cache key from just `collection_name` to `(endpoint, db_name, collection_name)` to support multi-cluster/multi-db scenarios.
- **Context Propagation**: Updated `ts_utils.py`, `GrpcHandler`, and `AsyncGrpcHandler` to correctly propagate `endpoint` and `db_name` context to the cache.
- **Testing**: Added unit tests for cache isolation and verified strictly against existing tests.

### DIFF
--- a/pymilvus/client/prepare.py
+++ b/pymilvus/client/prepare.py
@@ -22,7 +22,7 @@ from pymilvus.orm.schema import (
 from pymilvus.orm.types import infer_dtype_by_scalar_data
 from pymilvus.settings import Config
 
-from . import __version__, blob, check, entity_helper, ts_utils, utils
+from . import __version__, blob, check, entity_helper, utils
 from .abstract import BaseRanker
 from .check import check_pass_param, is_legal_collection_properties
 from .constants import (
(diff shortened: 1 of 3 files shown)


pymilvus/client/prepare.py
```diff
@@ -1445,7 +1445,9 @@ def search_requests_with_expr(
         highlighter: Optional[Highlighter] = None,
         **kwargs,
     ) -> milvus_types.SearchRequest:
-        use_default_consistency = ts_utils.construct_guarantee_ts(collection_name, kwargs)
+        use_default_consistency = ts_utils.construct_guarantee_ts(
```
XuanYang-cn: lets move all use_default_consistency outside of prepare.py plz

XuanYang-cn (changes requested):
passin use_default_consisitency is way simpler and consists than pass in two underscore params

pymilvus/client/prepare.py
jac0626: done

Issue #3213, opened 2026-01-19
[FEATURE]: Support Columnar Search Results for High Performance


jac0626:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Description

The current `SearchResult` implementation in PyMilvus eagerly iterates over all results and converts them into Python objects (row-based). For large `top_k` queries (e.g., top_k=10,000) or high QPS scenarios, this deserialization becomes a significant client-side bottleneck.

This issue proposes introducing a **Columnar Search Result** mechanism that:
1.  Wraps the raw Protobuf `SearchResultData` structure directly.
2.  Provides **Lazy Access**: Only materializes fields when they are actually accessed by the user.
3.  Uses efficient columnar data access (e.g., accessing the entire `ids` list at once vs. iterating rows).
4.  Maintains compatibility with the existing `dict`-like interface of `SearchResult` hits.

## Motivation

Profiling shows that `SearchResult.__init__` consumes a large portion of CPU time during searches with large result sets. By avoiding eager object creation for every hit, we can achieve:
-   **Faster Initialization**: Zero-copy (or near zero-copy) initialization.
-   **Lower Memory Usage**: No intermediate Python objects for unused fields.

### New Class: `ColumnarSearchResult`
-   initialized with `schema_pb2.SearchResultData`.
-   Implements `__getitem__` and `__iter__` to return lightweight `RowProxy` objects.
-   Caching mechanism for accessing dynamic fields (parsing JSON only once per field/batch if needed).

### API Changes
-   Add `use_columnar=True` (or similar config param) parameter to `search()` and `hybrid_search()` methods in `MilvusClient` and `GrpcHandler`.
-   When enabled, returns `ColumnarSearchResult` instead of `SearchResult`.

## Tasks

- [x] Implement core `ColumnarSearchResult` class.
- [x] Implement lightweight `RowProxy` and `ColumnarHits` classes.
- [x] Support all PyMilvus data types:
    - [x] Scalars (int64, float, bool, varchar)
    - [x] Vectors (FloatVector, BinaryVector, Float16, BFloat16, Int8, Sparse)
    - [x] Complex types (JSON, Array, ArrayOfStruct, ArrayOfVector)
    - [x] Special types (Geometry, Timestamptz)
- [x] Implement Fallback logic for complex types not natively supported by the columnar structure.
- [x] Achieve 100% Unit Test Coverage.
- [x] Verify compatibility with existing `SearchResult` behavior.
[...]

PR #3208, opened 2026-01-15
Support element-filter

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/42148

PR #3214, opened 2026-01-19
feat:support columnar search result to better performance

jac0626:
see #3213

### DIFF
--- a/docs/design/columnar_search_result_en.md
+++ b/docs/design/columnar_search_result_en.md
@@ -1,0 +1,115 @@
+# Columnar Search Result
+
+> silas.jiang · 2026-01-29 · Implemented
+
+## Background
+
+While profiling large TopK searches, we noticed that `SearchResult.__init__` was taking a significant amount of time. The root cause is that it pre-creates all Hit objects during initialization—for a `nq=100, topk=10000` scenario, that means creating 1 million Python objects upfront.
+
+This design was fine when TopK was small, but as user scenarios became more complex, it turned into a client-side performance bottleneck.
+
+## Core Idea
+
+Switch to columnar storage with lazy object creation.
+
+Specifically:
+1. During initialization, just store references to the Protobuf—don't iterate through the data
+2. When the user accesses `result[0][5]`, create a lightweight RowProxy on demand
+3. When accessing `hit['field']`, that's when we actually extract the value from Protobuf
+
+This reduces initialization complexity from O(nq × topk) to O(nq).
+
+## Overall Structure
+
+```
+ColumnarSearchResult (extends list)
+    │
+    └── ColumnarHits (one per query)
+            │
+            ├── ids, distances (lazy slicing)
+            │
+            └── get_value(field, idx)
+                    │
+                    └── AccessorFactory → type-specific Accessors
+```
+
+**ColumnarSearchResult** is the top-level container that holds a reference to the raw Protobuf data. During initialization, it does one thing: create nq ColumnarHits based on the `topks` array, each recording its own `[start, end)` range.
+
+**ColumnarHits** stores results for a single query. `ids` and `distances` are lazy properties—slicing happens only on first access. Field access goes through `get_value()`, which internally uses AccessorFactory to create the appropriate type-specific accessor.
+
+**RowProxy** is an extremely lightweight object (just 3 fields) that implements the `Mapping` interface. All data access is delegated to ColumnarHits—it doesn't store any data itself.
+
+**AccessorFactory** uses the strategy pattern to select the right accessor based on DataType. For example, `FLOAT_VECTOR` needs to slice by dimension, `JSON` requires orjson parsing, while `VARCHAR` can just index directly.
+
+## Why This Design
+
+**Why use RowProxy instead of just returning a dict?**
+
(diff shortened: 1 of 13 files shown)


jac0626:
More work is needed in the future to improve the code's readability, maintainability, extensibility, and performance @jac0626 
- [ ] Extract base class - Dedupe shared logic with SearchResult(highlight, metadata parsing)
- [ ] Replace if-elif chains - Use TypeHandler registry pattern in _bind_accessor()
- [ ] Consolidate Accessor classes - Reduce boilerplate with generic/factory pattern
- [ ] Unify dynamic field handling - Merge $meta path with static field path
- [ ] Define Protocol/ABC - Formal interface for type checking compatibility

XuanYang-cn:
#3208 made some changes in search_result, might need extra attention.

But lets not hurry into this new feature, I believe we can do some small changes that we missed on the perf test, mainly engineering improvements. Let's not skip the **engineering improvements**, that's more like to be released quickly and harmlessly.

And for a new feature like this one, we need a design not a decision. In the design, I'd like some of these questions answered:
1. Why do we choose to add a flag in search, and make it complete compatible to the old search_result? Why not replace?
2. Why we need a complete compatible result?
3. can we provide to_pandas, to_arrow, etc.? can we make the new return results typed for better usages?
4. Why we choose to ignore those complex types? which are most likely to be beneficial from this feature.

Anyway, let's discuss designs based on the issue #3213, and then implement based on the final design.

For compatibility test(if we're choosing this way): the new code should pass OLD ut, new ut doesn't prove anything.

jac0626:
> #3208 made some changes in search_result, might need extra attention.
> 
> But lets not hurry into this new feature, I believe we can do some small changes that we missed on the perf test, mainly engineering improvements. Let's not skip the **engineering improvements**, that's more like to be released quickly and harmlessly.
> 
> And for a new feature like this one, we need a design not a decision. In the design, I'd like some of these questions answered:
> 
> 1. Why do we choose to add a flag in search, and make it complete compatible to the old search_result? Why not replace?
> 2. Why we need a complete compatible result?
> 3. can we provide to_pandas, to_arrow, etc.? can we make the new return results typed for better usages?
> 4. Why we choose to ignore those complex types? which are most likely to be beneficial from this feature.
> 
> Anyway, let's discuss designs based on the issue #3213, and then implement based on the final design.
> 
> For compatibility test(if we're choosing this way): the new code should pass OLD ut, new ut doesn't prove anything.

I will try to do engineering improvements firstly, then I would upload a design doc soon.

jac0626:
> #3208 made some changes in search_result, might need extra attention.
> 
> But lets not hurry into this new feature, I believe we can do some small changes that we missed on the perf test, mainly engineering improvements. Let's not skip the **engineering improvements**, that's more like to be released quickly and harmlessly.
> 
> And for a new feature like this one, we need a design not a decision. In the design, I'd like some of these questions answered:
> 
> 1. Why do we choose to add a flag in search, and make it complete compatible to the old search_result? Why not replace?
> 2. Why we need a complete compatible result?
> 3. can we provide to_pandas, to_arrow, etc.? can we make the new return results typed for better usages?
> 4. Why we choose to ignore those complex types? which are most likely to be beneficial from this feature.
> 
> Anyway, let's discuss designs based on the issue #3213, and then implement based on the final design.
> 
> For compatibility test(if we're choosing this way): the new code should pass OLD ut, new ut doesn't prove anything.

@XuanYang-cn Thanks for the feedback!

**Regarding #3208**: Already addressed — the columnar implementation has been updated to accommodate those changes.

**On engineering improvements**: Done.see https://github.com/milvus-io/pymilvus/pull/3240, I have investigating some other improvements, but just make little sense. 

**On design**: I've prepared a design doc:
- [Design Doc ZH](https://github.com/jac0626/pymilvus/blob/9808767cdac967b343028ccab45cc133100ea729/docs/design/columnar_search_result_zh.md)
- [Design Doc EN](https://github.com/jac0626/pymilvus/blob/feature/columnar-search-result/docs/design/columnar_search_result_en.md)

To answer your specific questions:

1. **Why compatible, not replace?**
   last version we provide a flag, now we are doing a direct replacement — no flag, no dual versions. "Compatible" here means the new `ColumnarSearchResult` maintains the same API contract as the old `SearchResult`, so existing user code continues to work without changes.

2. **Return types**:
   - For the standard iteration API (`hit.id`, `hit['field']`), return types are identical to the original.
   - For the new `get_column()` API
     - `return_type="list"`: Works for all types.
     - `return_type="numpy"`: Returns native `np.ndarray` for numeric/vector types. **For complex types (JSON, Dynamic, Sparse etc.) where numpy offers no benefit, we will raise an error instead of returning inefficient object arrays.**

3. **to_pandas / to_arrow**:
   Not implemented yet. We can add these as follow-up work

4. **Complex types**:
   Fully supported — JSON, ARRAY, dynamic fields ($meta), and all vector types are covered.

**On testing**:
Currently using patch-based compatibility tests (`test_columnar_compat.py`) to verify the new code passes the old SearchResult tests. Full validation would benefit from e2e testing against a real Milvus instance.

Let me know if you'd like me to update anything in the design doc!

XuanYang-cn:
Closing for stale

Issue #3236, opened 2026-01-27
[Bug]: v2.6.7 db name do not work


Sun-ZhenXing:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

It appears that the database specified by the URI was not loaded, and the `default` database was used instead.

### Expected Behavior

Use db specified by URI.

### Steps/Code To Reproduce behavior

```markdown
Waiting to be added...
```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

yhmo:
Could you show me the snippet to reproduce this bug?
I can't reproduce it with the following script.
```
if __name__ == '__main__':
    # I already have a database named "AAA"
    # connect, database is specified by the URI
    client = MilvusClient(uri="http://localhost:19530/AAA")
    COLLECTION_NAME = "my_collection"
    dim = 128
    client.drop_collection(collection_name=COLLECTION_NAME)
    schema = MilvusClient.create_schema()
    schema.add_field(field_name="id", datatype=DataType.INT64, is_primary=True, auto_id=False)
    schema.add_field(field_name="vector", datatype=DataType.FLOAT_VECTOR, dim=dim)
    schema.add_field(field_name="text", datatype=DataType.VARCHAR, max_length=100)

    index_params = client.prepare_index_params()
    index_params.add_index(
        field_name="vector",
        index_name="vector_index",
        index_type="FLAT",
        metric_type="L2",
    )
    client.create_collection(collection_name=COLLECTION_NAME,
                             schema=schema,
                             index_params=index_params,
                             )
    print("collection is created in AAA")

    C1 = MilvusClient(uri="http://localhost:19530", db_name="default")
    print("collections in default db:", C1.list_collections())

    C2 = MilvusClient(uri="http://localhost:19530", db_name="AAA")
    print("collections in AAA db:", C2.list_collections())

```

The output:
```
collection is created in AAA
collections in default db: []
collections in AAA db: ['my_collection']
```

Sun-ZhenXing:
@yhmo I can reproduce this problem accurately.


```python
from pymilvus import AsyncMilvusClient


async def main():
    client = AsyncMilvusClient(uri="http://4090.alexsun.top:19530/kb1")
    collections = await client.list_collections()
[...]

PR #3237, opened 2026-01-27
fix:parse dbname from uri

jac0626:
see #3236

### DIFF
--- a/tests/test_async_milvus_client.py
+++ b/tests/test_async_milvus_client.py
@@ -303,8 +303,35 @@ class TestAsyncMilvusClientNewFeatures:
         result = AsyncMilvusClient.create_struct_field_schema()
         assert isinstance(result, StructFieldSchema)
 
+    @pytest.mark.parametrize(
+        "uri, db_name, expected_db_name",
+        [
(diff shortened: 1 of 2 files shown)


tests/test_async_milvus_client.py
```diff
@@ -302,3 +302,43 @@ def test_create_struct_field_schema(self):
...
+            "pymilvus.orm.connections.Connections._fetch_handler", return_value=mock_handler
+        ), patch(
+            "pymilvus.orm.connections.Connections.has_connection", return_value=False
+        ), patch(
+            "pymilvus.orm.connections.Connections.connect"
+        ), patch.object(
+            AsyncMilvusClient, "get_server_type", return_value="milvus"
+        ):
+            # Test: URI with db_name in path, no explicit db_name parameter
+            client = AsyncMilvusClient(uri="http://localhost:19530/mydb")
```
XuanYang-cn: what about "http://xxxx/mydb/" see https://github.com/milvus-io/pymilvus/blob/b5ed60c87671e798712a319d51cf0c98c175b735/tests/test_connections.py#L463

Let's follow this test to enhance the test here, use tabled params

PR #3282, opened 2026-02-13
ci: require 90% patch coverage and auto-remove ci-passed on re-run

XuanYang-cn:
Enable codecov/patch status check with a 90% target on changed lines.
Add it as a condition for the ci-passed label in Mergify. Also add a
complementary Mergify rule that removes ci-passed when any required
check is not passing, so the label is cleared when authors push new
commits and CI re-triggers.


XuanYang-cn:
This PR doesn't need to cp to 2.6 branch

XuanYang-cn:
Yes it needed cp to 2.6 and its #3297

PR #3285, opened 2026-02-17
fix: retry on REPLICATE_VIOLATION for global cluster region switch

huanghaoyuanhhy:
## Summary
- When a Global Cluster switches its primary region, write operations to the old primary (now secondary) fail with `STREAMING_CODE_REPLICATE_VIOLATION`
- Previously this `MilvusException` was not handled in the retry decorator, so writes failed for up to 5 minutes until the background `TopologyRefresher` (300s interval) detected the change
- Add `_handle_global_routing_error()` in `GrpcHandler` to detect `REPLICATE_VIOLATION` and trigger immediate topology refresh
- Hook it into the `retry_on_rpc_failure` decorator's `MilvusException` branch (both sync and async) so the operation retries automatically after refresh

## Test plan
- [x] Deploy a Global Cluster on Zilliz Cloud (UAT)
- [x] Run continuous insert + search loop
- [x] Switch primary region in console
- [x] **Before fix**: INSERT fails with `REPLICATE_VIOLATION` for ~5 minutes until background refresh
- [x] **After fix**: INSERT auto-recovers in ~10 seconds (topology refresh + retry backoff)
- [x] Unit tests for `_handle_global_routing_error` (4 new tests, all passing)

### DIFF
--- a/pymilvus/decorators.py
+++ b/pymilvus/decorators.py
@@ -235,13 +235,13 @@ def retry_on_rpc_failure(
                             raise MilvusException(
                                 code=e.code, message=f"{to_msg}, message={e.message}"
                             ) from e
-                        if _retry_on_rate_limit and (
-                            e.code == ErrorCode.RATE_LIMIT
-                            or e.compatible_code == common_pb2.RateLimit
-                        ):
-                            await asyncio.sleep(back_off)
-                            back_off = min(back_off * back_off_multiplier, max_back_off)
-                        elif (
+                        if (
+                            _retry_on_rate_limit
+                            and (
+                                e.code == ErrorCode.RATE_LIMIT
+                                or e.compatible_code == common_pb2.RateLimit
+                            )
+                        ) or (
                             args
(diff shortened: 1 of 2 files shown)


pymilvus/client/grpc_handler.py
```diff
@@ -222,6 +222,20 @@ def _handle_global_connection_error(self, error: grpc.RpcError) -> None:
...
 
+    def _handle_global_routing_error(self, error: MilvusException) -> bool:
+        """Handle routing errors for global connections (e.g. primary switched).
+
+        Returns True if a topology refresh was triggered and the caller should retry.
+        """
+        if self._global_stub is None:
+            return False
+
+        if "REPLICATE_VIOLATION" in str(error.message):
```
bigsheeper: it would be better if grep string is more specific

huanghaoyuanhhy:
Fixed the review comment and lint issues. Verified locally with `make lint` - all checks passed.

PR #3141, opened 2025-12-15
feat: Support nullable vectors in Python SDK

marcelo-cjl:
related: #3154 

- Add None value handling in pack_field_value_to_field_data() for all vector types
- Add get_physical_index() helper for logical-to-physical index mapping
- Update extract_row_data_from_fields_data() with valid_data check and physical index mapping
- Update HybridHits.materialize() and HybridExtraList._extract_lazy_fields() for nullable vectors
- Update len_of() to use valid_data length as logical row count
- Remove nullable restriction for vector fields in insert method

## Summary by CodeRabbit

* **New Features**
  * Support for nullable vectors across all vector types; vector fields may be added to existing collections when nullable=True.
  * Automatic inference of missing vector dimensions from metadata.

* **Bug Fixes / Improvements**
  * More robust handling of nullable/sparse data, safer slicing and explicit type/validation checks.
  * Added caching to speed index mapping for nullable vectors.

* **Tests**
  * New unit tests covering nullable vector packing, extraction, and client-side validation.

* **Examples**
  * Added an example script demonstrating nullable-vector workflows and dynamic column addition.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #3289, opened 2026-02-24
fix: avoid calling len() on scipy sparse matrix in entity_to_field_data

### LINKED ISSUE #3288: Bug: len() on scipy sparse matrix raises TypeError in entity_to_field_data since v2.6.6
### Current Behavior

Inserting sparse float vectors using `scipy.sparse.csr_matrix` raises `DataNotMatchException`:

```
pymilvus.exceptions.DataNotMatchException: <DataNotMatchException: (message=The Input data type is inconsistent with defined schema, 
{sparse_float_vector} field should be a SPARSE_FLOAT_VECTOR, but got a {<class 'scipy.sparse._csr.csr_matrix'>} instead. 
Detail: sparse array length is ambiguous; use getnnz() or shape[0])>
```

### Expected Behavior

`scipy.sparse.csr_matrix` should be accepted as valid input for SPARSE_FLOAT_VECTOR fields during insert, as it was in v2.6.5 and earlier.

### Steps to Reproduce

```python
from scipy.sparse import random as sparse_random

# Create collection with sparse float vector field
# Insert using scipy sparse matrix
data = sparse_random(1000, 400000, density=0.01, format='csr')
client.insert(collection_name, [{"id": i, "sparse_vec": data[i]} for i in range(1000)])
```

### Root Cause

PR #3141 (commit 67b50168, "Support nullable vectors in Python SDK") introduced `if len(entity_values) > 0:` guard for all vector types in `entity_to_field_data()` at [entity_helper.py#L733](https://github.com/milvus-io/pymilvus/blob/master/pymilvus/client/entity_helper.py#L733):

```python
# Before (v2.6.5, working):
elif entity_type == DataType.SPARSE_FLOAT_VECTOR:
    field_data.vectors.sparse_float_vector.CopyFrom(sparse_rows_to_proto(entity_values))

# After (v2.6.6+, broken):
elif entity_type == DataType.SPARSE_FLOAT_VECTOR:
    if len(entity_values) > 0:  # <-- TypeError here
        field_data.vectors.sparse_float_vector.CopyFrom(sparse_rows_to_proto(entity_values))
```

Calling `len()` on a `scipy.sparse.csr_matrix` raises `TypeError: sparse array length is ambiguous; use getnnz() or shape[0]`. This `TypeError` propagates up and gets caught by the `except (TypeError, ValueError)` block, which wraps it as `DataNotMatchException`.

For other vector types (FLOAT_VECTOR, BINARY_VECTOR, etc.), `entity_values` is always a Python list so `len()` works fine. But for SPARSE_FLOAT_VECTOR, `entity_values` can be a `scipy.sparse.csr_matrix`.

### Suggested Fix

Remove the `len()` guard for the sparse vector branch since `sparse_rows_to_proto()` already handles the data correctly:

```python
elif entity_type == DataType.SPARSE_FLOAT_VECTOR:
    field_data.vectors.sparse_float_vector.CopyFrom(sparse_rows_to_proto(entity_values))
```

### Environment

- **pymilvus version**: 2.6.9rc5 (regression introduced in v2.6.6)
- **Python version**: 3.12.9
- **Last working version**: v2.6.5

### Stack Trace

```
File "pymilvus/client/grpc_handler.py", line 881, in batch_insert
    request = self._prepare_batch_insert_request(...)
File "pymilvus/client/grpc_handler.py", line 864, in _prepare_batch_insert_request
    Prepare.batch_insert_param(collection_name, entities, partition_name, fields_info)
File "pymilvus/client/prepare.py", line 1249, in batch_insert_param
    return cls._parse_batch_request(...)
File "pymilvus/client/prepare.py", line 1225, in _parse_batch_request
    field_data = entity_helper.entity_to_field_data(...)
File "pymilvus/client/entity_helper.py", line 759, in entity_to_field_data
    raise DataNotMatchException(...)
```

wangting0128:
## Problem
`entity_to_field_data()` calls `len(entity_values)` on `scipy.sparse.csr_matrix`, which raises `TypeError: sparse array length is ambiguous; use getnnz() or shape[0]`.

This was introduced in PR #3141 (commit 67b50168) which added `if len(entity_values) > 0:` guard for all vector types, but `scipy.sparse` matrices do not support `len()`.

## Fix
Use `entity_values.shape[0]` for scipy sparse types and `len()` for others to get the entity count, then check if greater than 0. This keeps the empty-value guard consistent for both scipy sparse and non-scipy inputs.

Fixes #3288

Signed-off-by: wangting0128 <ting.wang@zilliz.com>

### DIFF
--- a/pymilvus/client/entity_helper.py
+++ b/pymilvus/client/entity_helper.py
@@ -770,7 +770,12 @@ def entity_to_field_data(entity: Dict, field_info: Any, num_rows: int) -> schema
                 field_data.vectors.dim = field_info.get("params", {}).get("dim", 0)
             field_data.vectors.bfloat16_vector = b"".join(entity_values)
         elif entity_type == DataType.SPARSE_FLOAT_VECTOR:
-            if len(entity_values) > 0:
+            entity_len = (
+                entity_values.shape[0]
+                if SciPyHelper.is_scipy_sparse(entity_values)
+                else len(entity_values)
+            )
+            if entity_len > 0:
                 field_data.vectors.sparse_float_vector.CopyFrom(sparse_rows_to_proto(entity_values))
         elif entity_type == DataType.INT8_VECTOR:
             if len(entity_values) > 0:


wangting0128:
@marcelo-cjl @zhengbuqian 
could you please review whether we need to consider scenarios where SciPy sparse is empty?

pymilvus/client/entity_helper.py
```diff
@@ -770,7 +770,7 @@ def entity_to_field_data(entity: Dict, field_info: Any, num_rows: int) -> schema
                 field_data.vectors.dim = field_info.get("params", {}).get("dim", 0)
             field_data.vectors.bfloat16_vector = b"".join(entity_values)
         elif entity_type == DataType.SPARSE_FLOAT_VECTOR:
-            if len(entity_values) > 0:
+            if SciPyHelper.is_scipy_sparse(entity_values) or len(entity_values) > 0:
```
marcelo-cjl: For the scipy.sparse type, use SciPyHelper.is_scipy_sparse(entity_values) and entity_values.shape[0] > 0 to determine if it is not null, so that all logic is consistent?

PR #3295, opened 2026-02-26
fix: support explicit db_name in snapshot APIs for cross-database operations

zhuwenxing:
## Summary

- Add explicit `db_name` parameter to 4 snapshot APIs (`create_snapshot`, `list_snapshots`, `restore_snapshot`, `list_restore_snapshot_jobs`) across all layers (Prepare, GrpcHandler, AsyncGrpcHandler, MilvusClient, AsyncMilvusClient)
- The `db_name` is now populated in the protobuf request body, enabling cross-database snapshot operations (e.g., restore a snapshot from `db_a` into `db_b`)
- This complements #3292 which added gRPC metadata context — now `db_name` is sent via **both** proto body and metadata header

### Why both proto body and metadata?

| Path | When it works |
|------|--------------|
| Proto body `db_name` | When user passes explicit `db_name` parameter (cross-db operations) |
| Metadata header `dbname` (via `CallContext`) | When user omits `db_name`, falls back to client's `self._db_name` via `DatabaseInterceptor` |

### Affected APIs
| API | `db_name` in proto body |
|-----|------------------------|
| `create_snapshot` | added |
| `list_snapshots` | added |
| `restore_snapshot` | added |
| `list_restore_snapshot_jobs` | added |

### Example: cross-database restore
```python
client_a = MilvusClient(uri="...", db_name="db_a")
# Create snapshot in db_a
client_a.create_snapshot(collection_name="my_col", snapshot_name="snap1")

# Restore to a different database
client_a.restore_snapshot(
    snapshot_name="snap1",
    collection_name="restored_col",
    db_name="db_b",  # explicit target database
)
```

## Test plan

- [x] Unit tests: 29 snapshot tests pass (including 4 new cross-db tests)
- [x] Integration tests: 36 scenarios pass against Milvus master (`338bac3`), covering:
  - 3 databases with different collections/snapshots
  - `create_snapshot` with explicit `db_name` from default-db client
  - `list_snapshots` per database with `db_name` filter
  - Cross-db restore: `db_a` → `db_b`, `db_a` → `db_c` (empty db)
  - Same-db restore: `db_b` → `db_b`
  - Same-db restore without explicit `db_name` (uses client's db via metadata)
  - Data isolation verification across all 3 databases

### DIFF
--- a/pymilvus/client/async_grpc_handler.py
+++ b/pymilvus/client/async_grpc_handler.py
@@ -2499,6 +2499,7 @@ class AsyncGrpcHandler:
         collection_name: str,
         snapshot_name: str,
         description: str = "",
+        db_name: str = "",
         timeout: Optional[float] = None,
         context: Optional[CallContext] = None,
         **kwargs,
@@ -2507,6 +2508,7 @@ class AsyncGrpcHandler:
             snapshot_name=snapshot_name,
             collection_name=collection_name,
             description=description,
+            db_name=db_name,
         )
         status = await self._async_stub.CreateSnapshot(
             request, timeout=timeout, metadata=_api_level_md(context)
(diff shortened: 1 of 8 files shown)


XuanYang-cn:
I think two db_names might not work in Milvus, the API in pymilvus is explicitly designed to omit db_name to reduce the complexity, that all(at least majority) API doesn't need db_name.

XuanYang-cn:
This is a good catch though, APIs are not aligned with each other on how to pass in db_name. For now, we tends to set the db_name in interceptors only.

XuanYang-cn:
snapshot will be bind with collection_name and the API will be redesigned soon. Let's make db_name in PyMIlvus consists with other APIs.

PR #3324, opened 2026-03-19
fix: correct parameter name typos in QueryIterator causing partition filter bypass

alohaha22:
## Summary

- Fix typo in `QueryIterator.__seek_to_offset` and `__setup_ts_by_request`: `output_field` → `output_fields`, `partition_name` → `partition_names`
- The singular parameter names didn't match `grpc_handler.query()`'s signature, so they fell into `**kwargs` and were silently ignored
- This caused the seek phase to scan **all partitions** instead of the target one when both `partition_names` and `offset` are specified, producing an incorrect PK cursor

## Bug Details

When `query_iterator` is called with both `partition_names` and `offset`:

1. The seek phase queries all partitions instead of the specified one (because `partition_names=None` on the server side short-circuits the partition filter)
2. The PK cursor is set based on global data, not the target partition's data
3. `next()` correctly filters by partition (it already uses plural names), but the cursor position is wrong
4. The returned row count does not match expectations

For example, with `partition_names=["part_b"]` and `offset=50` on a partition with 100 rows (PKs 100–199):
- **Before fix**: returns 100 rows (expected 50)
- **After fix**: returns 50 rows (correct)

## Changes

- `pymilvus/orm/iterator.py`: Fix 4 parameter names (2 in `__seek_to_offset`, 2 in `__setup_ts_by_request`)
- `tests/orm/test_iterator.py`: Add `TestQueryIteratorPartitionNamesPassthrough` with 4 tests verifying correct parameter names are passed through to `conn.query()`

## Test plan

- [x] New tests fail before the fix (confirmed via TDD red phase)
- [x] New tests pass after the fix
- [x] Full iterator test suite passes (136/136)

### DIFF
--- a/pymilvus/orm/iterator.py
+++ b/pymilvus/orm/iterator.py
@@ -143,8 +143,8 @@ class QueryIterator:
                 res = self._conn.query(
                     collection_name=self._collection_name,
                     expr=expr,
-                    output_field=[],
-                    partition_name=self._partition_names,
+                    output_fields=[],
+                    partition_names=self._partition_names,
                     timeout=self._timeout,
                     **seek_params,
                 )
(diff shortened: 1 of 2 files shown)


XuanYang-cn (approved):
/lgtm

Issue #3328, opened 2026-03-20
[FEATURE]: support query ORDER BY in pymilvus client


MrPresent-Han:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Milvus server supports ORDER BY clauses in query requests, but the pymilvus client does not expose this capability through its `query()` API. Users cannot sort query results by scalar fields (e.g., `price ASC`, `rating DESC`) without post-processing on the client side.

### Describe the solution you'd like

Add `order_by` parameter support to the pymilvus query API (`MilvusClient.query()` and `Prepare.query_request()`), allowing users to specify ordering fields and directions. The parameter should accept:
- A list of strings: `["price:asc", "rating:desc"]`
- A list of dicts: `[{"field": "price", "order": "asc"}]`
- A plain string: `"price:asc,rating:desc"`

Also improve the existing search `order_by_fields` parameter to accept the same flexible formats.

### Describe alternatives you've considered

Client-side sorting after query, but this is inefficient and doesn't leverage server-side optimization.

### Anything else?

Related: #3217 (search ORDER BY support)

Issue #3217, opened 2026-01-21
[FEATURE]: support search order by


MrPresent-Han:
### Is there an existing issue for this?

- [x] I have searched the existing issues

XuanYang-cn:
Resolved by #3218 (backported to 2.6 via #3241) and #3329. All PRs merged. Closing.

PR #3329, opened 2026-03-20
feat: support query ORDER BY in pymilvus client

MrPresent-Han:
## Summary
- Add `order_by` parameter support to `Prepare.query_request()` for query ORDER BY
- Improve `order_by_fields` in search to accept flexible formats (list of str, list of dict, plain str)
- Add example script `examples/query_order_by.py`

issue: #3328
Related: #3217

## Test plan
- [ ] Run existing unit tests to ensure no regression
- [ ] Test query with `order_by` parameter using different formats (str list, dict list, plain str)
- [ ] Test search `order_by_fields` with the new string format support
- [ ] Verify example script `examples/query_order_by.py` works against a running Milvus instance

### DIFF
--- a/examples/query_order_by.py
+++ b/examples/query_order_by.py
@@ -1,0 +1,277 @@
+import numpy as np
+from pymilvus import (
+    FieldSchema, CollectionSchema, DataType,
+)
+from pymilvus.milvus_client import MilvusClient
+import random
+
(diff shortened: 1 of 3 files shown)


XuanYang-cn:
3.0 only, no need to backport

PR #3331, opened 2026-03-24
fix: handle None response in search to prevent AttributeError

wangting0128:
## What

Three fixes to prevent `AttributeError: 'NoneType' object has no attribute 'status'` in search operations:

1. **`asynch.py` logical bug fix**: Change `if self._results:` → `if self._results is not None:` in `Future.result()`. `SearchResult` inherits `list`, so an empty (0-hit) result is falsy — the old check incorrectly fell through to re-call `on_response(self._response)`.

2. **`asynch.py` defensive None guard**: `SearchFuture.on_response()` now raises a clear `MilvusException` if `response is None`, instead of letting `response.status` raise `AttributeError`.

3. **`grpc_handler.py` defensive None guard**: `_execute_search()` now checks `response is None` after `stub.Search()` before calling `check_status(response.status)`.

## Why

Observed in production: with `nq=10000, timeout=60s`, the Milvus proxy task queue delayed the request by **59 seconds**, leaving only 0.857s of deadline. The first call received `DEADLINE_EXCEEDED`; `retry_on_rpc_failure` retried. During the retry, `stub.Search()` returned `None` under server load conditions — causing `check_status(None.status)` → `AttributeError` → wrapped by `error_handler` as:

```
MilvusException: (code=1, message=Unexpected error, message=<'NoneType' object has no attribute 'status'>)
```

Fix 3 directly addresses the production error. Fixes 1 & 2 address the async path and the logical bug with empty results.

Related: https://github.com/milvus-io/milvus/issues/43341

## Test

All existing tests pass:
```
pytest tests/test_asynch.py -v  # 40 passed
```

### DIFF
--- a/pymilvus/client/asynch.py
+++ b/pymilvus/client/asynch.py
@@ -3,6 +3,8 @@ import inspect
 import threading
 from typing import Any, Callable, Optional
 
+import grpc
+
 from pymilvus.exceptions import MilvusException
 from pymilvus.grpc_gen import milvus_pb2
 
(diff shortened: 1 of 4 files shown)


XuanYang-cn (approved):
LGTM with one nit.

The three fixes are correct and well-tested. The `is not None` fix for empty `SearchResult` (falsy `list` subclass) is the most valuable — it's a real logic bug.

One gap: `_execute_hybrid_search` (`grpc_handler.py:1156-1159`) has the identical unguarded `check_status(response.status)` pattern. Same gRPC unary call mechanism, same risk — could you add the None guard there too for consistency?

XuanYang-cn (approved):
/lgtm

PR #3375, opened 2026-04-01
fix: include password in connection alias to prevent incorrect reuse

3em0:
## Summary

- **Bug**: When `user`/`password` auth is used, only the username is included in the connection alias (`auth_fmt = f"{user}"`). This means two `MilvusClient` instances with the same URI and username but **different passwords** produce the same alias, causing the second to silently reuse the first connection without re-authenticating.
- **Fix**: Hash both `user` and `password` together (via MD5, consistent with the existing `token` branch) so different credentials always produce distinct aliases.
- **Tests**: Updated existing tests and added `test_different_passwords_produce_different_aliases` to cover this case.

## Reproduction

```python
# Same process, same URI, same user, different passwords
client1 = MilvusClient(uri="localhost:19530", user="alice", password="correct")
# → alias = "localhost:19530-alice" → new connection, auth succeeds

client2 = MilvusClient(uri="localhost:19530", user="alice", password="wrong")
# → alias = "localhost:19530-alice" → has_connection() = True → reuses client1!
# No authentication error despite wrong password
```

After this fix, the two calls produce different aliases (based on `md5("alice:correct")` vs `md5("alice:wrong")`), so the second call correctly triggers a new `connections.connect()`.

## Test plan

- [x] All 17 tests in `test_create_connection.py` pass
- [x] New test `test_different_passwords_produce_different_aliases` verifies the fix

### DIFF
--- a/pymilvus/milvus_client/_utils.py
+++ b/pymilvus/milvus_client/_utils.py
@@ -29,7 +29,9 @@ def create_connection(
 
         auth_fmt = ""
         if user:
-            auth_fmt = f"{user}"
+            md5 = hashlib.new("md5", usedforsecurity=False)
+            md5.update(f"{user}:{password}".encode())
+            auth_fmt = md5.hexdigest()
         elif token:
             md5 = hashlib.new("md5", usedforsecurity=False)
             md5.update(token.encode())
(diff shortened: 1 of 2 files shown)


3em0:
@longjiquan @mausch

XuanYang-cn:
@3em0 Thanks for the contribution. But this functions is deprecated. In the latest code , MilvusClient and AsyncMilvusClient go through a complete new ConnectionManager, not sharing alias at all.

3em0:
ok，i will close it

PR #3376, opened 2026-04-02
fix: pass context to file_resource methods and add async variants

XuanYang-cn:
## Summary

- `MilvusClient.add_file_resource`, `remove_file_resource`, and `list_file_resources` were not accepting `**kwargs` or passing `context=self._generate_call_context(**kwargs)` to the grpc handler, unlike every other method in the class
- All three methods were entirely absent from `AsyncMilvusClient`, leaving async users with no access to file resource APIs
- The async grpc handler already had the implementations (`async_grpc_handler.py`); only the client layer was missing

## Changes

- Add `**kwargs` + `context=self._generate_call_context(**kwargs)` to the three methods in `MilvusClient`
- Implement `add_file_resource`, `remove_file_resource`, `list_file_resources` async variants in `AsyncMilvusClient`
- Add context-passing tests for all six method calls (3 sync, 3 async)

See also: #3356

### DIFF
--- a/pymilvus/milvus_client/async_milvus_client.py
+++ b/pymilvus/milvus_client/async_milvus_client.py
@@ -2142,3 +2142,45 @@ class AsyncMilvusClient(BaseMilvusClient):
             context=self._generate_call_context(**kwargs),
             **kwargs,
         )
+
+    async def add_file_resource(
+        self,
+        name: str,
+        path: str,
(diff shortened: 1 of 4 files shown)


XuanYang-cn:
Closing: PR was filed under wrong identity (user account instead of pymilvus-bot). Will re-file correctly.

PR #3406, opened 2026-04-13
feat: add drop_collection_field and drop_collection_function APIs

sijie-ni-0214:
## Summary
- Add `drop_collection_field` API to drop a field by name or ID from a collection
- Add `drop_collection_function` API to drop a function (with cascade deletion of output fields and indexes)
- Both APIs internally call `AlterCollectionSchema` RPC with `DropRequest`
- Update proto submodule to include `function_name` support in `DropRequest`

## Compatibility

This PR changes the underlying RPC for the existing `drop_collection_function` API:
- Before: `DropCollectionFunction` gRPC method
- After: `AlterCollectionSchema` gRPC method

As a result, this version of pymilvus requires a Milvus server that implements
`AlterCollectionSchema` with `DropRequest` (see linked milvus PR). Calling
`drop_collection_function` against an older Milvus will fail with an
`UNIMPLEMENTED` error. `drop_collection_field` is new and only works on the
new server.

RBAC migration: this is intended behavior for Milvus 3.0. Because
`drop_collection_function` is now backed by `AlterCollectionSchema`,
authorization follows the new `AlterCollectionSchema` privilege on the target
collection instead of the old global `CreateCollection` privilege used by
`DropCollectionFunction`. Roles that only grant the old privilege for this API
need to grant `AlterCollectionSchema` on the affected collections.

issue: https://github.com/milvus-io/milvus/issues/48983
pr: https://github.com/milvus-io/milvus-proto/pull/583

Signed-off-by: sijie-ni-0214 <sijie.ni@zilliz.com>


pymilvus/client/grpc_handler.py
```diff
@@ -461,21 +461,65 @@ def add_collection_field(
         check_status(status)
 
     @retry_on_rpc_failure()
-    def drop_collection_function(
+    def alter_collection_schema_drop(
```
czs007: **File:** PR description + `pymilvus/client/grpc_handler.py` and
`pymilvus/client/async_grpc_handler.py`

The existing `drop_collection_function` was calling the standalone gRPC
method `DropCollectionFunction`. After this PR it calls the new
`AlterCollectionSchema` gRPC method. Effect on upgrade:

- Old pymilvus + old Milvus: OK
- **New pymilvus + old Milvus (no `AlterCollectionSchema` method): breaks
  `drop_collection_function` with `UNIMPLEMENTED` / `method not found`**
- New pymilvus + new Milvus: OK

The switch is an implicit, undocumented compatibility break on an
existing public API. The PR description currently says "Both APIs internally call `AlterCollectionSchema` RPC with `DropRequest`", which is accurate but does not call out that this is a **change in behavior** for a method that already existed.

pymilvus/client/grpc_handler.py
```diff
@@ -461,21 +461,65 @@ def add_collection_field(
...
+            field_name=field_name,
+            field_id=field_id,
+            function_name=function_name,
+            db_name=db_name,
+        )
+        response = self._stub.AlterCollectionSchema(
             request, timeout=timeout, metadata=_api_level_md(context)
         )
-        check_status(status)
+        check_status(response.alter_status)
```
czs007, line 491: The `AlterCollectionSchemaResponse` protobuf carries two status fields:
`alter_status` and `index_status`. The client only checks the first.

In the current server (`milvus#48988`), the cascade index drop lives
inside the ack callback — a cascade failure fails the whole ack, so
`alter_status` reflects it and the second field is either empty or
redundant:

- The proto having two separate fields signals the server may populate
  them independently in the future.
- If a later Milvus version does split them (e.g., for partial success
  reporting), this client will silently miss index-side failures.

**Fix:** either defensively check both, or document the decision.

```python
check_status(response.alter_status)
if response.HasField("index_status"):
    check_status(response.index_status)
```

pymilvus/client/grpc_handler.py
czs007: alter_collection_schema_drop is an internal dispatcher; users should reach it via drop_collection_field or drop_collection_function. The name has no leading underscore, suggesting it is part of the handler's public surface. Either rename to _alter_collection_schema_drop, or document that it is an internal utility and users should prefer the specialized wrappers.

XuanYang-cn:
3.0 only, no need to backport to 2.6

sijie-ni-0214:
Handled as intended behavior for Milvus 3.0.

`drop_collection_function` now follows `AlterCollectionSchema` authorization, so callers need `AlterCollectionSchema` privilege on the target collection. I updated the PR compatibility section to call out the RBAC migration from the old global `CreateCollection` privilege.

XuanYang-cn (approved):
/lgtm

Issue #2917, opened 2025-07-31
[Bug]: type support issue : unrecognized dtype for key (WindwosPath、UUID)


frechi9:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

I'm encountering a type support issue when using Milvus to insert data with metadata fields that include `WindowsPath` (from `pathlib.Path`) and `UUID` (from Python's `uuid.UUID`). The error message reads:

```
type support issue: unrecognized dtype for key (WindowsPath, UUID)
```

These are common Python types used for file path handling and unique identifiers. I would like to request that Milvus support these types — either through automatic conversion to string (e.g., via `str()`) during metadata processing, or by allowing users to register custom type converters.

Supporting these native Python types would improve compatibility and streamline the development experience.

Is there a recommended way to handle this currently? Or is support for these types planned?

### Steps/Code To Reproduce behavior

```markdown

```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

pangwangshu:
Picking this up. Went through the discussion on #3408 first — agree with @XuanYang-cn's diagnosis that dtype inference alone doesn't close this out, since `convert_to_str_array`/`entity_to_str_arr` in `entity_helper.py` independently reject non-`str` VARCHAR values regardless of what `infer_dtype_bydata` returns.

Planning a smaller version of the `StrCoercerRegistry` idea rather than the full registry: since there are only two known offending types right now (`uuid.UUID`, `os.PathLike`), I'll add direct `isinstance` coercion at the two places that actually matter —

- `orm/types.py`: recognize `UUID`/`PathLike` as `VARCHAR` in `infer_dtype_by_scalar_data`, and catch `TypeError` (not just `IndexError`) on the `data[0]` probe in `infer_dtype_bydata`, so `PurePosixPath` no longer crashes it outright.
- `entity_helper.py`: coerce at the `convert_to_str_array`/`entity_to_str_arr` boundary, same as the registry's default registrations would (`UUID → str`, `PathLike → os.fspath`).

[...]

PR #3408, opened 2026-04-14
fix: infer uuid.UUID and os.PathLike as VARCHAR in dtype inference

korbonits:
## Summary

- `uuid.UUID` values returned `DataType.UNKNOWN` because `infer_dtype_by_scalar_data` had no branch for them (pandas considers UUID scalar, so `infer_dtype_bydata` delegates to `infer_dtype_by_scalar_data`, which fell through to `return DataType.UNKNOWN`).
- `os.PathLike` objects (e.g. `WindowsPath`, `PurePosixPath`) caused an unhandled `TypeError` crash in `infer_dtype_bydata` — only `IndexError` was caught on the `data[0]` subscript probe.

Both types represent string-serializable values and should map to `DataType.VARCHAR`.

issue: #2917

## Changes

**`pymilvus/orm/types.py`**
- Added `isinstance(data, (uuid.UUID, os.PathLike))` → `DataType.VARCHAR` in `infer_dtype_by_scalar_data`.
- Added the same guard near the top of `infer_dtype_bydata` (before the list/array branch) to catch PathLike objects that pandas does not consider scalar.
- Broadened the subscript probe `except` to catch `TypeError` in addition to `IndexError`.

**`tests/test_types.py`**
- Added `test_infer_dtype_bydata_uuid_and_pathlike` covering fixed UUID, PurePosixPath, and PureWindowsPath cases.

## Test plan

- [x] `python3 -m pytest tests/test_types.py` — 39 passed (35 pre-existing + 4 new)
- [x] All existing parametrized `test_infer_dtype_bydata` cases still pass

### DIFF
--- a/pymilvus/orm/types.py
+++ b/pymilvus/orm/types.py
@@ -10,6 +10,8 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 
+import os
+import uuid
 from typing import Any
 
 import numpy as np
@@ -92,6 +94,8 @@ def infer_dtype_by_scalar_data(data: Any, dtype: DataType = None):
         return DataType.VARCHAR
     if isinstance(data, bytes):
         return DataType.BINARY_VECTOR
+    if isinstance(data, (uuid.UUID, os.PathLike)):
+        return DataType.VARCHAR
 
     return DataType.UNKNOWN
 
@@ -100,6 +104,9 @@ def infer_dtype_bydata(data: Any):
     d_type = DataType.UNKNOWN
     if is_scalar(data):
         return infer_dtype_by_scalar_data(data)
+
+    if isinstance(data, (uuid.UUID, os.PathLike)):
+        return DataType.VARCHAR
 
     if isinstance(data, dict):
         return DataType.JSON
@@ -121,7 +128,7 @@ def infer_dtype_bydata(data: Any):
     if d_type == DataType.UNKNOWN:
         try:
             elem = data[0]
-        except IndexError:
+        except (IndexError, TypeError):
             elem = None
 
         if elem is not None and is_scalar(elem):
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -9,6 +9,9 @@
 # is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
 # or implied. See the License for the specific language governing permissions and limitations under the License.
 
+import uuid
+from pathlib import PurePosixPath, PureWindowsPath
+
 import numpy as np
 import pytest
 from pymilvus import DataType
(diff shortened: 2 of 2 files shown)


XuanYang-cn:
Nice start, but I think this PR is **incomplete** — it only patches dtype inference. The real rejection happens later, at `convert_to_str_array` (`entity_helper.py:181`), which enforces `isinstance(s, str)` strictly:

```python
client.insert("col", [{"k": uuid.uuid4()}])
# ParamError: field (k) expects string input, got: <class 'uuid.UUID'>
```

Any e2e insert test would catch this. #2917 explicitly asked for `str()` conversion during processing — which this PR doesn't deliver.

### Proposal: a `StrCoercerRegistry` at the serialization boundary

Rather than per-type patches to inference, coerce at the single place VARCHAR values are serialized. As a registry class it's thread-safe, dup-detecting, and extensible for the next stringy type (`Decimal`, `IPv4Address`, `enum.StrEnum`, domain IDs):

```python
from pymilvus import str_coercer
str_coercer.register(Decimal)
str_coercer.register(MyId, parser=lambda x: x.to_wire())
```

Apply at the two VARCHAR call sites in `entity_helper.py`:
```python
# row path (~L667)
convert_to_str_array(str_coercer.coerce(field_value), field_info, CHECK_STR_ARRAY)
# column path (~L791)
entity_to_str_arr([str_coercer.coerce(v) for v in entity_values], field_info, CHECK_STR_ARRAY)
```

Default registrations: `uuid.UUID → str`, `os.PathLike → os.fspath`. Unregistered types still hit the loud `ParamError` — no blanket `str(anything)`.

### Verified end-to-end
```
[INSERT row]   ['12345678-...', '/foo/bar', 'plain-str']   ✓
[UPSERT col]   ['12345678-...', '/foo/bar', 'plain-str']   ✓
[safety]       dict → DataNotMatchException                 ✓
```
Insert + upsert share this path. Filter expressions and vector `data=` are out of scope.

### Why better than the current approach
1. Actually fixes #2917 (serializer, not just inference).
2. Extensible — no pymilvus PR needed per new type.
3. Thread-safe with explicit dup detection.
4. Safe — unknown types still rejected.
5. Smaller net diff — `types.py` changes become unnecessary.

Happy to open a follow-up PR with the registry + e2e tests if this direction works.

korbonits:
Thanks for the thorough analysis @XuanYang-cn — you're right that this PR is incomplete. Dtype inference gets the value past schema resolution, but `convert_to_str_array` at `entity_helper.py:181` still rejects non-str with `ParamError`, so an end-to-end insert of `uuid.UUID` / `PathLike` still fails. That's exactly what #2917 asked to be fixed.

Your `StrCoercerRegistry` design is the right approach — coercing once at the serialization boundary, with explicit opt-in registration, preserves the loud-reject behavior for unknown types while cleanly extending to `Decimal`, `IPv4Address`, domain IDs, etc. Much better than per-type patches to the inference layer.

Closing this in favor of your follow-up PR. Happy to review or test against UUID/PathLike once you open it — please tag me. Thanks again for taking the time to write up the full design.

Issue #2985, opened 2025-09-15
[Bug]: The value of consistency_level returned by describe collection is a number rather than a corresponding semantic term, which is not user-friendly.


zhuwenxing:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

` 'consistency_level': 0`

```
{'collection_name': 'client_geometry_3yiyc8Bb', 'auto_id': False, 'num_shards': 1, 'description': 'test geometry collection', 'fields': [{'field_id': 100, 'name': 'id', 'description': '', 'type': <DataType.INT64: 5>, 'params': {}, 'is_primary': True}, {'field_id': 101, 'name': 'vector', 'description': '', 'type': <DataType.FLOAT_VECTOR: 101>, 'params': {'dim': 128}}, {'field_id': 102, 'name': 'geo', 'description': '', 'type': <DataType.GEOMETRY: 24>, 'params': {}}], 'functions': [], 'aliases': [], 'collection_id': 460756692454159104, 'consistency_level': 0, 'properties': {}, 'num_partitions': 1, 'enable_dynamic_field': False, 'created_timestamp': 460829215984713733, 'update_timestamp': 460829215984713733}
```

### Steps/Code To Reproduce behavior

```markdown

```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory):
- Method of installation (Docker, or from source):
- Milvus version (v0.3.1, or v0.4.0):
- Milvus configuration (Settings you made in `server_config.yaml`):
```

### Anything else?

 pymilvus version: 2.5.16rc3

PR #3409, opened 2026-04-14
fix: return consistency_level as name string from describe_collection

korbonits:
## Summary

`CollectionSchema.dict()` exposed the raw protobuf integer for `consistency_level` (e.g. `0`) instead of its human-readable name (e.g. `"Strong"`), making the output of `describe_collection()` difficult to interpret without consulting the enum definition.

issue: #2985

## Changes

**`pymilvus/client/abstract.py`**
- Import `ConsistencyLevel` alongside `DataType` and `FunctionType`.
- In `CollectionSchema.dict()`, replace `self.consistency_level` with `ConsistencyLevel.Name(self.consistency_level)`.
- `self.consistency_level` keeps its integer value internally so existing search/query code paths that forward it to gRPC are unaffected.

**`tests/test_client_abstract.py`**
- Added `test_collection_schema_dict_consistency_level_is_string` covering all five enum values.

## Before / After

```python
# Before
client.describe_collection("my_col")
# {'consistency_level': 0, ...}

# After
client.describe_collection("my_col")
# {'consistency_level': 'Strong', ...}
```

## Test plan

- [x] `python3 -m pytest tests/test_client_abstract.py tests/test_types.py` — 116 passed

### DIFF
--- a/pymilvus/client/abstract.py
+++ b/pymilvus/client/abstract.py
@@ -13,7 +13,7 @@ from .constants import DEFAULT_CONSISTENCY_LEVEL, RANKER_TYPE_RRF, RANKER_TYPE_W
 # ruff: noqa: F401
 # TODO: This is a patch for older version
 from .search_result import Hit, Hits, SearchResult
-from .types import DataType, FunctionType
+from .types import ConsistencyLevel, DataType, FunctionType
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


XuanYang-cn:
Quick note on the tradeoff here before merging:

This is a **backward-incompatible change to a public return value** — `describe_collection()["consistency_level"]` flips from `int` to `str`. User code like `resp["consistency_level"] == 0` or `== ConsistencyLevel.Strong` will silently change behavior, no exception raised.

- **Readability (win):** `'Strong'` is much more useful than `0` for humans inspecting the dict.
- **Compatibility (cost):** public return type changes without a deprecation window.

Internal pymilvus consumers (schema cache, insert/upsert prep, replicas, iterators, ORM `Collection.__init__`) are unaffected — verified none of them break on the type change. The risk is entirely on downstream users.

Also — please run `make format` on the branch before merging to keep CI green. I'll call the compat break out explicitly in the release notes.

korbonits:
@XuanYang-cn thank you so much for the extremely thoughtful feedback across multiple PRs -- I will grok each and work through them 🙏

XuanYang-cn (approved):
/lgtm

Issue #3407, opened 2026-04-14
[Bug]: Pymilvus Hybrid Search Filter not working when given expr in the AnnSearchRequest


czhao314:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

I'm using Milvus 2.5.12 hosted on watsonx.data (IBM product), and pymilvus 2.5.12.
When I run a hybrid search against my collection and try to filter both the dense vector and sparse vector search using an expression that filters for a specific document name, the expressions do not apply and return passages from all documents.

In the regular dense vector search, the filter expression I'm using works exactly as expected, so it seems strange that it's not filtering my outputs in hybrid search. I've tried the fuzzy match and exact match filter expressions and both seem to not apply the filter at all. 

I've tried upgrading my pymilvus to the latest vers as well (2.6.12), but no change from the previous version used. 

### Expected Behavior

When I run hybrid search with a query in my VDB and filter by document_name X_X_X.pdf with the expression 'document_name like "%{ X_X_X.pdf}%"', it should return only results from chunks that were from document_name  X_X_X.pdf. 

### Steps/Code To Reproduce behavior

```markdown
I'm searching for information in different insurance documents. document_name is just the filename (like X_X_X.pdf). I only want to get results from a single document. 
This is my simple code to run hybrid search based off the docs here(https://milvus.io/docs/multi-vector-search.md#Multi-Vector-Hybrid-Search)


limit = 5

def run_single_document_search(
    query,
    collection,
    embeddings_model,
    document_name,
    limit)

    """Run hybrid search for a single document."""
    dense_query_vector = embeddings_model.embed_query(query)
    expression = 'document_name like "%{document_name}%"'

    dense_req = AnnSearchRequest(
        data=[dense_query_vector],
        anns_field="vector_embeddings",
        param={"metric_type": "COSINE", "params": {"nprobe": 10}},
        limit=limit,
        expr=expression,
    )

    sparse_req = AnnSearchRequest(
        data=[query],
        anns_field="sparse_embeddings",
        param={"metric_type": "BM25", "params": {}},
        limit=limit,
        expr=expression,
    )

    results = collection.hybrid_search(
        reqs=[dense_req, sparse_req],
        rerank=RRFRanker(),
        limit=limit,
        output_fields=["document_id", "document_name", "text"],
        filter =expression
    )
```

### Environment details

```markdown
- Mac OS
- Installed Pymilvus using Pip
- Milvus version: 2.5.12
```

[...]

Issue #2664, opened 2025-02-21
[Enhancement]: HybridSearch enable to use filter instead of expr


XuanYang-cn:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Unify the expression params name

PR #3410, opened 2026-04-14
fix: accept filter= as alias for expr= in AnnSearchRequest

korbonits:
## Summary

Every other pymilvus API surface (`search`, `query`, `delete`, top-level `hybrid_search`) uses `filter=` for scalar filtering. `AnnSearchRequest` was the sole exception, requiring `expr=`. This inconsistency caused silent no-ops when users passed `filter=` as a keyword — it was silently swallowed by `**kwargs` and ignored — leading to hybrid searches that returned unfiltered results (reported in issues #3407 and #2664).

issue: #2664

## Changes

**`pymilvus/client/abstract.py`**
- Added `filter: Optional[str] = None` parameter to `AnnSearchRequest.__init__`.
- If only `filter=` is given, it is stored as `self._expr` — gRPC layer unchanged.
- If only `expr=` is given, behaviour is identical to before.
- If both are given, `ValueError` is raised.
- Added `filter` property returning `self._expr` alongside the existing `expr` property.

**`tests/test_client_abstract.py`**
- `test_ann_search_request_with_filter` — `filter=` sets `expr` and `filter` correctly.
- `test_ann_search_request_filter_and_expr_raises` — providing both raises `ValueError`.
- `test_ann_search_request_invalid_filter_type` — non-str `filter=` raises `DataTypeNotMatchException`.

## Before / After

```python
# Before — silently ignored, no filtering applied
AnnSearchRequest(data, "vec", param, 10, filter="age > 30")

# After — works correctly, consistent with the rest of the API
AnnSearchRequest(data, "vec", param, 10, filter="age > 30")
AnnSearchRequest(data, "vec", param, 10, expr="age > 30")   # still works
```

## Test plan

- [x] `python3 -m pytest tests/test_client_abstract.py tests/test_bytes_vector_search.py` — 102 passed

### DIFF
--- a/pymilvus/client/abstract.py
+++ b/pymilvus/client/abstract.py
@@ -505,15 +505,19 @@ class AnnSearchRequest:
         limit: int,
         expr: Optional[str] = None,
         expr_params: Optional[dict] = None,
+        filter: Optional[str] = None,
     ):
         self._data = data
         self._anns_field = anns_field
         self._param = param
         self._limit = limit
 
-        if expr is not None and not isinstance(expr, str):
-            raise DataTypeNotMatchException(message=ExceptionsMessage.ExprType % type(expr))
-        self._expr = expr
+        if expr is not None and filter is not None:
+            raise ValueError("Provide either 'expr' or 'filter', not both.")
+        resolved = filter if expr is None else expr
+        if resolved is not None and not isinstance(resolved, str):
+            raise DataTypeNotMatchException(message=ExceptionsMessage.ExprType % type(resolved))
+        self._expr = resolved
         self._expr_params = expr_params
 
     @property
@@ -534,6 +538,10 @@ class AnnSearchRequest:
 
     @property
     def expr(self):
+        return self._expr
+
+    @property
+    def filter(self):
         return self._expr
 
     @property
--- a/pymilvus/client/abstract.py
+++ b/pymilvus/client/abstract.py
@@ -13,7 +13,7 @@ from .constants import DEFAULT_CONSISTENCY_LEVEL, RANKER_TYPE_RRF, RANKER_TYPE_W
 # ruff: noqa: F401
 # TODO: This is a patch for older version
 from .search_result import Hit, Hits, SearchResult
-from .types import DataType, FunctionType
+from .types import ConsistencyLevel, DataType, FunctionType
 
 logger = logging.getLogger(__name__)
 
@@ -327,7 +327,7 @@ class CollectionSchema:
             "functions": [f.dict() for f in self.functions],
             "aliases": self.aliases,
             "collection_id": self.collection_id,
-            "consistency_level": self.consistency_level,
+            "consistency_level": ConsistencyLevel.Name(self.consistency_level),
             "properties": self.properties,
             "num_partitions": self.num_partitions,
             "enable_dynamic_field": self.enable_dynamic_field,
--- a/pymilvus/client/abstract.py
+++ b/pymilvus/client/abstract.py
@@ -4,7 +4,7 @@ from typing import Any, Dict, List, Optional, Union
 
 import orjson
 
-from pymilvus.exceptions import DataTypeNotMatchException, ExceptionsMessage
+from pymilvus.exceptions import DataTypeNotMatchException, ExceptionsMessage, ParamError
 from pymilvus.settings import Config
 
 from . import utils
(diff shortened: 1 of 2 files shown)


pymilvus/client/abstract.py
```diff
@@ -13,7 +13,7 @@
 # ruff: noqa: F401
 # TODO: This is a patch for older version
 from .search_result import Hit, Hits, SearchResult
-from .types import DataType, FunctionType
+from .types import ConsistencyLevel, DataType, FunctionType
```
XuanYang-cn, line 16: Unrelated to this PR — the `ConsistencyLevel` import is from #3409. Please drop it when rebasing onto master.

pymilvus/client/abstract.py
```diff
@@ -327,7 +327,7 @@ def dict(self):
             "functions": [f.dict() for f in self.functions],
             "aliases": self.aliases,
             "collection_id": self.collection_id,
-            "consistency_level": self.consistency_level,
+            "consistency_level": ConsistencyLevel.Name(self.consistency_level),
```
XuanYang-cn, line 330: Unrelated to this PR — this `ConsistencyLevel.Name(...)` change is from #3409. Please revert to `self.consistency_level` here (the #3409 change will land separately).

pymilvus/client/abstract.py
```diff
@@ -505,15 +505,19 @@ def __init__(
...
         self._data = data
         self._anns_field = anns_field
         self._param = param
         self._limit = limit
 
-        if expr is not None and not isinstance(expr, str):
-            raise DataTypeNotMatchException(message=ExceptionsMessage.ExprType % type(expr))
-        self._expr = expr
+        if expr is not None and filter is not None:
+            raise ValueError("Provide either 'expr' or 'filter', not both.")
```
XuanYang-cn, line 516: Please raise `ParamError` instead of `ValueError` here — pymilvus has its own exception hierarchy (`pymilvus.exceptions.ParamError`) that is already used elsewhere for constructor misuse. Keeping this consistent matters for users who catch pymilvus exceptions by their base class.

```python
from ..exceptions import ParamError
...
if expr is not None and filter is not None:
    raise ParamError(message="Provide either 'expr' or 'filter', not both.")
```

Please update the corresponding test assertion (`pytest.raises(ParamError, ...)`) as well.

pymilvus/client/abstract.py
```diff
@@ -505,15 +505,19 @@ def __init__(
...
+        resolved = filter if expr is None else expr
```
XuanYang-cn, line 517: Please invert the precedence so `filter` is the default and `expr` is the fallback — this matches the direction the rest of the SDK is moving (`filter=` is the canonical name; `expr=` is legacy):

```python
resolved = filter if filter is not None else expr
```

Reads as "use filter, fall back to expr" and signals intent to new callers.

tests/test_client_abstract.py
```diff
@@ -516,6 +516,24 @@ def test_collection_schema_dict_without_timestamps(self):
         assert "created_timestamp" not in d
         assert "update_timestamp" not in d
 
+    def test_collection_schema_dict_consistency_level_is_string(self):
```
XuanYang-cn, line 519: This test is for the #3409 change (consistency_level in `dict()`), not for this PR. Please drop it when rebasing.

XuanYang-cn (changes requested):
Three changes requested before this can merge. See inline comments — summary:

1. **Drop the `consistency_level` hunks** — they belong to #3409, not this PR. Please rebase onto current `master` so this branch contains only the `AnnSearchRequest.filter` change.
2. **Invert the resolution order** — `filter` should be the primary, `expr` the fallback.
3. **Raise `ParamError`, not `ValueError`** — pymilvus has its own exception hierarchy for constructor misuse.

XuanYang-cn:
Also — please run `make format` on the branch before pushing the next revision to keep CI green. Thanks!

korbonits:
Thanks for the review @XuanYang-cn! All four items addressed in `b393f2bd`:

1. **Rebased onto master** — the `ConsistencyLevel` commit dropped cleanly since #3409 is now merged. This branch now contains only the `AnnSearchRequest.filter` change.
2. **Precedence inverted** — `resolved = filter if filter is not None else expr` (filter is primary, expr is the fallback).
3. **`ParamError` instead of `ValueError`** — imported from `pymilvus.exceptions`; test updated to `pytest.raises(ParamError, match=...)`.
4. **`make format`** — ran clean, no diff.

Local `tests/test_client_abstract.py::TestAnnSearchRequest` — 8 passed.

XuanYang-cn (approved):
/lgtm

PR #3427, opened 2026-04-22
enhance: Add target_size and target_size_unit in compact API

XuanYang-cn:


liliu-z:
**pymilvus/client/grpc_handler.py:2272** Positional arguments are in the wrong order. `Prepare.manual_compaction` signature is `(collection_name, is_clustering, is_l0, collection_id=None, target_size=None)`, so `response.collectionID` is being passed as `is_l0` and `target_size` as `collection_id`. Compare with the correct call in `async_grpc_handler.py:2431` and use keyword arguments or add the missing `is_l0` argument.

pymilvus/milvus_client/optimize_task.py
```diff
@@ -15,8 +15,12 @@
     "pb": 1024**5,
 }
 
+# Sentinel passed to the server when optimize() is called without a target_size,
+# meaning "let the server choose the segment size automatically".
+_OPTIMIZE_DEFAULT_SIZE_MB = 1 << 63 - 1
```
liliu-z: Operator precedence bug: `1 << 63 - 1` evaluates as `1 << (63 - 1)` = `1 << 62` (4611686018427387904), not the intended int64 max `(1 << 63) - 1` (9223372036854775807). Parentheses are needed around `(1 << 63)`.

PR #3432, opened 2026-04-22
feat: add FieldOp helpers for array partial update operators

weiliu1031:
issue: https://github.com/milvus-io/milvus/issues/49241
milvus pr: https://github.com/milvus-io/milvus/pull/49251
proto pr: https://github.com/milvus-io/milvus-proto/pull/586

## Summary

Exposes the new `UpsertRequest.field_ops` semantics in pymilvus so users can request `ARRAY_APPEND` / `ARRAY_REMOVE` on Array fields without building proto messages by hand.

### Proto regen
- Bumps `pymilvus/grpc_gen/milvus-proto` submodule to `04b6d9f`.
- Regenerates `schema_pb2` (gains `FieldPartialUpdateOp`) and `milvus_pb2` (`UpsertRequest.field_ops`).

### New public API — `pymilvus.client.field_ops`
- `FieldOpType` enum: `REPLACE` / `ARRAY_APPEND` / `ARRAY_REMOVE` (exported as `FieldOp.*()` factories).
- `normalize_field_ops(dict | None)` accepts:
  - `FieldPartialUpdateOp` proto instance
  - `FieldOpType` enum value
  - raw `int`
  - string alias (`"array_append"`, `"array_remove"`, `"replace"`)
  - `None` value → dropped
- REPLACE (all forms) → dropped from the resulting dict (equivalent to no-op).
- Unknown string → `ParamError("unknown field_op alias")`; unsupported type → `ParamError("unsupported field_op type")`.

### Prepare wiring — `pymilvus/client/prepare.py`
- `Prepare.row_upsert_param` / `Prepare.batch_upsert_param` accept `field_ops=` kwarg.
- Any non-REPLACE op auto-promotes `partial_update=True`.
- Ops attached via `request.field_ops.add(...)` — `FieldData` stays a pure data carrier.
- Unknown-field ops are forwarded so the server can return the authoritative validation error.

### gRPC handlers
- `grpc_handler` / `async_grpc_handler` thread `field_ops` through to `Prepare.*`.
- `MilvusClient.upsert` / `upsert_async` expose the kwarg.

## Test plan
- [x] `tests/prepare/test_partial_op.py` — 21 cases covering: factory/enum/string/int/None inputs, REPLACE drop-out, unknown-string error, unsupported-type error, auto-promote, unknown-field forwarding, row + batch variants, `_apply_field_ops` direct
- [x] Full `pytest tests/` suite green (3756 passed)
- [x] FieldData proto regression guard — no `partial_op` field
- [ ] CI green

### DIFF
--- a/pymilvus/client/field_ops.py
+++ b/pymilvus/client/field_ops.py
@@ -1,0 +1,128 @@
+# Licensed to the LF AI & Data foundation under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Field-level partial-update helpers.
+
+These helpers wrap :class:`schema_pb2.FieldPartialUpdateOp` so callers do not
+need to import the generated protobuf module directly. ``FieldOp.array_append``
+and ``FieldOp.array_remove`` return ready-to-use proto messages that can be
+attached to a field during ``MilvusClient.upsert``.
+"""
+
+from typing import Mapping, Optional, Union
+
+from pymilvus.exceptions import ParamError
+from pymilvus.grpc_gen import schema_pb2
+
+__all__ = ["FieldOp", "FieldOpType", "normalize_field_ops"]
+
+# Re-exported for callers that want to compare against the low-level enum
+# (e.g. in tests) without reaching into the generated module.
+FieldOpType = schema_pb2.FieldPartialUpdateOp.OpType
+
+# FieldOpsInput describes the accepted input shape of the ``field_ops`` kwarg
+# across the public upsert API. The value per field may be:
+#
+#   * a :class:`schema_pb2.FieldPartialUpdateOp` message directly;
+#   * a :class:`FieldOpType` enum value;
+#   * a string alias like ``"array_append"`` / ``"array_remove"``.
+FieldOpsInput = Optional[
+    Mapping[str, Union[schema_pb2.FieldPartialUpdateOp, "FieldOpType", str]]
+]
+
+
+class FieldOp:
+    """Factory for building :class:`FieldPartialUpdateOp` messages.
+
+    Prefer these factories over constructing the proto message directly so
+    the surface stays stable if the proto changes.
+    """
+
+    @staticmethod
+    def replace() -> schema_pb2.FieldPartialUpdateOp:
+        """Return an explicit REPLACE op — equivalent to omitting the op.
+
+        Exposed for completeness; typical callers just leave ``field_ops``
+        unset for the REPLACE (default) behavior.
+        """
+        return schema_pb2.FieldPartialUpdateOp(op=FieldOpType.REPLACE)
+
+    @staticmethod
+    def array_append() -> schema_pb2.FieldPartialUpdateOp:
+        """Return an ARRAY_APPEND op, appending the payload to the base row."""
+        return schema_pb2.FieldPartialUpdateOp(op=FieldOpType.ARRAY_APPEND)
+
+    @staticmethod
+    def array_remove() -> schema_pb2.FieldPartialUpdateOp:
+        """Return an ARRAY_REMOVE op, deleting every matching element."""
+        return schema_pb2.FieldPartialUpdateOp(op=FieldOpType.ARRAY_REMOVE)
+
+
+# Map from string alias to the concrete enum value. Kept out of the class so
+# callers can extend it in the future without touching ``FieldOp``.
+_STRING_ALIASES = {
+    "replace": FieldOpType.REPLACE,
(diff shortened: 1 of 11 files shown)


pymilvus/client/prepare.py
```diff
@@ -1287,16 +1300,38 @@ def batch_upsert_param(
         partition_name: str,
         fields_info: Any,
         partial_update: bool = False,
+        field_ops: FieldOpsInput = None,
     ):
         location = cls._pre_upsert_batch_check(entities, fields_info, partial_update)
         tag = partition_name if isinstance(partition_name, str) else ""
```
XuanYang-cn: **Blocking**: this pre-check runs with the un-promoted `partial_update`. `_pre_upsert_batch_check` (line 1229) only skips the `expected_num_input_fields` check when `partial_update=True`, so a caller doing `client.upsert(coll, [pk_values, tags_values], field_ops={"tags": FieldOp.array_append()})` against a 3-field schema (pk/vec/tags) raises `ParamError("expected number of fields: 3, actual: 2")`. The auto-promotion promise silently fails on the batch path; the row path (line 1168) uses the promoted flag correctly.

Suggested fix: move the `ops = normalize_field_ops(field_ops)` / `effective_partial_update = ...` computation above this line and pass `effective_partial_update` into `_pre_upsert_batch_check`.

Please also add a batch regression test with a 3-field schema and 2-field entities + `field_ops={...}` that asserts no `ParamError`. The current `test_batch_upsert_param_emits_field_op` cannot catch this because its fixture schema has matching arity with the entities.

pymilvus/client/field_ops.py
```diff
@@ -0,0 +1,124 @@
...
+        return None
+    if isinstance(op, schema_pb2.FieldPartialUpdateOp):
+        return op
+    if isinstance(op, str):
+        alias = op.lower()
+        if alias not in _STRING_ALIASES:
+            raise ParamError(message=f"unknown field_op alias {op!r} for field {field_name!r}")
+        return schema_pb2.FieldPartialUpdateOp(op=_STRING_ALIASES[alias])
+    # Fall back: treat it as a raw enum value if it is an int.
+    if isinstance(op, int):
```
XuanYang-cn, line 126: `isinstance(op, int)` has two traps:

1. `bool` subclasses `int`, so `field_ops={"tags": True}` silently becomes `FieldPartialUpdateOp(op=1)` = `ARRAY_APPEND` instead of raising.
2. Out-of-range ints are forwarded to the server rather than rejected here.

Suggest an explicit bool guard and a range check against `FieldOpType._value2member_map_`:

```python
if isinstance(op, bool):
    raise ParamError(message=f"unsupported field_op type bool for field {field_name!r}")
if isinstance(op, int):
    if op not in FieldOpType._value2member_map_:
        raise ParamError(message=f"invalid field_op enum value {op} for field {field_name!r}")
    return schema_pb2.FieldPartialUpdateOp(op=op)
```

pymilvus/client/field_ops.py
```diff
@@ -0,0 +1,124 @@
...
+and ``FieldOp.array_remove`` return ready-to-use proto messages that can be
+attached to a field during ``MilvusClient.upsert``.
+"""
+
+from typing import Mapping, Optional, Union
+
+from pymilvus.exceptions import ParamError
+from pymilvus.grpc_gen import schema_pb2
+
+__all__ = ["FieldOp", "FieldOpType", "normalize_field_ops"]
```
XuanYang-cn: `normalize_field_ops` is in `__all__` here but not re-exported from `pymilvus/__init__.py` (only `FieldOp` and `FieldOpType` are). If it is meant to be an internal helper, drop it from `__all__`; if it is meant to be public, add it to the top-level package exports. Right now callers who want it must reach into `pymilvus.client.field_ops`.

tests/prepare/test_partial_op.py
```diff
@@ -0,0 +1,228 @@
...
+def test_apply_field_ops_empty_is_noop():
+    req = milvus_pb2.UpsertRequest()
+    Prepare._apply_field_ops(req, {})
+    Prepare._apply_field_ops(req, None)
+    assert len(req.field_ops) == 0
+
+
+def test_field_data_message_has_no_partial_op_field():
+    """Regression guard: FieldData must remain a pure data carrier."""
+    assert "partial_op" not in schema_pb2.FieldData.DESCRIPTOR.fields_by_name
```
XuanYang-cn, line 222: This regression guard is duplicated — it also appears around line 107-111 inside a row-test. The module-scope version here is sufficient; please drop the one embedded in the row test.

XuanYang-cn (changes requested):
Thanks for the feature. One blocking correctness bug and a few nits before this can merge. Inline comments below.

## Blocking

`batch_upsert_param` runs the pre-upsert arity check with the un-promoted `partial_update` flag, so the PR's headline promise that `field_ops` auto-implies `partial_update=True` silently fails on the column (batch) path. Row path is correct. Repro and suggested fix inline at `pymilvus/client/prepare.py:1306`.

The existing batch test `tests/prepare/test_partial_op.py::test_batch_upsert_param_emits_field_op` cannot catch this because the fixture `array_schema` has 2 fields matching the 2-field entity shape. Please add a regression test with a 3-field schema + 2-field entities that asserts no `ParamError`.

## Other

- `AsyncMilvusClient.upsert` docstring (`pymilvus/milvus_client/async_milvus_client.py:476-480`) never mentions `field_ops`. It flows through `**kwargs` but is invisible to async users — please mirror the sync docstring added at `milvus_client.py:296-302`. (Not inline because the file isn't in the diff.)

Remaining nits posted as inline comments. Happy to re-review once the blocker is fixed.

XuanYang-cn:
This is 3.0 only, no need to backport to 2.6

XuanYang-cn:
2.6 needs it now. Plz cp manuallly @weiliu1031

PR #3436, opened 2026-04-23
fix: place PK cursor before user filter in QueryIterator pagination

### LINKED ISSUE #3435: [Bug]: QueryIterator is incompatible with element_filter() on struct array due to PK-cursor AND-operand order

### Describe the bug

`client.query_iterator(filter="element_filter(structA, $[...])", ...)` fails on the second and subsequent pages because `QueryIterator` appends the PK cursor on the **right** of AND, but Milvus's parser requires `element_filter()` to be the **right-most** operand of an AND expression (this constraint is by design on the server side).

Error on the second `iterator.next()` call:

```
MilvusException: (code=1100, message=failed to create query plan: cannot parse expression:
  (element_filter(structA, $[int_val] >= 20000)) and id > 211,
  error: element filter expression can only be the last expression in the logical and expression:
  invalid parameter)
```

Offending code at `pymilvus/orm/iterator.py` `QueryIterator.__setup_next_expr` (master @ `9bd39311`):

```python
def __setup_next_expr(self) -> str:
    current_expr = self._expr
    if self._next_id is None:
        return current_expr
    filtered_pk_str = ""
    if self._pk_str:
        filtered_pk_str = f'{self._pk_field_name} > "{self._next_id}"'
    else:
        filtered_pk_str = f"{self._pk_field_name} > {self._next_id}"
    if current_expr is None or len(current_expr) == 0:
        return filtered_pk_str
    return "(" + current_expr + ")" + " and " + filtered_pk_str   # <-- places user filter on the LEFT
```

The concatenation produces e.g. `"(element_filter(...)) and id > 211"`, which violates the server's "element_filter must be the right-most operand" rule. The parse rejection happens on page 2+ because page 1 has no PK cursor yet and emits the user filter unchanged.

`SearchIteratorV2` does not hit this because it uses a distance cursor (`search_params.last_bound`) and does not rewrite the filter.

### Expected Behavior

`query_iterator()` should advance through pages normally when the user-supplied filter is `element_filter()` on a struct array field — mirroring the behaviour of plain `query()` and `search_iterator()`.

Against a Milvus build that includes PR https://github.com/milvus-io/milvus/pull/49182 (element-level struct array search support):

```python
from pymilvus import MilvusClient, DataType
import random, numpy as np


def _vec(seed, dim=128):
    rng = np.random.RandomState(seed)
    v = rng.rand(dim).astype(np.float32)
    return (v / np.linalg.norm(v)).tolist()


c = MilvusClient(uri="http://<milvus>:19530")
name = "qiter_ef_repro"
c.drop_collection(name)

schema = c.create_schema(auto_id=False)
schema.add_field("id", DataType.INT64, is_primary=True)
schema.add_field("normal_vector", DataType.FLOAT_VECTOR, dim=128)
ss = c.create_struct_field_schema()
ss.add_field("embedding", DataType.FLOAT_VECTOR, dim=128)
ss.add_field("int_val", DataType.INT64)
schema.add_field(
    "structA", DataType.ARRAY, element_type=DataType.STRUCT,
    struct_schema=ss, max_capacity=100,
)

ip = c.prepare_index_params()
ip.add_index(field_name="normal_vector", index_type="HNSW", metric_type="COSINE",
             params={"M": 16, "efConstruction": 200})
ip.add_index(field_name="structA[embedding]", index_type="HNSW", metric_type="COSINE",
             params={"M": 16, "efConstruction": 200})
c.create_collection(name, schema=schema, index_params=ip)

rows = []
for i in range(500):
    r = random.Random(i)
    ne = r.randint(3, 6)
    rows.append({
        "id": i,
        "normal_vector": _vec(i + 999),
        "structA": [{"embedding": _vec(i * 100 + j), "int_val": i * 100 + j} for j in range(ne)],
    })
c.insert(name, rows)
c.flush(name)
c.load_collection(name)

# Plain query WORKS:
print(len(c.query(
    name, filter="element_filter(structA, $[int_val] >= 20000)",
    output_fields=["id"], limit=16384,
)))

# query_iterator FAILS on page 2+:
it = c.query_iterator(
    name, batch_size=50,
    filter="element_filter(structA, $[int_val] >= 20000)",
    output_fields=["id"],
)
while True:
    b = it.next()    # raises on the second call
    if not b:
        break
it.close()
```

Empirical check on the same cluster confirms the fix direction:

| Expression | Result |
| --- | --- |
| `(element_filter(structA, $[int_val] >= 0)) and id > 100` (what QueryIterator emits) | ❌ rejected by server parser |
| `id > 100 and (element_filter(structA, $[int_val] >= 0))` (swapped) | ✅ OK, 5 rows |
| `id > 100 and element_filter(structA, $[int_val] >= 0)` (swapped, no parens) | ✅ OK, 5 rows |

### Environment details

- Hardware/Software conditions: macOS client, Milvus deployed on Linux K8s 1.24.2 (Rocky Linux 8.x), 2 CPU / 8 GiB
- Method of installation: pymilvus installed via pip / from source (upstream `master @ 9bd39311`)
- Milvus version: `master-20260422-e91e8825f7` (commit `e91e8825f7`, includes PR #49182 for struct-array element-level search)
- pymilvus version: `2.7.0rc197` on PyPI, also confirmed on upstream `master @ 9bd39311` (same code path)

### Anything else?

**Suggested fix (1 line)**

Put the PK cursor on the **left** of AND. `AND` is commutative for regular predicates, so this is a no-op for existing users, and it preserves the right-most position for operators like `element_filter()` that require it.

```python
# pymilvus/orm/iterator.py  QueryIterator.__setup_next_expr
-    return "(" + current_expr + ")" + " and " + filtered_pk_str
+    # Put the PK cursor on the LEFT of AND so that operators with strict
+    # position constraints (e.g. `element_filter()` on struct array) keep
+    # their position across paginated calls.
+    return filtered_pk_str + " and (" + current_expr + ")"
```

I've applied this 1-line patch locally and verified it gets the iterator past the parse check; `query_iterator(filter=element_filter(...))` then advances normally through sealed segments.

**Related**

- Milvus server issue uncovered after applying the fix above: https://github.com/milvus-io/milvus/issues/49260 — once the parse rejection is removed, a distinct server-side `ElementFilterBitsNode` assertion fires when the query reaches growing segments. That is a separate server bug; this pymilvus issue tracks only the client-side order fix.
- `SearchIteratorV2` is unaffected (distance cursor, no filter rewrite).

I'm happy to send a PR with the 1-line fix.

zhuwenxing:
## Summary

Fixes #3435.

`QueryIterator.__setup_next_expr` previously concatenated the
paginated expression as `(<user_filter>) and <pk_cursor>`, which
breaks on page 2+ whenever the user filter is `element_filter()`
on a struct array field — the Milvus server parser requires
`element_filter()` to be the **right-most** operand of a logical
AND (by design).

Repro error from the issue:

```
MilvusException: (code=1100, message=failed to create query plan: cannot parse expression:
  (element_filter(structA, $[int_val] >= 20000)) and id > 211,
  error: element filter expression can only be the last expression in the logical and expression)
```

This PR swaps the operands so the PK cursor sits on the **left**
of AND. `AND` is commutative for regular predicates, so existing
users are unaffected, and operators with strict position
constraints (like `element_filter()`) keep their right-most slot
across paginated calls.

## Test plan

- [x] Added regression test `test_setup_next_expr_element_filter_stays_right_most`
  asserting the exact emitted expression ordering.
- [x] Strengthened existing `test_setup_next_expr_with_int_cursor`
  to verify the PK cursor precedes the user filter.
- [x] `pytest tests/orm/test_iterator.py` — 136 passed.
- [x] `ruff check` + `black --check` clean on changed files.
- [x] Manual verification on a live Milvus (`master-20260422-e91e8825f7`,
  includes milvus#49182):
  - Before fix: iterator raises on the 2nd `next()` call (parse rejection).
  - After fix: iterator advances through 26 pages / 1292 rows on a
    500-row dataset without the parse rejection. Plain `query()`
    baseline returns 1347 rows; the remaining gap is caused by a
    **separate server-side bug** (`ElementFilterBitsNode` assertion
    in growing segments), tracked at milvus-io/milvus#49260 and
    explicitly called out in the originating issue — out of scope
    for this client-side fix.

### DIFF
--- a/pymilvus/orm/iterator.py
+++ b/pymilvus/orm/iterator.py
@@ -379,7 +379,11 @@ class QueryIterator:
             filtered_pk_str = f"{self._pk_field_name} > {self._next_id}"
         if current_expr is None or len(current_expr) == 0:
             return filtered_pk_str
-        return "(" + current_expr + ")" + " and " + filtered_pk_str
+        # Put the PK cursor on the LEFT of AND so that operators with strict
+        # position constraints (e.g. `element_filter()` on struct array, which
+        # must be the right-most operand of a logical AND) keep their position
+        # across paginated calls.
+        return filtered_pk_str + " and (" + current_expr + ")"
 
     def __update_cursor(self, res: List) -> None:
         if len(res) == 0:
--- a/tests/orm/test_iterator.py
+++ b/tests/orm/test_iterator.py
@@ -578,6 +578,25 @@ class TestQueryIteratorNextExpr:
         expr = qi._QueryIterator__setup_next_expr()
         assert "pk > 42" in expr
         assert "(pk > 0)" in expr
+        # PK cursor must precede the user filter so that right-most-operand
+        # constraints (e.g. `element_filter()`) are preserved across pages.
+        assert expr.index("pk > 42") < expr.index("(pk > 0)")
+
+    def test_setup_next_expr_element_filter_stays_right_most(self):
+        conn = _make_mock_conn(session_ts=100)
+        user_filter = "element_filter(structA, $[int_val] >= 20000)"
+        qi = QueryIterator(
+            connection=conn,
+            collection_name="test",
+            batch_size=10,
+            expr=user_filter,
+            output_fields=["pk"],
+            schema=_SCHEMA_DICT,
+        )
+        qi._next_id = 211
+        expr = qi._QueryIterator__setup_next_expr()
+        # element_filter() must remain the right-most operand of AND.
+        assert expr == f"pk > 211 and ({user_filter})"
 
     def test_setup_next_expr_with_varchar_cursor(self):
         conn = _make_mock_conn(session_ts=100)


zhuwenxing:
/assign @SpadeA-Tang
PTAL

XuanYang-cn:
struct array related, no need to backport to 2.6

XuanYang-cn (approved):
/lgtm

Issue #3437, opened 2026-04-23
[FEATURE]: Support search aggregation on pymilvus


MrPresent-Han:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Milvus now ships a \`common.SearchAggregationSpec\` proto that lets a single vector search return hierarchical bucket aggregations (category → brand → sku, with per-bucket metrics and top_hits). PyMilvus has no user-facing API for this yet — users can only fall back to the legacy single-field \`group_by_field\` / \`group_size\` kwargs, which cannot express:

- multi-field composite grouping keys
- per-bucket metrics (avg/sum/count/min/max)
- bucket-level ordering with tiebreakers
- intra-bucket document ordering (\`top_hits.sort\`)
- N-level nested grouping

Note: this tracks **search** aggregation (vector ANN + bucket grouping), distinct from #3177 which is about **query** aggregation (pure scalar query).

### Describe the solution you'd like

Add request + response classes that mirror the proto 1:1, plus an explicit kwarg on \`MilvusClient.search\` / \`AsyncMilvusClient.search\` / \`Collection.search\`:

\`\`\`python
from pymilvus import MilvusClient, SearchAggregation, TopHits

client.search(
    collection_name=\"products\",
    data=[query_vector],
    anns_field=\"embedding\",
    search_aggregation=SearchAggregation(
        fields=[\"category\"],
        size=5,
        metrics={\"total_revenue\": {\"sum\": \"price\"}},
        order=[{\"total_revenue\": \"desc\"}],
        top_hits=TopHits(size=2, sort=[{\"_score\": \"desc\"}]),
        sub_aggregation=SearchAggregation(
            fields=[\"brand\"],
            size=3,
            metrics={\"avg_rating\": {\"avg\": \"rating\"}},
            order=[{\"avg_rating\": \"desc\"}],
            top_hits=TopHits(size=3, sort=[{\"price\": \"asc\"}]),
        ),
    ),
    search_params={
        \"metric_type\": \"L2\",
        \"params\": {\"nprobe\": 10},
        \"aggregation_safe_factor\": 4,
    },
)
\`\`\`

### Phase 1 scope

- [x] Request classes \`SearchAggregation\` / \`TopHits\` — serialize to \`SearchRequest.search_aggregation\` (proto field 23)
- [x] Composite-field grouping, JSON path expressions
[...]

PR #3438, opened 2026-04-23
feat: support search aggregation API(#3437)

MrPresent-Han:
support search aggregation on pymilvus

issue: #3437


pymilvus/grpc_gen/schema_pb2.pyi
```diff
@@ -438,7 +456,9 @@ class SearchResultData(_message.Message):
     PRIMARY_FIELD_NAME_FIELD_NUMBER: _ClassVar[int]
     HIGHLIGHT_RESULTS_FIELD_NUMBER: _ClassVar[int]
     ELEMENT_INDICES_FIELD_NUMBER: _ClassVar[int]
-    ORDER_BY_FIELD_VALUES_FIELD_NUMBER: _ClassVar[int]
+    GROUP_BY_FIELD_VALUES_FIELD_NUMBER: _ClassVar[int]
```
XuanYang-cn, line 459: **Blocking — silent breaking change.** The proto submodule bump renames `SearchResultData.order_by_field_values` → `group_by_field_values`. Any downstream consumer that handcrafted against this `.pyi` breaks with no warning. Either call this rename out explicitly in the PR description, or split the proto bump into its own PR so it lands with its own changelog entry.

pymilvus/client/prepare.py
```diff
@@ -1572,6 +1586,21 @@ def search_requests_with_expr(
...
+                or aggregation_safe_factor <= 0
+            ):
+                raise ParamError(
+                    message=(
+                        f"{AGGREGATION_SAFE_FACTOR} must be a positive int, "
+                        f"got {aggregation_safe_factor!r}"
+                    )
+                )
+            search_params[AGGREGATION_SAFE_FACTOR] = aggregation_safe_factor
+
```
XuanYang-cn: **Blocking — dual ingestion path, one validated.** `aggregation_safe_factor` passed as a kwarg hits this validator and becomes a top-level `search_params` KV. But users can also put it inside `search_params["aggregation_safe_factor"]` — which is what `examples/search_aggregation.py:240` does — and it then flows through `get_params` (`utils.py:278-293`) into the serialized `params` blob, unvalidated, with a different wire shape. Invalid values (floats, negatives, bools, strings) ship silently via that path. Pick one canonical form and either reject the other or validate both.

pymilvus/client/search_result.py
```diff
@@ -410,6 +416,24 @@ def materialize(self):
...
+        res: schema_pb2.SearchResultData,
+    ) -> List[List[AggregationBucket]]:
+        if len(res.agg_buckets) == 0:
+            return []
+        parsed = parse_agg_buckets(res.agg_buckets)
+        # agg_topks[i] = number of top-level buckets belonging to query i.
+        # If the server didn't set agg_topks (older responses or nq==1), fall
+        # back to putting everything under a single query.
+        topks = list(res.agg_topks) if len(res.agg_topks) > 0 else [len(parsed)]
+        out: List[List[AggregationBucket]] = []
```
XuanYang-cn, line 440: **Blocking — silent bucket drops / shape collapse.**

1. No check that `sum(topks) == len(parsed)`. A buggy/future server that sends `agg_topks=[2,1]` with 4 buckets silently loses the trailing bucket.
2. The `else [len(parsed)]` fallback triggers whenever `agg_topks` is empty — including an `nq > 1` query where middleware drops the field. A 3-vector query then gets a one-element outer list shaped like nq=1.

Add at minimum a `logger.warning` + `assert offset == len(parsed)` after the loop, and an nq-aware fallback (e.g. use `res.num_queries` to emit the right number of empty inner lists).

pymilvus/client/search_aggregation.py
```diff
@@ -0,0 +1,306 @@
...
+"""
+
+from typing import Any, Dict, List, Optional, Union
+
+from pymilvus.exceptions import ParamError
+from pymilvus.grpc_gen import common_pb2, schema_pb2
+
+_VALID_METRIC_OPS = {"avg", "sum", "count", "min", "max"}
+_VALID_DIRECTIONS = {"asc", "desc"}
+_SPECIAL_ORDER_KEYS = {"_count", "_key"}
```
XuanYang-cn, line 15: Nit — `_MAX_NESTING_DEPTH = 4` is a client-side magic number. Does it mirror a server cap, a proto constraint, or is it arbitrary? Either source it from proto/config or add a one-line comment with the rationale so it's not brittle when the server limit changes.

pymilvus/client/search_aggregation.py
```diff
@@ -0,0 +1,306 @@
...
+                    message=f"SearchAggregation.fields must contain non-empty str, got {f!r}"
+                )
+            # JSON path expressions (e.g. "meta['region']") are not yet supported by
+            # the server in search_aggregation. Reject client-side until enabled.
+            if "[" in f or "]" in f:
+                raise ParamError(
+                    message=(
+                        f"SearchAggregation.fields does not yet support JSON path "
+                        f"expressions, got {f!r}"
+                    )
```
XuanYang-cn, line 94: Nit — bracket-only JSON-path rejection. This catches `meta['region']` and `a[0]` but lets `meta.region` through silently. Either widen the regex to cover dotted JSON paths, or narrow the error message to say "bracketed JSON path expressions not yet supported" so the rejection surface is explicit.

pymilvus/client/search_aggregation.py
```diff
@@ -0,0 +1,306 @@
...
+        return out
+
+    def _check_depth(self, current: int) -> None:
+        if current > _MAX_NESTING_DEPTH:
+            raise ParamError(
+                message=f"SearchAggregation nesting depth exceeds limit {_MAX_NESTING_DEPTH}"
+            )
+        if self._sub_aggregation is not None:
+            self._sub_aggregation._check_depth(current + 1)
+
```
XuanYang-cn: Nit — `_oneof_value` is dead code. Neither `AggregationHit` nor `AggregationBucket` calls it; each inlines its own `WhichOneof` loop. Either use it for consistency across the two call sites, or drop it.

pymilvus/client/search_aggregation.py
```diff
@@ -0,0 +1,306 @@
...
+        if self._sub_aggregation is not None:
+            spec.sub_aggregation.CopyFrom(self._sub_aggregation.to_proto())
+        return spec
+
+
+def _oneof_value(msg: Any) -> Any:
+    """Resolve the populated oneof branch on a proto message; returns None if none set."""
+    which = msg.WhichOneof("value") if msg.DESCRIPTOR.oneofs_by_name.get("value") else None
+    return getattr(msg, which) if which else None
+
```
XuanYang-cn, line 189: Nit — inconsistent missing-name fallback. `AggregationHit` falls back to `str(field_id)` when the proto has no `field_name`, but `AggregationBucket.key` falls back to empty string (verified in `tests/test_search_aggregation.py:394` vs `:450-454`). Pick one behavior — the asymmetry will trip users who iterate both sides of the response.

pymilvus/milvus_client/milvus_client.py
```diff
@@ -431,6 +433,9 @@ def search(
                 Defaults to None.
             ids (Optional[Union[List[int], List[str], str, int]]): The ids to use for the search.
                 Defaults to None.
+            search_aggregation (SearchAggregation, optional): Hierarchical bucket aggregation spec.
+                Mutually exclusive with group_by_field. When set, `limit` is ignored and the root
```
XuanYang-cn, line 437: Nit — docstring says "`limit` is ignored" when `search_aggregation` is set, but nothing in the client enforces or warns. A user passing both will silently have `limit` dropped at the server. Consider at least `logger.warning` when a non-default `limit` collides with `search_aggregation`, so the silent drop isn't a debugging black hole.

XuanYang-cn (changes requested):
Phase-1 search-aggregation API is well-scoped, mirrors issue #3437, and has strong unit coverage for request-side validation and response-side parsing. Public API changes are default-safe. Three blockers before a clean LGTM — see inline comments.

**Blocking**
1. `schema_pb2.pyi` — silent rename of `SearchResultData.order_by_field_values` → `group_by_field_values` from the proto submodule bump. Undocumented breaking surface change. Call out in PR body or split proto bump into its own PR so downstream consumers aren't surprised.
2. `prepare.py` — `aggregation_safe_factor` has two ingestion paths (kwargs vs inside the `search_params` dict) that send different wire encodings and only the kwargs path is validated. The example itself uses the un-validated `search_params` dict path (`examples/search_aggregation.py:240`). Invalid values (floats/bools/negatives) ship silently. Unify on one canonical form or validate both.
3. `search_result.py:_split_agg_buckets` — never checks `sum(topks) == len(parsed)` and collapses multi-query responses (nq > 1) into an nq=1 shape when `agg_topks` is absent. Add at least a warning + nq-aware fallback so bucket drops or shape collapses don't go silent.

**Nits** (inline): JSON-path rejection catches `[...]` only (not dotted `meta.region`); `_MAX_NESTING_DEPTH = 4` magic; unused `_oneof_value` helper; bucket-key vs hit-field missing-name fallback inconsistency; docstring claims `limit` is ignored with no in-client warn; no end-to-end test through `MilvusClient.search` / `AsyncMilvusClient.search` / `Collection.search` — only `Prepare` is exercised.

pymilvus/grpc_gen/schema_pb2.pyi
MrPresent-Han: `order_by_field_values` was never consumed by any Milvus client (zero references in pymilvus/Go SDK) and never shipped in a released server build, so the rename is a no-op in practice. Field 16 is reserved on the proto side. Resolving this thread.

pymilvus/client/prepare.py
MrPresent-Han: Good catch. `aggregation_safe_factor` is not formally modeled in `SearchAggregationSpec` yet, so exposing it from the Python SDK was premature. I removed the SDK-side handling and also removed it from the example to avoid implying support before the proto/schema officially carries it.

pymilvus/client/search_aggregation.py
MrPresent-Han: Removed the client-side hardcoded nesting-depth check. This limit should be enforced by Milvus server side rather than duplicated as an SDK constant, so the Python side now just builds the nested spec and leaves authoritative validation to the server.

pymilvus/client/search_aggregation.py
MrPresent-Han: Narrowed the error message here to make the current rejection surface explicit: the client currently rejects bracketed JSON-path expressions, rather than claiming to reject all JSON-path forms.

pymilvus/client/search_aggregation.py
MrPresent-Han: Switched both `AggregationHit` and `AggregationBucket` over to `_oneof_value`, so the helper is now used consistently instead of leaving duplicated `WhichOneof` handling at each call site.

pymilvus/milvus_client/milvus_client.py
MrPresent-Han: Added a warning on both sync and async search paths when `search_aggregation` is set, so callers are explicitly told that `limit` is ignored and the top-level bucket count is controlled by `SearchAggregation.size`.

MrPresent-Han:
Added lightweight coverage through the public client entry points as well: `MilvusClient.search`, `AsyncMilvusClient.search`, and `Collection.search` now each have a focused test asserting that `search_aggregation` is forwarded down correctly.

pymilvus/client/search_aggregation.py
MrPresent-Han: changed

pymilvus/client/search_result.py
MrPresent-Han: Fixed this path with warning + fallback handling. When `agg_topks` is missing for `nq > 1`, the SDK now logs a warning and returns the correct outer shape (`nq` empty groups) instead of collapsing to an `nq=1` layout. When `agg_topks` does not match the parsed bucket count, it now logs a warning and preserves trailing buckets instead of silently dropping them.

XuanYang-cn:
This is master/3.0 feature, no need to backport
/lgtm
/approve

XuanYang-cn (approved):

PR #3445, opened 2026-04-28
Feat: Add cluster session routing for DQL

xiaocai2333:


XuanYang-cn:
Given the intended lifecycle, I think `session()` should be modeled as a dependent session/proxy rather than a shallow-cloned `MilvusClient`.

Desired behavior seems to be:

- `session.close()` only closes the session and must not close the parent `MilvusClient`
- `MilvusClient.close()` closes the parent connection, so existing sessions become unusable
- if the parent client's gRPC handler reconnects/rebuilds, sessions should follow the parent

That points to a session object that stores only `parent_client`, `cluster_id`, and its own `_closed` flag. DQL methods inject `cluster_id` and delegate to the parent. Other APIs can be forwarded to the parent if we want to preserve the full client API surface, but `close()` should be implemented on the session itself so it does not fall through to `parent.close()`.

I’d avoid copying `_handler`, `_manager`, `_config`, `_using`, or `_owns_connection` into the session. If calls resolve through the parent, the session naturally follows parent recovery/reconnect behavior, and parent close naturally makes the session unusable. This keeps `ConnectionManager` as the owner of the actual transport lifecycle while giving the session only local close state.

For concurrency, the session only needs to guard its local `_closed` flag. `close()` should prevent new calls but not cancel or wait for in-flight RPCs, matching normal client-style close semantics. Also, each DQL method should copy `kwargs` before injecting `cluster_id` to avoid mutating caller-owned/shared kwargs.

Conceptually:

```python
class ClusterSession:
    def __init__(self, parent, cluster_id):
        self._parent = parent
        self._cluster_id = cluster_id
        self._closed = False
        self._lock = threading.RLock()

    def close(self):
        with self._lock:
            self._closed = True

    def _check_open(self):
        with self._lock:
            if self._closed:
                raise MilvusException(message="session is closed")

    def search(self, *args, **kwargs):
        self._check_open()
        kwargs = dict(kwargs)
        kwargs.setdefault("cluster_id", self._cluster_id)
        return self._parent.search(*args, **kwargs)

    def __getattr__(self, name):
        attr = getattr(self._parent, name)
        if not callable(attr):
            self._check_open()
            return attr

        def wrapped(*args, **kwargs):
            self._check_open()
            return attr(*args, **kwargs)

        return wrapped
```

PR #3466, opened 2026-05-07
enhance: return external spec in refresh job info

weiliu1031:
Related to [#45881](https://github.com/milvus-io/milvus/issues/45881)

## What changed

- Bump the milvus-proto submodule and regenerate Python gRPC proto files.
- Expose `external_spec` on `RefreshExternalCollectionJobInfo`.
- Pass through `external_spec` from refresh external collection progress/list responses.
- Add sync/async handler and client tests for the new field.

## Why

Milvus now returns a redacted `external_spec` in refresh external collection job info. PyMilvus needs to expose that field so users can inspect the effective external table spec associated with refresh jobs.


XuanYang-cn:
this is 3.0 only, no need to backport to 2.6

XuanYang-cn (approved):
/lgtm

Issue #3154, opened 2025-12-24
[FEATURE]: Support nullable vectors in Python SDK


marcelo-cjl:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Currently, pymilvus does not support nullable vector fields. When inserting data with `None` values for vector fields, it raises an error.

### Describe the solution you'd like

Support nullable vectors for all 6 vector types:
  - FloatVector
  - BinaryVector
  - Float16Vector
  - BFloat16Vector
  - SparseFloatVector
  - Int8Vector

### Describe alternatives you've considered

Handle `None` values in vector field packing and use `valid_data` for logical-to-physical index mapping in query/search results

XuanYang-cn:
should be done

PR #3479, opened 2026-05-12
fix: preserve nullable vectors in CSV bulk writer

marcelo-cjl:
## Summary
- Preserve `None` values when serializing nullable vector fields in BulkWriter CSV so the default `nullkey=""` emits empty cells instead of the literal `null`.
- Handle nullable float16/bfloat16 vector rows without calling `np.frombuffer()` on `None`.
- Add CSV coverage for default and custom nullkey behavior across nullable vector types.

related: #3154
issue: milvus-io/milvus#49678

## Test plan
- [x] `PYTHONPATH=. pytest tests/test_bulk_writer_buffer.py -k nullable_vector -q`
- [x] `PYTHONPATH=. pytest tests/test_bulk_writer_buffer.py -q`
- [x] E2E against `milvusdb/milvus:master-20260509-0a139eba`: legacy `null` CSV reproduces the dim-0 import failure; fixed `RemoteBulkWriter` CSV imports 2 rows and queries the nullable vector row as `None`.

### DIFF
--- a/pymilvus/bulk_writer/buffer.py
+++ b/pymilvus/bulk_writer/buffer.py
@@ -373,19 +373,22 @@ class Buffer:
             }:
                 arr = []
                 for val in v:
-                    arr.append(json.dumps(val))
+                    arr.append(None if val is None else json.dumps(val))
                 data[k] = pd.Series(arr, dtype=np.dtype("str"))
             elif field_schema.dtype in {DataType.FLOAT16_VECTOR, DataType.BFLOAT16_VECTOR}:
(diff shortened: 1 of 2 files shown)


XuanYang-cn:
@marcelo-cjl is this 3.x only feature? I see nullable vectors.

marcelo-cjl:
> @marcelo-cjl is this 3.x only feature? I see nullable vectors.

It will be backported to the cloud with version 2.6.17.

XuanYang-cn (approved):
/lgtm

PR #3497, opened 2026-05-17
[codex] fix FieldSchema bool flag validation

pragnyanramtha:
## What

Tightens `FieldSchema` validation so `is_dynamic` and `nullable` must be booleans, matching the existing strict handling for neighboring schema flags such as `is_primary`, `auto_id`, `is_partition_key`, and `is_clustering_key`.

## Why

Before this change, truthy non-boolean values like `nullable="true"` or `is_dynamic="true"` were accepted and serialized into schema dictionaries. That can leak invalid schema metadata to later client/server paths instead of failing early with a clear parameter error.

## Validation

- `.venv/bin/python -m pytest tests/orm/test_schema.py::TestFieldSchemaValidation -q`
- `.venv/bin/python -m pytest tests/orm/test_schema.py -q`
- `.venv/bin/python -m pytest tests/prepare/test_collection.py::TestGetFieldSchema -q`
- `git diff --check`


pragnyanramtha:
Hi @tedxu, could you please review this when you have a chance? DCO is green and Codecov reports all modified coverable lines covered.

Issue #3486, opened 2026-05-13
[FEATURE]: Adapt uv


XuanYang-cn:
### Is there an existing issue for this?

- [x] I have searched the existing issues

PR #3512, opened 2026-05-22
chore: migrate project workflow to uv

zhuwenxing:
## Summary
- migrate local development, test, and package build commands to uv
- add uv.lock for reproducible dependency resolution
- move development-only dependencies to a uv dependency group instead of publishing them as a package extra
- clean package metadata and sdist contents, and add distribution checks to publish workflows
- keep Python 3.14 CI working by resolving grpcio to a version with Python 3.14 wheels

## Validation
- make lint
- make coverage: 4004 passed, 4 skipped, 1 xpassed
- uv build --sdist --wheel
- uvx twine check dist artifacts
- uvx check-wheel-contents wheel artifact


XuanYang-cn:
See also: #3486

.github/workflows/publish_dev_package.yml
```diff
@@ -24,22 +24,32 @@ jobs:
...
+        uv
         build
         --sdist
         --wheel
-        --outdir dist/
+        --out-dir dist/
         .
+    - name: Check distribution metadata
+      run: >-
+        uvx
```
XuanYang-cn: This release gate is still resolving `twine` at publish time via `uvx`, outside `uv.lock`. Since this PR is adding locked uv-based workflows, the TestPyPI publish job can still start failing on unrelated upstream `twine` changes and the PR checks will not catch it. Please add the release tools to a locked dependency group, for example `release = ["twine", "check-wheel-contents"]`, and run them with `uv run --frozen --group release ...` instead of bare `uvx`.

.github/workflows/publish_on_release.yml
```diff
@@ -23,20 +23,31 @@ jobs:
...
+        uv
         build
         --sdist
         --wheel
-        --outdir dist/
+        --out-dir dist/
         .
+    - name: Check distribution metadata
+      run: >-
+        uvx
```
XuanYang-cn: Same issue in the PyPI release workflow: this `uvx` invocation resolves the checker at release time outside the lockfile. Please use the same locked release group and `uv run --frozen --group release ...` here too, otherwise package publishing remains dependent on floating tooling.

XuanYang-cn (changes requested):
Requested changes focus on the publish workflows. The PR moves the project toward locked/reproducible uv workflows, but the release validation tools are still resolved at publish time outside the lockfile.

Makefile
```diff
@@ -1,39 +1,45 @@
...
+	$(UV) sync --group dev
+
 unittest:
-	PYTHONPATH=`pwd` python3 -m pytest tests --ignore=tests/benchmark --cov=pymilvus -v
+	PYTHONPATH=$(CURDIR) $(UV_RUN_DEV) python -m pytest tests --ignore=tests/benchmark --cov=pymilvus -v
 
 lint:
-	PYTHONPATH=`pwd` python3 -m black pymilvus tests --check --diff
-	PYTHONPATH=`pwd` python3 -m ruff check pymilvus tests
+	PYTHONPATH=$(CURDIR) $(UV_RUN_DEV) python -m black pymilvus tests --check --diff
```
XuanYang-cn: Simplify these calls to the console scripts exposed by the dev tools: `$(UV_RUN_DEV) black ...` and `$(UV_RUN_DEV) ruff check ...`. Keeping `python -m pytest` for tests is still reasonable.

Makefile
```diff
@@ -1,39 +1,45 @@
...
+	PYTHONPATH=$(CURDIR) $(UV_RUN_DEV) python -m ruff check pymilvus tests
 
 format:
-	pip install -e ".[dev]"
-	PYTHONPATH=`pwd` python3 -m black pymilvus tests
-	PYTHONPATH=`pwd` python3 -m ruff check pymilvus tests --fix
+	$(UV) sync --group dev
```
XuanYang-cn: `uv run --group dev` already syncs the project environment before running the command. This explicit `uv sync --group dev` makes `make format` do redundant environment work; rely on `$(UV_RUN_DEV)` here.

Makefile
```diff
@@ -1,39 +1,45 @@
...
-	python3 -m build --sdist --wheel --outdir dist/ .
+	$(UV) build --sdist --wheel --out-dir dist/ .
 
 get_proto:
 	git submodule update --init
 
 gen_proto:
-	pip install -e ".[dev]"
-	cd pymilvus/grpc_gen && ./python_gen.sh
+	$(UV) sync --group dev
```
XuanYang-cn: Same redundant-sync point here. The `PYTHON_CMD="$(UV) run --group dev python"` command will sync before invoking codegen. If the goal is to avoid Python 3.14 for proto generation, make that explicit with `--python 3.13`.

.github/workflows/pull_request.yml
```diff
@@ -29,10 +29,13 @@ jobs:
         run: |
           git fetch --prune --unshallow --tags
 
+      - name: Install uv
+        run: |
+          python -m pip install --upgrade uv
```
XuanYang-cn: Since GitHub Actions is installing uv in several jobs, use `astral-sh/setup-uv` with caching instead of `python -m pip install --upgrade uv`. It standardizes uv setup and avoids re-downloading the uv package in every job.

.github/workflows/code_checker.yml
```diff
@@ -19,12 +19,15 @@ jobs:
         uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Install uv
+        run: |
+          python -m pip install --upgrade uv
       - name: Check pyproject.toml install
         run: |
-          pip install -e .
+          uv sync --frozen --python python
```
XuanYang-cn, line 30: This first runtime-only sync and the following dev sync both install the project. Keep this only if the goal is to explicitly validate a non-dev install; otherwise the `uv sync --group dev --frozen --python python` step is enough for the lint job.

XuanYang-cn (commented):
Optional follow-up suggestions from the uv migration review. These are not additional blockers beyond the release-tooling lockfile issue already requested.

XuanYang-cn (approved):
/lgtm

PR #3528, opened 2026-05-27
chore: use hatchling version source

zhuwenxing:
## Summary
- switch the build backend from setuptools to hatchling
- use Hatch's code version source to keep the existing rc-distance version scheme without a root-level _version_helper.py or setup.py
- use the hatchling CLI for version detection in Makefile and the Milvus update workflow
- remove the obsolete manifest include for _version_helper.py
- add focused tests for the rc version scheme

## Version behavior
- non-tag builds continue to resolve to rc-distance versions, e.g. 3.1.0rc16
- exact tag builds continue to resolve to the tag version

## Validation
- make version
- uv run --python 3.8 --no-project --with 'hatchling>=1.27,<1.28' --with 'setuptools-scm[toml]>=8' python -m hatchling version
- uv build --python 3.8 --no-sources --sdist --wheel
- uv build --python 3.13 --no-sources --sdist --wheel
- twine check
- check-wheel-contents
- black --check pymilvus/_version.py tests/test_version.py
- ruff check pymilvus/_version.py tests/test_version.py
- PYTHONPATH=/Users/jeremy.zhu/workspace/pymilvus.refactor_to_uv uv run --group dev pytest tests/test_version.py -q


zhuwenxing:
Updated this PR to use hatchling's code version source while preserving the existing rc-distance behavior.

- `make version` now returns `3.1.0rc16` on this branch.
- The package no longer switches to pure `.devN`, so the README and `get_commit()` concerns from the previous revision should no longer apply.
- No `setup.py` or root-level `_version_helper.py`; the rc scheme lives in `pymilvus/_version.py` and is wired from `[tool.hatch.version]`.

/assign @XuanYang-cn

.github/workflows/update-milvus-pymilvus-dev.yml
```diff
@@ -92,7 +92,7 @@ jobs:
     - name: Determine latest dev version
       id: version
       run: |
-        VERSION=$(uv run --no-project --with setuptools --with setuptools-scm python -c "import _version_helper; print(_version_helper.version)")
+        VERSION=$(uv run --no-project --with "hatchling>=1.27,<1.28" --with "setuptools-scm[toml]>=8" python -m hatchling version)
```
XuanYang-cn, line 95: The scheduled workflow still checks out `2.6` and `3.0` in the matrix. Those branches still use `setuptools.build_meta`/`_version_helper.py` and have no `[tool.hatch.version]`, so after this change the scheduled jobs fail as soon as they run this line. I verified the new command against current `upstream/2.6` and `upstream/3.0`; both exit with `ValueError: Missing `tool.hatch.version` configuration`. Please keep this step compatible with the pre-hatch release branches, or land/backport the hatch version config before switching the shared scheduled workflow.

XuanYang-cn (changes requested):
Found one blocking issue in the update workflow. I verified the PR head locally with `python -m hatchling version`, `pytest tests/test_version.py -q`, and `uv build --sdist --wheel`; those pass for the PR branch, but the scheduled branch matrix still breaks on release branches.

XuanYang-cn (approved):

Issue #3128, opened 2025-12-04
[FEATURE]: Support AlterCollectionSchema api


MrPresent-Han:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Support AlterCollectionSchema api

PR #3568, opened 2026-06-04
feat: add interface for alter_collection_schema (#3128)

MrPresent-Han:
Add AlterCollectionSchema API supporting both Add and Drop operations:
- Add operation: provide field_schema, func, index_param
- Drop operation: provide drop_field_name or drop_field_id

related: #3128

### DIFF
--- a/pymilvus/client/grpc_handler.py
+++ b/pymilvus/client/grpc_handler.py
@@ -568,6 +568,48 @@ class GrpcHandler:
         check_status(status)
         self._invalidate_schema(collection_name, db_name=(context.get_db_name() if context else ""))
 
+    def _alter_collection_schema_drop(
+        self,
+        collection_name: str,
+        field_name: str = "",
+        field_id: int = 0,
+        function_name: str = "",
+        db_name: str = "",
+        timeout: Optional[float] = None,
+        context: Optional[CallContext] = None,
+        **kwargs,
+    ):
+        check_pass_param(collection_name=collection_name, timeout=timeout)
+        request = Prepare.alter_collection_schema_request(
+            collection_name=collection_name,
+            drop_field_name=field_name,
+            drop_field_id=field_id,
+        )
+        response = self._stub.AlterCollectionSchema(
+            request, timeout=timeout, metadata=_api_level_md(context)
+        )
+        check_status(response.alter_status)
+
+    @retry_on_rpc_failure()
+    def drop_collection_field(
+        self,
+        collection_name: str,
+        field_name: str = "",
+        field_id: int = 0,
+        timeout: Optional[float] = None,
+        context: Optional[CallContext] = None,
+        **kwargs,
+    ):
+        self._alter_collection_schema_drop(
+            collection_name,
+            field_name=field_name,
+            field_id=field_id,
+            timeout=timeout,
+            context=context,
+            **kwargs,
+        )
+        self._invalidate_schema(collection_name, db_name=(context.get_db_name() if context else ""))
+
     @retry_on_rpc_failure()
     def drop_collection_function(
         self,
@@ -621,6 +663,32 @@ class GrpcHandler:
             request, timeout=timeout, metadata=_api_level_md(context)
         )
         check_status(status)
+
+    @retry_on_rpc_failure()
+    def alter_collection_schema(
+        self,
+        collection_name: str,
+        field_schema: Optional[FieldSchema] = None,
+        func: Optional[Function] = None,
+        timeout: Optional[float] = None,
+        context: Optional[CallContext] = None,
+        drop_field_name: Optional[str] = None,
+        drop_field_id: Optional[int] = None,
+        **kwargs,
+    ):
+        check_pass_param(collection_name=collection_name, timeout=timeout)
+        request = Prepare.alter_collection_schema_request(
+            collection_name=collection_name,
+            field_schema=field_schema,
+            func=func,
+            drop_field_name=drop_field_name,
+            drop_field_id=drop_field_id,
+        )
+        response = self._stub.AlterCollectionSchema(
+            request, timeout=timeout, metadata=_api_level_md(context)
+        )
+        check_status(response.alter_status)
+        self._invalidate_schema(collection_name, db_name=(context.get_db_name() if context else ""))
 
     @retry_on_rpc_failure()
     def alter_collection_properties(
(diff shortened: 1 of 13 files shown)


pymilvus/milvus_client/milvus_client.py
```diff
@@ -1362,6 +1363,143 @@ def drop_collection_function(
             **kwargs,
         )
 
+    def drop_collection_field(
```
XuanYang-cn, line 1366: must-change: `drop_collection_field` exposes a `DropRequest` path that current Milvus rejects. Proxy and RootCoord still require `add_request` for `AlterCollectionSchema`, so this public API always fails against the supported server path. Remove or gate the drop API until matching server support lands, or land the server support before merging this client change.

pymilvus/client/grpc_handler.py
```diff
@@ -622,6 +666,31 @@ def alter_collection_function(
...
+            collection_name=collection_name,
+            field_schema=field_schema,
+            func=func,
+            drop_field_name=drop_field_name,
+            drop_field_id=drop_field_id,
+        )
+        response = self._stub.AlterCollectionSchema(
+            request, timeout=timeout, metadata=_api_level_md(context)
+        )
+        check_status(response.alter_status)
```
XuanYang-cn, line 690: must-change: `alter_collection_schema` leaves cached schemas stale after a successful add. `_get_schema` can keep returning the pre-alter field list and schema version to schema-dependent operations after `add_function_field` succeeds. Invalidate `GlobalCache.schema` here and in `AsyncGrpcHandler.alter_collection_schema` after `alter_status` passes, matching the other schema-mutating paths.

XuanYang-cn (changes requested):
Reviewed at `ecfaa847ede1a233c5983dd9c5f14ded6d677a0f`. Requesting changes for two correctness issues: the new drop API is exposed before the current Milvus AlterCollectionSchema path accepts `DropRequest`, and successful add operations leave PyMilvus schema cache entries stale.

pymilvus/milvus_client/milvus_client.py
MrPresent-Han: Milvus master already supports dropping fields via `AlterCollectionSchema.DropRequest`, so this comment is stale. The `drop_collection_field` client path targets that supported server-side behavior.

pymilvus/client/grpc_handler.py
MrPresent-Han: good catch, really

pymilvus/milvus_client/milvus_client.py
```diff
@@ -1362,6 +1363,143 @@ def drop_collection_function(
...
+        """Drop a field from the collection."""
+        conn = self._get_connection()
+        conn.drop_collection_field(
+            collection_name,
+            field_name=field_name,
+            field_id=field_id,
+            timeout=timeout,
+            context=self._generate_call_context(**kwargs),
+            **kwargs,
+        )
```
XuanYang-cn, line 1381: Is this needed?

pymilvus/milvus_client/async_milvus_client.py
```diff
@@ -1089,6 +1089,73 @@ async def drop_collection_function(
             **kwargs,
         )
 
+    async def alter_collection_schema(
```
XuanYang-cn: ditto with milvus client part

XuanYang-cn (changes requested):

pymilvus/milvus_client/milvus_client.py
```diff
@@ -1362,6 +1363,143 @@ def drop_collection_function(
...
+    def alter_collection_schema(
+        self,
+        collection_name: str,
+        field_schema: Optional[FieldSchema] = None,
+        func: Optional[Function] = None,
+        timeout: Optional[float] = None,
+        drop_field_name: Optional[str] = None,
+        drop_field_id: Optional[int] = None,
+        **kwargs,
+    ):
```
XuanYang-cn, line 1392: `MilvusClient.alter_collection_schema` should expose an operation-oriented
  Python API instead of protobuf-shaped oneof parameters.

  The current signature:

      alter_collection_schema(
          collection_name,
          field_schema=None,
          func=None,
          drop_field_name=None,
          drop_field_id=None,
      )

  is hard to use because it makes the caller reason about transport-level combinations.
  `field_schema + func` means “add a function-backed field,” but that intent is not
  visible at the call site. `drop_field_name` and `drop_field_id` expose the server
  request shape rather than the Python client model. The method also has to reject many
  invalid combinations after the user has already constructed a call.

  The public MilvusClient API should make the operation explicit and group related inputs
  together:

      client.alter_collection_schema(
          collection_name,
          add_fields=[
              FieldSchema("sparse", DataType.SPARSE_FLOAT_VECTOR),
          ],
          add_functions=[
              Function("bm25_fn", FunctionType.BM25, ["text"], ["sparse"]),
          ],
      )

      client.alter_collection_schema(
          collection_name,
          drop_fields=["old_field"],
      )

  This is easier to read because the call now says what the user is doing: adding fields/
  functions or dropping fields. It also makes validation cleaner: exactly one operation
  family should be set per call. For example, `add_fields` / `add_functions` can be
  accepted together, but `add_*` and `drop_*` cannot be mixed. If the server currently
  supports only a narrower add shape, such as one BM25 function with one output field,
  enforce that as validation inside the request builder while keeping the user-facing
  names operation-oriented.

  Suggested public signature:

      def alter_collection_schema(
          self,
          collection_name: str,
          *,
          add_fields: Optional[list[FieldSchema]] = None,
          add_functions: Optional[list[Function]] = None,
          drop_fields: Optional[list[str]] = None,
          drop_field_ids: Optional[list[int]] = None,
          drop_functions: Optional[list[str]] = None,
          **kwargs,
      ):
          ...

  Then map this API to the current protobuf internally in
  `Prepare.alter_collection_schema_request`. The handler layer can still work with the
  protobuf-shaped add/drop request, but `MilvusClient` should not expose that shape
  directly.

  Keep small convenience helpers on top of the generic method:

      client.add_function_field(collection_name, field_schema, function)
      client.drop_field(collection_name, field_name)

  Those helpers are the easiest path for common use cases, while `alter_collection_schema`
  remains the advanced grouped-operation API. This keeps the Python client readable now
  and leaves room for future multi-field or function-drop support without another public
  signature rewrite.

pymilvus/milvus_client/milvus_client.py
MrPresent-Han: truly need

pymilvus/milvus_client/milvus_client.py
MrPresent-Han: has made _alter_collection_schema a private method and not expose to external caller
add_function_field and drop_collection_field call this private method in milvus-client

pymilvus/milvus_client/async_milvus_client.py
MrPresent-Han: changed

MrPresent-Han: changed

XuanYang-cn (approved):
/lgtm

Issue #3581, opened 2026-06-08
[Bug]: drop_collection_function routes through the wrong RPC with text embedding functions


XuanYang-cn:
## Is there an existing issue for this?

- [x] I have searched the existing issues

## Describe the bug

PyMilvus `3.1.0rc30` fails `drop_collection_function` flows for text embedding functions.

This surfaced in Milvus `ci-v2/e2e-default` after bumping from `pymilvus==3.1.0rc8` to `pymilvus==3.1.0rc30`.

CI link:
https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pipeline-gcp/34619/pipeline-overview/log?nodeId=156

Concrete failures:

```text
FAILED testcases/test_text_embedding_function_e2e.py::TestTextEmbeddingFunctionCURD::test_drop_collection_function_then_add_again
pymilvus.exceptions.MilvusException: <MilvusException: (code=1100, message=cannot drop function tei: it would leave no vector field in the collection: invalid parameter)>

FAILED testcases/test_text_embedding_function_e2e.py::TestTextEmbeddingFunctionCURD::test_drop_collection_function_verify_crud
pymilvus.exceptions.MilvusException: <MilvusException: (code=1100, message=cannot drop function tei: it would leave no vector field in the collection: invalid parameter)>

FAILED testcases/test_text_embedding_function_e2e.py::TestTextEmbeddingFunctionCURD::test_drop_collection_function_one_of_multiple
AssertionError: Response of API insert expect True, but got False
```

## Expected Behavior

`drop_collection_function` should remove the named function through the correct RPC path and allow the text embedding function workflows to continue.

Dropping one text embedding function should not be rejected as if the operation would leave the collection without a vector field when the workflow is valid.

## Steps/Code To Reproduce behavior

Use a collection with text embedding functions and call `drop_collection_function` in the text embedding function CRUD flows.

The failure is currently reproducible in Milvus `ci-v2/e2e-default` with PyMilvus `3.1.0rc30`.

## Environment details

- PyMilvus version: `3.1.0rc30`
- Previous working Milvus test requirement: `3.1.0rc8`
- Failing CI: `ci-v2/e2e-default`
- Install method: Milvus CI Python client test requirements

## Anything else?

[...]

PR #3593, opened 2026-06-09
feat: Add drop_function_field API

sijie-ni-0214:
issue: #3581

## What
- Add `drop_function_field` to sync and async MilvusClient APIs.
- Forward `drop_function_output_fields` through AlterCollectionSchema drop requests.
- Regenerate gRPC stubs for the new drop flag.
- Add handler, prepare, and client delegation tests.


pymilvus/client/prepare.py
```diff
@@ -428,6 +429,7 @@ def alter_collection_schema_request(
                 drop_request.field_id = drop_field_id
             else:
                 drop_request.function_name = drop_function_name
+                drop_request.drop_function_output_fields = drop_function_output_fields
```
XuanYang-cn, line 432: must-change: `drop_function_output_fields` depends on a Milvus server proto/handler update that is not on current `milvus-io/milvus` master. That branch still pins `milvus-proto/go-api/v3` at `489331a5a41f` and has no `DropFunctionOutputFields` handling, so servers built from current master ignore this field and `drop_collection_function()` keeps the old cascading behavior that triggers #3581. Merge this PR only after the Milvus server update using `8daa6dc555e6a42c3c5b83cf6386e94bd783481c` is merged, or document and enforce that release dependency in this PR.

XuanYang-cn (dismissed):
I found one required release-contract issue.\n\nProto validation: `pymilvus/grpc_gen/milvus-proto` points to `8daa6dc555e6a42c3c5b83cf6386e94bd783481c`, which exists on `milvus-io/milvus-proto` master. The range from the previous `551b5154e14318a76761fdeaf2ce1c2ccf781892` also includes `489331a` and `e90ee63`, so the generated snapshot/export and delete-namespace updates are expected. `make check_proto_product` passed locally under Python 3.12, and the focused request/handler/client unit slice passed (`23 passed`).\n\nThe blocker is that current `milvus-io/milvus` master still pins `milvus-proto/go-api/v3` at `489331a5a41f` and has no `DropFunctionOutputFields` handling, so this SDK change needs the server update before it can fix #3581.

pymilvus/client/grpc_handler.py
```diff
@@ -630,6 +632,25 @@ def drop_collection_function(
         )
         self._invalidate_schema(collection_name, db_name=(context.get_db_name() if context else ""))
 
+    @retry_on_rpc_failure()
+    def drop_function_field(
```
XuanYang-cn: must-change: `GrpcHandler.drop_function_field()` adds a handler-level API path that has no production caller. The public sync/async `MilvusClient.drop_function_field()` methods route through `_alter_collection_schema(..., drop_function_output_fields=True)` and then `alter_collection_schema()`, so this method is only exercised by its direct unit test. Remove this unused handler method and its direct handler test, and keep the coverage on the public client-to-`alter_collection_schema` path that actually ships the request.

XuanYang-cn (changes requested):
One blocking cleanup issue: the PR adds a handler-level `drop_function_field()` path that the public clients do not call. Please remove the unused handler methods/tests or route the public clients through them so there is one production path.

pymilvus/client/grpc_handler.py
sijie-ni-0214: Removed the unused handler-level path from both sync and async handlers, along with the direct handler tests that only exercised that path. The public `MilvusClient.drop_function_field()` / `AsyncMilvusClient.drop_function_field()` coverage now remains on the production `alter_collection_schema(..., drop_function_output_fields=True)` route.

Pushed in `c20dceb7`. Local checks:
- focused pytest slice: 79 passed, 1 warning
- `ruff check` on touched files: passed
- `git diff --check`: passed

XuanYang-cn (approved):
/lgtm

PR #3602, opened 2026-06-12
feat: support user descriptions in auth APIs

shaoting-huang:
- bump milvus-proto to 43795e8f8f6eac62d673592f9ad28ea6ea3acdad and regenerate gRPC stubs
- add optional user description passthrough for create_user/update_password and description-only update_user APIs
- expose user description through UserItem/UserInfo and MilvusClient describe_user paths
- add sync/async unit coverage for description passthrough and update_user guards

### DIFF
--- a/pymilvus/client/grpc_handler.py
+++ b/pymilvus/client/grpc_handler.py
@@ -2665,13 +2665,11 @@ class GrpcHandler:
     def update_user(
         self,
         user: str,
-        description: Optional[str] = None,
+        description: str,
         timeout: Optional[float] = None,
         context: Optional[CallContext] = None,
         **kwargs,
     ):
-        if description is None:
-            raise ParamError(message="description is required")
         req = Prepare.update_password_request(user, "", "", description=description)
         resp = self._stub.UpdateCredential(req, timeout=timeout, metadata=_api_level_md(context))
         check_status(resp)
--- a/pymilvus/client/grpc_handler.py
+++ b/pymilvus/client/grpc_handler.py
@@ -2636,10 +2636,11 @@ class GrpcHandler:
         password: str,
         timeout: Optional[float] = None,
         context: Optional[CallContext] = None,
+        description: Optional[str] = None,
         **kwargs,
     ):
         check_pass_param(user=user, password=password, timeout=timeout)
-        req = Prepare.create_user_request(user, password)
+        req = Prepare.create_user_request(user, password, description=description)
         resp = self._stub.CreateCredential(req, timeout=timeout, metadata=_api_level_md(context))
         check_status(resp)
 
@@ -2651,9 +2652,27 @@ class GrpcHandler:
         new_password: str,
         timeout: Optional[float] = None,
         context: Optional[CallContext] = None,
+        description: Optional[str] = None,
         **kwargs,
     ):
-        req = Prepare.update_password_request(user, old_password, new_password)
+        req = Prepare.update_password_request(
+            user, old_password, new_password, description=description
+        )
+        resp = self._stub.UpdateCredential(req, timeout=timeout, metadata=_api_level_md(context))
+        check_status(resp)
+
+    @retry_on_rpc_failure()
+    def update_user(
+        self,
+        user: str,
+        description: Optional[str] = None,
+        timeout: Optional[float] = None,
+        context: Optional[CallContext] = None,
+        **kwargs,
+    ):
+        if description is None:
+            raise ParamError(message="description is required")
+        req = Prepare.update_password_request(user, "", "", description=description)
         resp = self._stub.UpdateCredential(req, timeout=timeout, metadata=_api_level_md(context))
         check_status(resp)
 
--- a/pymilvus/milvus_client/async_milvus_client.py
+++ b/pymilvus/milvus_client/async_milvus_client.py
@@ -1423,15 +1423,12 @@ class AsyncMilvusClient(BaseMilvusClient):
         **kwargs,
     ):
         conn = await self._get_connection()
-        extra_kwargs = {}
-        if description is not None:
-            extra_kwargs["description"] = description
         await conn.create_user(
             user_name,
             password,
             timeout=timeout,
             context=self._generate_call_context(**kwargs),
-            **extra_kwargs,
+            description=description,
             **kwargs,
         )
 
@@ -1451,16 +1448,13 @@ class AsyncMilvusClient(BaseMilvusClient):
         **kwargs,
     ):
         conn = await self._get_connection()
-        extra_kwargs = {}
-        if description is not None:
-            extra_kwargs["description"] = description
         await conn.update_password(
             user_name,
             old_password,
             new_password,
             timeout=timeout,
             context=self._generate_call_context(**kwargs),
-            **extra_kwargs,
+            description=description,
             **kwargs,
         )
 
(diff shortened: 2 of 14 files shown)


shaoting-huang:
Codex workpad summary for ENG-12:

- Implemented user `description` support for create_user/update_password and description-only update_user, including sync and async surfaces.
- Bumped `pymilvus/grpc_gen/milvus-proto` to `43795e8f8f6eac62d673592f9ad28ea6ea3acdad` and regenerated stubs.
- `UserItem` / `UserInfo` now expose `description`; high-level describe_user returns it.
- Review findings addressed: added `utility.update_password(description=...)`, added update_password passthrough coverage, and made omitted `update_user` descriptions raise `ParamError` to avoid accidental clears.

Validation:
- `git submodule update --init && make gen_proto > /tmp/gen.log 2>&1; echo exit=$?` -> exit 0
- `make lint > /tmp/lint.log 2>&1; echo exit=$?` -> exit 0
- `make unittest > /tmp/test.log 2>&1; echo exit=$?` -> exit 0 (`4234 passed, 3 skipped, 1 xpassed`)

Server-side milvus PR is not required for these mock-based SDK tests; merge sequencing remains operator-controlled.

pymilvus/milvus_client/milvus_client.py
```diff
@@ -1629,13 +1629,24 @@ def get_partition_stats(
...
+    def create_user(
+        self,
+        user_name: str,
+        password: str,
+        timeout: Optional[float] = None,
+        description: Optional[str] = None,
+        **kwargs,
+    ):
         conn = self._get_connection()
+        extra_kwargs = {}
```
XuanYang-cn: must-change: `extra_kwargs` adds conditional forwarding for an explicit parameter. This wrapper branch repeats across sync and async clients and makes `description=None` handling harder to reason about. Pass `description=description` directly to `conn.create_user` and `conn.update_password` in both clients.

pymilvus/milvus_client/milvus_client.py
```diff
@@ -1684,7 +1718,10 @@ def describe_user(self, user_name: str, timeout: Optional[float] = None, **kwarg
         )
         if res.groups:
             item = res.groups[0]
-            return {"user_name": user_name, "roles": item.roles}
+            description = getattr(item, "description", "")
```
XuanYang-cn: must-change: `describe_user` hides invalid `UserItem.description` state with `getattr` and a type fallback. This masks the type-wrapper contract at every caller and duplicates normalization across sync and async clients. Make `UserItem`/`UserInfo` guarantee a string description, then return `item.description` directly here.

tests/unit/prepare/test_utility.py
```diff
@@ -336,11 +336,25 @@ def test_transfer_replica(self):
...
+        assert req.description == "reader account"
+
     def test_update_password(self):
         """Test update password request."""
         req = Prepare.update_password_request("user", "old_pass", "new_pass")
         assert req.username == "user"
 
+    def test_update_password_description_only(self):
+        """Test update user description request without password change."""
+        req = Prepare.update_password_request("user", "", "", description="updated account")
```
XuanYang-cn, line 365: must-change: this test only covers a non-empty description value. The feature depends on optional proto field presence, especially `description=""` for clearing a description. Add assertions that `description=""` keeps `HasField("description") == True` and `description=None` leaves the field unset.

XuanYang-cn (changes requested):
Requested changes on the public user-description API plumbing. The issues are small, but they should be fixed before merge because they add avoidable wrapper complexity and miss the optional-field presence contract this feature depends on.

pymilvus/orm/utility.py
XuanYang-cn, line 1: orm will be deprecated soon, no need to change here. Please remove the change.

XuanYang-cn (changes requested):

shaoting-huang:
Addressed the user description review feedback:

- Passed description directly through sync and async MilvusClient.create_user / update_password, leaving proto field presence gating in Prepare.
- Simplified sync and async describe_user to return UserItem.description directly.
- Added tests for description empty string vs description None field presence on create/update credential requests.
- Restored deprecated ORM utility wrappers to their previous signatures and removed the matching wrapper tests.

Local validation:

- make lint -> exit 0
- make unittest -> exit 0 (4240 passed, 3 skipped, 1 xpassed)

pymilvus/milvus_client/milvus_client.py
shaoting-huang: description is now passed directly to conn.create_user/conn.update_password in both sync and async clients

pymilvus/orm/utility.py
shaoting-huang: reverted the change

pymilvus/milvus_client/async_milvus_client.py
```diff
@@ -1447,6 +1454,25 @@ async def update_password(
...
+
+    async def update_user(
+        self,
+        user_name: str,
+        description: Optional[str] = None,
+        timeout: Optional[float] = None,
+        **kwargs,
+    ):
+        if description is None:
+            raise ParamError(message="description is required")
```
XuanYang-cn: must-change: `description` is required by `update_user`, but the new public API declares it as `Optional[str] = None` and then rejects `None` at runtime. Make `description` a required `str` parameter in sync and async `update_user` signatures, while still allowing `description=""` to clear the description.

```python
def abc(
    description: str
): 
```

pymilvus/client/grpc_handler.py
```diff
@@ -2651,9 +2652,27 @@ def update_password(
...
+    def update_user(
+        self,
+        user: str,
+        description: Optional[str] = None,
+        timeout: Optional[float] = None,
+        context: Optional[CallContext] = None,
+        **kwargs,
+    ):
+        if description is None:
+            raise ParamError(message="description is required")
```
XuanYang-cn: ditto.

XuanYang-cn (changes requested):

shaoting-huang:
Addressed the update_user signature review feedback:

- Made sync and async MilvusClient.update_user require description: str.
- Made sync and async grpc handler update_user require description: str.
- Removed the runtime None guard from all four update_user layers.
- Added tests for required public signatures, omitted-description TypeError behavior, and description="" forwarding.

Local validation:

- python -m pytest tests/unit/grpc_handler/test_auth.py tests/unit/async_grpc_handler/test_async_auth.py tests/unit/test_milvus_client.py tests/unit/test_async_milvus_client_ops.py -> 433 passed
- make lint -> exit 0
- make unittest -> exit 0 (4242 passed, 3 skipped, 1 xpassed)

pymilvus/milvus_client/async_milvus_client.py
shaoting-huang: Updated: sync and async MilvusClient.update_user now require description: str with no default, and the old runtime None guard is removed. Added unit coverage for the required public signature, omitted-description TypeError behavior, and description="" forwarding.

pymilvus/client/grpc_handler.py
shaoting-huang: Updated: sync and async grpc handler update_user signatures now also require description: str with no default and no runtime None guard, matching the public client shape.

XuanYang-cn (approved):
/lgtm

PR #3611, opened 2026-06-15
enhance: surface error classification on MilvusException

czs007:
## What

Expose the server's error classification on `MilvusException`:

- `is_input_error` — the request itself was malformed (caller's fault); the server never retries it or fails it over, and forces it non-retriable.
- `retriable` — whether blindly retrying the same request may succeed.

Both are populated by a new `MilvusException.from_status(status)` classmethod, which all status-based raise sites now use: `check_status` plus the `describe_collection` / `has_collection`-style custom raises in both the sync and async gRPC handlers.

## Why

Since Milvus standardized error handling, a server error carries a structured classification beyond its message: `Status.ExtraInfo["is_input_error"]` and `Status.retriable`. The SDK discarded both, so callers (and the test suites) could only match on the error message — which is not a stable contract and breaks on harmless rewording.

## Compatibility

`from_status` defaults a bare error (no `extra_info`, no `retriable`) to a non-input, non-retriable system error, so behavior is unchanged against servers that do not set these fields. The new constructor params are keyword-only with defaults, so existing `MilvusException(code, message, compatible_code)` callers are unaffected.

## Test

Added `TestCheckStatusClassification` in `tests/unit/test_utils.py` covering input-error surfacing, retriable system error, unset defaults, and subclass preservation through `from_status`. `ruff` and `black` clean. Also validated end-to-end against a local standalone: `search`/`create` on a bad request surface `is_input_error=True`, and `describe` on a missing collection now carries the classification too.

### DIFF
--- a/pymilvus/client/async_grpc_handler.py
+++ b/pymilvus/client/async_grpc_handler.py
@@ -483,7 +483,7 @@ class AsyncGrpcHandler:
         if is_successful(status):
             return CollectionSchema(raw=response).dict()
 
-        raise DescribeCollectionException(status.code, status.reason, status.error_code)
+        raise DescribeCollectionException.from_status(status)
 
     @retry_on_rpc_failure()
     async def has_collection(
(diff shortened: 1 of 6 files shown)


XuanYang-cn:
@czs007 please increase ut cov to 90% to pass codecov/patch check

XuanYang-cn (approved):
/lgtm

PR #3629, opened 2026-06-24
feat(client): support MinHash function field in add_function_field

MrPresent-Han:
## What this PR does

Extends `add_function_field` (sync `MilvusClient` and `AsyncMilvusClient`) to support **MinHash** function fields, in addition to the existing BM25 support.

Previously `add_function_field` only accepted `FunctionType.BM25` producing a `SPARSE_FLOAT_VECTOR` output field. This PR generalizes the validation to a single source-of-truth map of accepted `(function type -> output dtype)` pairs:

| Function type | Required output field dtype |
| --- | --- |
| `FunctionType.BM25` | `SPARSE_FLOAT_VECTOR` |
| `FunctionType.MINHASH` | `BINARY_VECTOR` |

Validation errors are updated to reflect both supported combinations.

## Notes

- Pure client-side parameter validation; no proto/wire change is required — `FunctionType.MINHASH` and `DataType.BINARY_VECTOR` already exist in the generated proto.
- Behavior for BM25 is unchanged.

## Tests

- Parametrized the existing `add_function_field` unit tests to cover both BM25 and MinHash for the delegate path and the reject paths, on both the sync and async clients.
- Added `examples/test_add_function_field_minhash_search.py` demonstrating an end-to-end MinHash search on a `BINARY_VECTOR` field added at runtime.

```
$ pytest tests/test_milvus_client.py tests/test_async_milvus_client.py -k add_function_field
10 passed
```

### DIFF
--- a/examples/test_add_function_field_minhash_search.py
+++ b/examples/test_add_function_field_minhash_search.py
@@ -1,0 +1,507 @@
+"""
+Validate MinHash search on a binary field added through add_function_field.
+"""
+
+import logging
+import os
+import random
+import time
+from contextlib import contextmanager
+from typing import Any
+
+from pymilvus import DataType, FieldSchema, Function, FunctionType, MilvusClient
+from pymilvus.exceptions import MilvusException
+
+HOST = os.environ.get("MILVUS_URI", "http://localhost:19530")
+COLLECTION_NAME = os.environ.get(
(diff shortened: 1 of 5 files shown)


XuanYang-cn (approved):
/lgtm

PR #3631, opened 2026-06-25
fix: support struct vector fields in BulkWriter

zhuwenxing:
## What this PR does

- Adds `BulkFileType.JSONL` and writes one JSON object per line with `.jsonl` suffix.
- Allows Local/Remote BulkWriter to persist/upload JSONL files instead of skipping them.
- Allows BulkWriter struct-array subfields to use FLOAT_VECTOR, FLOAT16_VECTOR, BFLOAT16_VECTOR, INT8_VECTOR, and BINARY_VECTOR.
- Normalizes struct vector payloads for Parquet output, including FLOAT16/BFLOAT16 byte payloads.
- Uses `ml_dtypes` for BFLOAT16 JSON/CSV byte-vector decoding and includes it in the `bulk_writer` extra.
- Adds unit coverage for JSON, JSONL, CSV, and Parquet BulkWriter output across the supported struct vector subfield types.

## Tests

- `python -m pytest -q tests/unit/test_bulk_writer_buffer.py tests/unit/test_local_bulk_writer.py tests/unit/test_remote_bulk_writer.py`
  - 95 passed
- Milvus Python client E2E against local dev-cli deployment:
  - `test_import_struct_array_vector_subfield_all_file_types`
  - `test_import_struct_array_vector_subfield_with_local_bulk_writer`
  - 40 passed

### DIFF
--- a/pymilvus/bulk_writer/bulk_writer.py
+++ b/pymilvus/bulk_writer/bulk_writer.py
@@ -304,6 +304,24 @@ class BulkWriter:
 
         return row_size
 
+    def _normalize_struct_vector(self, origin_list: object, dtype: DataType):
+        if dtype == DataType.FLOAT_VECTOR:
+            return np.array(origin_list, dtype=NUMPY_TYPE_CREATOR[DataType.FLOAT.name])
+
+        if dtype in {DataType.FLOAT16_VECTOR, DataType.BFLOAT16_VECTOR}:
+            if self._file_type != BulkFileType.PARQUET:
+                return origin_list
+            if isinstance(origin_list, list):
+                values = np.asarray(origin_list, dtype=np.float32)
+                if dtype == DataType.FLOAT16_VECTOR:
+                    return values.astype(np.float16).view(np.uint8).tolist()
+                uint32_values = values.view(np.uint32)
+                return (uint32_values >> 16).astype(np.uint16).view(np.uint8).tolist()
+            if isinstance(origin_list, bytes):
+                return list(origin_list)
+
+        return origin_list
+
     def _verify_struct(self, x: object, field: StructFieldSchema):
         validator = TYPE_VALIDATOR[DataType.STRUCT.name]
         if not validator(x, field.max_capacity):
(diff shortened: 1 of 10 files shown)


zhuwenxing:
/assign @XuanYang-cn 
PTAL

pymilvus/bulk_writer/bulk_writer.py
```diff
@@ -304,6 +304,24 @@ def _verify_normal_field(self, row: dict):
 
         return row_size
 
+    def _normalize_struct_vector(self, origin_list: object, dtype: DataType):
+        if dtype == DataType.FLOAT_VECTOR:
+            return np.array(origin_list, dtype=NUMPY_TYPE_CREATOR[DataType.FLOAT.name])
+
+        if dtype in {DataType.FLOAT16_VECTOR, DataType.BFLOAT16_VECTOR}:
+            if self._file_type != BulkFileType.PARQUET:
+                return origin_list
```
XuanYang-cn, line 317: must-change: `_normalize_struct_vector` returns accepted FP16/BF16 `np.ndarray` inputs as raw bytes for JSON, JSONL, and CSV struct subfields; those nested bytes are not JSON serializable, so `LocalBulkWriter(..., file_type=JSON/JSONL/CSV)` fails before writing. Convert byte-form FP16/BF16 struct vectors to numeric lists for non-Parquet outputs, and add coverage using `np.ndarray(dtype=float16/bfloat16)` inputs for JSON, JSONL, and CSV.

XuanYang-cn (changes requested):
Found one blocking BulkWriter serialization regression.

nit: `pymilvus/bulk_writer/__init__.py` still omits `ml_dtypes` from the existing missing-package guard, so users who have the older bulk_writer extras except `ml_dtypes` get a raw import failure instead of the package install guidance; add `ml_dtypes` to `expected_pkgs`.

pyproject.toml
```diff
@@ -60,6 +60,7 @@ bulk_writer = [
     "requests",
     "minio>=7.0.0",
     "pyarrow>=12.0.0",
+    "ml_dtypes",
```
yhmo, line 63: pyproject.toml line:63
Low ---- Adding `ml_dtypes` here fixes fresh installs, but `pymilvus/bulk_writer/__init__.py` still only checks for minio/azure/requests/pyarrow before importing the module tree. In a partial or stale bulk-writer environment this will now fail with a raw `ModuleNotFoundError: ml_dtypes` from `buffer.py` instead of the package's existing "install pymilvus[bulk_writer]" guidance, so please add `ml_dtypes` to `expected_pkgs` too.

pymilvus/bulk_writer/bulk_writer.py
zhuwenxing: Done. Non-Parquet FP16/BF16 struct vector ndarray inputs are now normalized to numeric lists instead of nested bytes, and coverage was added for JSON, JSONL, and CSV outputs.

pyproject.toml
zhuwenxing: Done. Added `ml_dtypes` to `expected_pkgs` so partial bulk-writer installs still get the existing install guidance instead of a raw import error.

XuanYang-cn (approved):
/lgtm

PR #3636, opened 2026-06-25
feat: add function chain API

junjiejiangjjj:
issue: https://github.com/milvus-io/milvus/issues/50571

Introduce FunctionChain, FunctionChainExpr, ColumnRef, and helper functions for building function chain pipelines. Wire function_chains into search request preparation, update generated proto files, and add tests and examples.


pymilvus/function_chain/chain.py
```diff
@@ -0,0 +1,444 @@
...
+
+Examples:
+    Build a rerank chain that computes a score, sorts by it, and keeps the top
+    results::
+
+        from pymilvus.function_chain import FunctionChain, FunctionChainExpr
+        from pymilvus.function_chain import FunctionChainStage, col
+
+        chain = (
+            FunctionChain(FunctionChainStage.L0_RERANK, name="score_rerank")
```
liliu-z: The chain.py docstring examples (module docstring line 15, FunctionChainStage line 55, FunctionChain class line 288, plus the boost/normalize snippets in the ColumnRef and col docstrings) use FunctionChainStage.L0_RERANK and ad-hoc expression names (weighted_sum/boost/normalize) that don't match the search-usable API: search rejects any non-L2_RERANK chain (prepare.py:1639-1640) and the real helpers are fn.num_combine/decay/round_decimal (see examples/function_chain.py:94). The snippets only call to_proto() so they don't raise on their own, but copying the canonical example into a real search() call fails with 'Only L2_RERANK function chains are supported for search'. Update the docstrings to use L2_RERANK and the fn.* helpers so they match the working example.

pymilvus/client/prepare.py
```diff
@@ -1614,6 +1615,36 @@ def add_data(v: Any) -> schema_types.TemplateValue:
...
         return expression_template_values
 
+    @staticmethod
+    def function_chains_schema(
+        function_chains: Optional[Union[FunctionChain, List[FunctionChain]]],
+        ranker: Optional[Union[Function, FunctionScore]] = None,
+    ) -> List[schema_types.FunctionChain]:
+        if function_chains is None:
+            return []
+        if ranker is not None:
```
liliu-z, line 1627: An empty `function_chains=[]` slips past the `None` early-return at prepare.py:1620 and reaches the mutual-exclusivity check at line 1625, which then rejects the combination of `function_chains` and `ranker`. An empty list requests no chains and should behave identically to `None`, so passing `function_chains=[]` alongside a `ranker` (e.g., when the list is built conditionally and left empty) raises a spurious conflict error. Normalize the empty list to `None` before the conflict check so `[]` is treated like `None`.

pymilvus/function_chain/chain.py
```diff
@@ -0,0 +1,461 @@
...
+
+def _to_param_value(value: Any) -> schema_pb2.FunctionParamValue:
+    if value is None:
+        raise ParamError(message="Function chain parameters do not support None")
+
+    if isinstance(value, (bool, np.bool_)):
+        return schema_pb2.FunctionParamValue(bool_value=bool(value))
+
+    if isinstance(value, (int, np.integer)):
+        return schema_pb2.FunctionParamValue(int64_value=int(value))
```
liliu-z, line 136: In `_to_param_value`, an int is assigned straight to the int64 `int64_value` proto field with no bounds check, so a value outside int64 range (> 2^63-1) makes protobuf raise a bare `ValueError` instead of pymilvus's `ParamError` — inconsistent with the `None` and unsupported-type cases in the same function, which do raise `ParamError`. (The field is int64, not int32, so values up to 2^63-1 are fine — the reviewer's 2^31 bound is incorrect.) Range-check and raise `ParamError` on overflow.

pymilvus/function_chain/chain.py
```diff
@@ -0,0 +1,461 @@
...
+        return schema_pb2.FunctionParamValue(
+            array_value=schema_pb2.FunctionParamArray(
+                values=[_to_param_value(item) for item in value]
+            )
+        )
+
+    if isinstance(value, Mapping):
+        return schema_pb2.FunctionParamValue(
+            object_value=schema_pb2.FunctionParamObject(
+                fields={str(key): _to_param_value(item) for key, item in value.items()}
```
liliu-z: Nested param-dict keys are silently coerced via `str(key)` when building a `FunctionParamObject`, whereas top-level keys are validated in `_copy_param_map` and rejected with `ParamError` if non-string. A nested non-string key (e.g. `{"p": {1: ...}}`) is quietly stringified instead of rejected; apply the same key-type check to nested keys.

pymilvus/function_chain/chain.py
junjiejiangjjj: fixed

pymilvus/client/prepare.py
junjiejiangjjj: fixed

pymilvus/milvus_client/milvus_client.py
```diff
@@ -428,6 +429,7 @@ def search(
         partition_names: Optional[List[str]] = None,
         anns_field: Optional[str] = None,
         ranker: Optional[Union[Function, FunctionScore]] = None,
+        function_chains: Optional[Union[FunctionChain, List[FunctionChain]]] = None,
```
yhmo, line 432: pymilvus/milvus_client/milvus_client.py line:432
Low ---- The public `search()` signature now exposes `function_chains`, but the docstring below still documents `ranker` and `search_aggregation` without describing the new parameter (`pymilvus/milvus_client/milvus_client.py:443-458`). That leaves the generated API docs and IDE help out of sync with the shipped surface, and `Collection.search` has the same gap (`pymilvus/orm/collection.py:740-780`). Please add a short `function_chains` entry here and in `Collection.search`, including that it is only supported for ordinary search and is mutually exclusive with `ranker`.

pymilvus/milvus_client/milvus_client.py
```diff
@@ -482,6 +484,7 @@ def search(
             expr_params=kwargs.pop("filter_params", {}),
             timeout=timeout,
             ranker=ranker,
+            function_chains=function_chains,
```
yhmo, line 489: pymilvus/milvus_client/milvus_client.py line:487
Medium ---- `Prepare` already has good low-level serialization coverage in `tests/unit/test_function_chain.py`, but this new public wrapper path still lacks a regression test that `MilvusClient.search(..., function_chains=...)` forwards the exact object to `handler.search`. The neighboring wrapper tests already pin `search_aggregation` in `tests/unit/test_milvus_client.py:889-896` and `ranker`/`search_aggregation` in `tests/unit/test_async_milvus_client.py:213-235`, but there is no analogous assertion for `function_chains` on either wrapper. Since this PR extends both sync and async public search APIs, I’d add one forwarding test per wrapper so these call-throughs cannot drift later.

pymilvus/milvus_client/milvus_client.py
junjiejiangjjj: Thanks, fixed. Added `function_chains` documentation to both `MilvusClient.search` and `Collection.search`, including that it is for ordinary search and mutually exclusive with `ranker`.

pymilvus/milvus_client/milvus_client.py
junjiejiangjjj: Thanks, fixed. Added forwarding regression tests for both sync and async search wrappers to ensure `function_chains` is passed through to the handler unchanged.

Verified with:
`python -m pytest tests/unit/test_milvus_client.py::TestMilvusClientCRUD::test_search_passes_function_chains tests/unit/test_async_milvus_client.py::TestAsyncMilvusClientNewFeatures::test_search_passes_function_chains`

PR #3639, opened 2026-06-26
Update bulkwriter example for struct field

yhmo:


examples/bulk_import/bulk_writer_all_types.py
```diff
@@ -99,6 +148,9 @@ def build_schema():
     struct_schema.add_field("struct_double", DataType.DOUBLE)
     struct_schema.add_field("struct_varchar", DataType.VARCHAR, max_length=100)
     struct_schema.add_field("struct_float_vec", DataType.FLOAT_VECTOR, dim=DIM)
+    struct_schema.add_field("struct_binary_vec", DataType.BINARY_VECTOR, dim=DIM)
```
liliu-z, line 154: The struct schema adds `struct_binary_vec`/`struct_float16_vec`/`struct_int8_vec` but omits `struct_bfloat16_vec`, so the bespoke struct bfloat16 branch added by #3631 at `bulk_writer.py:322-323` is never exercised by this smoke test. A regression in the struct bfloat16 serialization path would therefore pass this smoke test undetected. Add a `struct_bfloat16_vec` field to the struct schema alongside the other three vector types.

PR #3646, opened 2026-07-01
fix: forward search group_by_fields instead of silently ignoring it

MrPresent-Han:
## What this PR does

Python high-level `MilvusClient.search()` silently dropped the plural
`group_by_fields` argument on the search-request prepare path. Only the
singular `group_by_field` was handled in `Prepare.search_requests_with_expr`;
`group_by_fields` was never read for search, so it was ignored — which also
bypassed the proxy's `group_by_fields` / `search_aggregation` mutual-exclusion
validation.

Mirror the existing singular `group_by_field` handling:
- reject the `search_aggregation` + `group_by_fields` combination client-side, and
- otherwise forward `group_by_fields` (comma-joined) into `search_params`, so the
  proxy's `group_by_fields` validation and group-by search apply.

The fix lives in the shared `Prepare.search_requests_with_expr`, so it covers both
the sync and async clients.

## Why

issue: milvus-io/milvus#50960

Previously `client.search(..., group_by_fields=[...], search_aggregation=...)`
succeeded and returned empty results instead of raising
`group_by_fields and search_aggregation cannot be used simultaneously`.

## Tests

`pytest tests/unit` — 331 passed. New coverage:
- `test_search_aggregation.py::TestPrepareIntegration` — plural mutex reject,
  forward-as-search-param, empty-list no-op, wrong-type guard.
- `test_async_milvus_client.py::test_search_passes_group_by_fields` — the
  high-level client forwards `group_by_fields` to the handler.

## Notes

- No proto change required.
- pymilvus-only fix. Flipping the `milvus-io/milvus` strict-xfail E2E test is a
  follow-up once a pymilvus release ships this change.

### DIFF
--- a/pymilvus/client/prepare.py
+++ b/pymilvus/client/prepare.py
@@ -1713,6 +1713,21 @@ class Prepare:
         if group_by_field is not None:
             search_params[GROUP_BY_FIELD] = group_by_field
 
+        # Plural group_by_fields must not be silently dropped on the search path.
+        # Mirror the singular handling above: reject the search_aggregation combination
+        # client-side, and otherwise forward the value so the proxy's group_by_fields
+        # validation and group-by search apply. See milvus-io/milvus#50960.
+        group_by_fields = kwargs.get(QUERY_GROUP_BY_FIELDS)
+        if group_by_fields is not None:
+            if not isinstance(group_by_fields, list):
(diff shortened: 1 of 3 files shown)


XuanYang-cn (approved):
/lgtm

PR #3649, opened 2026-07-02
fix: update schema key metadata on add_field

### LINKED ISSUE #3648: [Bug]: BulkWriter rejects schema created by MilvusClient.create_schema with primary field
## Current Behavior

`MilvusClient.create_schema()` returns a `CollectionSchema`, and `schema.add_field(..., is_primary=True)` can be used to create a collection. However, the same schema cannot be passed to `LocalBulkWriter` / `RemoteBulkWriter`.

The schema field has `is_primary=True`, but `schema.primary_field` remains `None`, so BulkWriter fails with:

```text
MilvusException: primary field is null
```

This is confusing because users naturally expect schemas created through `MilvusClient.create_schema()` to work with PyMilvus BulkWriter APIs.

## Expected Behavior

A schema built incrementally through `MilvusClient.create_schema()` should update its local key-field metadata when `add_field(..., is_primary=True)` is called.

`LocalBulkWriter` / `RemoteBulkWriter` should accept this schema the same way they accept an equivalent `CollectionSchema/FieldSchema` schema.

## Steps To Reproduce

```python
from tempfile import TemporaryDirectory

from pymilvus import CollectionSchema, DataType, FieldSchema, MilvusClient
from pymilvus.bulk_writer import BulkFileType, LocalBulkWriter


def try_writer(label, schema):
    print(label, "primary_field:", schema.primary_field)
    with TemporaryDirectory() as tmp:
        with LocalBulkWriter(schema=schema, local_path=tmp, file_type=BulkFileType.PARQUET) as writer:
            writer.append_row({"id": 1, "vector": [0.1] * 8})
            writer.commit()
            print(writer.batch_files)


client_schema = MilvusClient.create_schema(auto_id=False)
client_schema.add_field("id", DataType.INT64, is_primary=True)
client_schema.add_field("vector", DataType.FLOAT_VECTOR, dim=8)

legacy_schema = CollectionSchema(
    fields=[
        FieldSchema("id", DataType.INT64, is_primary=True),
        FieldSchema("vector", DataType.FLOAT_VECTOR, dim=8),
    ],
)

try_writer("MilvusClient.create_schema", client_schema)
try_writer("CollectionSchema/FieldSchema", legacy_schema)
```

## Actual Result

`MilvusClient.create_schema` path fails:

```text
primary_field: None
MilvusException: primary field is null
```

The equivalent `CollectionSchema/FieldSchema` path succeeds.

## Environment

- SDK: reproduced on current PyMilvus source checkout
- Python: 3.12
- OS: macOS
- Milvus server: not required; this reproduces locally in BulkWriter initialization

## Anything Else?

The likely cause is that `CollectionSchema.add_field()` appends the new `FieldSchema` but does not update local key-field metadata such as `_primary_field`, `_partition_key_field`, and `_clustering_key_field`.

This also affects SDK UX because `MilvusClient.create_schema()` explicitly supports incremental schema construction.

zhuwenxing:
## What this PR does

Fixes key-field metadata for schemas built incrementally with `MilvusClient.create_schema()` / `CollectionSchema.add_field()`.

`add_field(..., is_primary=True)` now updates `schema.primary_field` immediately, and key fields configured through schema kwargs are also reflected when their fields are added later. This lets BulkWriter accept schemas created through `MilvusClient.create_schema()`.

Fixes #3648.

## Changes

- Update `CollectionSchema.add_field()` to refresh primary, partition, and clustering key metadata.
- Preserve `auto_id=True` propagation to the primary field for incrementally built schemas.
- Add regression coverage for `MilvusClient.create_schema()` with `LocalBulkWriter`.

## Verification

```bash
uv run pytest tests/unit/orm/test_schema.py tests/unit/test_local_bulk_writer.py -q
uv run ruff check pymilvus/orm/schema.py tests/unit/orm/test_schema.py tests/unit/test_local_bulk_writer.py
```

### DIFF
--- a/pymilvus/orm/schema.py
+++ b/pymilvus/orm/schema.py
@@ -211,50 +211,62 @@ class CollectionSchema:
             raise ParamError(message="External collections do not support auto_id")
 
     def _update_key_field_metadata(self, field: Any):
+        primary_field = self._primary_field
+        partition_key_field = self._partition_key_field
+        clustering_key_field = self._clustering_key_field
         primary_field_name = self._kwargs.get("primary_field", None)
         partition_key_field_name = self._kwargs.get("partition_key_field", None)
         clustering_key_field_name = self._kwargs.get("clustering_key_field", None)
 
-        if primary_field_name and primary_field_name == field.name:
-            field.is_primary = True
+        try:
+            if primary_field_name and primary_field_name == field.name:
+                field.is_primary = True
 
-        if partition_key_field_name and partition_key_field_name == field.name:
-            field.is_partition_key = True
+            if partition_key_field_name and partition_key_field_name == field.name:
+                field.is_partition_key = True
 
-        if clustering_key_field_name and clustering_key_field_name == field.name:
-            field.is_clustering_key = True
+            if clustering_key_field_name and clustering_key_field_name == field.name:
+                field.is_clustering_key = True
 
-        if field.is_primary:
-            if self._primary_field is not None and self._primary_field.name != field.name:
-                msg = ExceptionsMessage.PrimaryKeyOnlyOne % (self._primary_field.name, field.name)
-                raise PrimaryKeyException(message=msg)
-            self._primary_field = field
-            if self._kwargs.get("auto_id", False):
-                self._primary_field.auto_id = True
+            if field.is_primary:
+                if self._primary_field is not None and self._primary_field.name != field.name:
+                    msg = ExceptionsMessage.PrimaryKeyOnlyOne % (
+                        self._primary_field.name,
+                        field.name,
+                    )
+                    raise PrimaryKeyException(message=msg)
+                self._primary_field = field
+                if self._kwargs.get("auto_id", False):
+                    self._primary_field.auto_id = True
 
-        if field.is_partition_key:
-            if (
-                self._partition_key_field is not None
-                and self._partition_key_field.name != field.name
-            ):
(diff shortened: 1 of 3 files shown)


XuanYang-cn:
@zhuwenxing Please add more ut to increase the patch code cov 90% to pass the check

zhuwenxing:
@XuanYang-cn I added UT coverage for the `add_field` key metadata conflict branches and rebased onto the latest `master` to resolve the merge conflicts. `codecov/patch`, `codecov/project`, and the GitHub Actions lint/proto/python test jobs are all passing now.

pymilvus/orm/schema.py
```diff
@@ -484,6 +530,7 @@ def add_field(self, field_name: str, datatype: DataType, **kwargs):
 
         field = FieldSchema(field_name, datatype, **kwargs)
         self._fields.append(field)
```
XuanYang-cn: This appends the new field before `_update_key_field_metadata()` validates duplicate key metadata. If `_update_key_field_metadata()` raises, for example when adding a second primary, partition, or clustering key, the exception leaves the rejected field in `schema.fields`, so callers that catch the exception keep a mutated invalid schema. Please validate before appending, or roll back the append on failure, and add a regression assertion that a failed `add_field()` leaves `schema.fields` unchanged.

XuanYang-cn (changes requested):
Requesting changes for the inline issue: failed add_field() validation currently leaves the rejected field in schema.fields. @zhuwenxing

PR #3659, opened 2026-07-03
fix(bulk_writer): normalize volume remote paths on Windows

VectorPeak:
Normalize `VolumeBulkWriter` remote volume paths so service-side volume paths use stable POSIX semantics and uploads land under the same path family that `batch_files` records.

## What Problem This Solves

`VolumeBulkWriter.__init__()` previously built the remote volume directory with host filesystem semantics:

```python
remote_dir_path = Path(remote_path) / super().uuid
self._remote_path = str(remote_dir_path) + "/"
```

On Windows, using `Path` for a service-side remote path can produce mixed separators in `_remote_path`. That is the wrong abstraction boundary: `remote_path` is a Milvus/Zilliz volume object prefix, not a client-local filesystem path.

There was also a second mismatch in the upload path. `VolumeBulkWriter._upload()` computes a per-file remote object path and records it in `batch_files`, but `_upload_object()` previously ignored that computed `object_name` and uploaded to `self._remote_path` as a directory instead.

That can make the writer record a path such as:

```text
test/data/<writer_uuid>/1/part.parquet
```

while the volume upload helper writes the single file under:

```text
test/data/<writer_uuid>/part.parquet
```

The mismatch matters because `batch_files` and `get_volume_upload_result()["path"]` are the import-path contract for generated bulk files.

## Change

- Treat `remote_path` as a service-side POSIX object prefix.
- Reject backslashes in `remote_path` instead of silently accepting Windows-local-style paths.
- Reject `.` and `..` path segments before storing `_remote_path`, so recorded paths and canonicalized upload targets cannot diverge.
- Build `_remote_path` with `PurePosixPath` and preserve the existing trailing slash.
- Make `_upload_object(file_path, object_name)` pass the computed object path's parent directory to `VolumeFileManager.upload_file_to_volume(...)`.
- Keep local staging, row serialization, volume credential refresh, retry handling, multipart upload, and the public `VolumeFileManager.upload_file_to_volume(...)` directory-upload contract unchanged.

## Evidence

Focused tests cover the final behavior:

```text
remote_path input: test/data
_remote_path after init: test/data/<uuid>/
```

Invalid remote prefixes are rejected before `_remote_path` is stored:

```text
remote_path input: test\data
result: ValueError("remote_path must be a POSIX path and must not contain backslashes")

remote_path input: test/../data
result: ValueError("remote_path must not contain '.' or '..' segments")

remote_path input: test/./data
result: ValueError("remote_path must not contain '.' or '..' segments")
```

The upload path now uses the computed object path's parent directory:

```text
object_name: test/data/<uuid>/1/local_file.parquet
upload_file_to_volume receives target path: test/data/<uuid>/1
```

The updated tests also assert that `get_volume_upload_result()["path"]` and `_upload()` object names do not contain Windows backslashes.

```text
User creates VolumeBulkWriter(remote_path=...)
  -> VolumeBulkWriter.__init__()
    -> validate remote_path as a POSIX volume prefix
    -> reject backslashes and dot segments
    -> store POSIX-style _remote_path
  -> append_row(...)
  -> commit()
  -> LocalBulkWriter._flush()
    -> persists files under local flush subdirectories
  -> VolumeBulkWriter._upload(file_list)
    -> computes remote_file_path from local relative path
    -> _upload_object(file_path, object_name=remote_file_path)
    -> VolumeFileManager.upload_file_to_volume(file_path, target parent path)
```

Affected surface: `VolumeBulkWriter` volume upload paths.

Sibling surfaces checked:

- `RemoteBulkWriter` has its own S3/Azure object-key handling and is not changed here.
- `VolumeFileManager.upload_file_to_volume(...)` remains a directory-upload helper for existing direct callers.
- Volume REST APIs, volume management, credential refresh, retry, and progress tracking are not changed by this PR.
- `LocalBulkWriter._exit()` now safely returns if validation fails before `_working_thread` is initialized, avoiding destructor warnings for partially initialized writer instances.

## Validation

- `uv run --python 3.11 --group dev black --check pymilvus/bulk_writer/volume_bulk_writer.py pymilvus/bulk_writer/local_bulk_writer.py tests/unit/test_bulk_writer_stage.py`
- `uv run --python 3.11 --group dev ruff check pymilvus/bulk_writer/volume_bulk_writer.py pymilvus/bulk_writer/local_bulk_writer.py tests/unit/test_bulk_writer_stage.py`
- `uv run --python 3.11 --group dev pytest tests/unit/test_bulk_writer_stage.py -k VolumeBulkWriter -q`
- `git diff --check -- pymilvus/bulk_writer/volume_bulk_writer.py pymilvus/bulk_writer/local_bulk_writer.py tests/unit/test_bulk_writer_stage.py`

Remote CI is also green for lint, proto checks, Python tests on Ubuntu/Windows, DCO, and Codecov patch/project coverage.

### DIFF
--- a/pymilvus/bulk_writer/volume_bulk_writer.py
+++ b/pymilvus/bulk_writer/volume_bulk_writer.py
@@ -1,5 +1,5 @@
 import logging
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from typing import Any, Dict, List, Optional
 
 from pymilvus.bulk_writer.volume_file_manager import VolumeFileManager
@@ -30,8 +30,8 @@ class VolumeBulkWriter(LocalBulkWriter):
         local_path = Path.cwd() / "bulk_writer"
         super().__init__(schema, str(local_path), chunk_size, file_type, config, **kwargs)
 
-        remote_dir_path = Path(remote_path) / super().uuid
-        self._remote_path = str(remote_dir_path) + "/"
+        remote_dir_path = PurePosixPath(remote_path.replace("\\", "/").strip("/")) / super().uuid
+        self._remote_path = f"{remote_dir_path.as_posix()}/"
         self._remote_files: List[List[str]] = []
         self._volume_name = volume_name
         self._volume_file_manager = VolumeFileManager(
@@ -110,5 +110,8 @@ class VolumeBulkWriter(LocalBulkWriter):
 
     def _upload_object(self, file_path: str, object_name: str):
         logger.info(f"Prepare to upload '{file_path}' to '{object_name}'")
-        self._volume_file_manager.upload_file_to_volume(file_path, self._remote_path)
+        target_volume_path = PurePosixPath(object_name).parent.as_posix()
+        if target_volume_path == ".":
+            target_volume_path = self._remote_path.rstrip("/")
+        self._volume_file_manager.upload_file_to_volume(file_path, target_volume_path)
         logger.info(f"Uploaded file '{file_path}' to '{object_name}'")
--- a/pymilvus/bulk_writer/volume_bulk_writer.py
+++ b/pymilvus/bulk_writer/volume_bulk_writer.py
@@ -11,6 +11,14 @@ from .local_bulk_writer import LocalBulkWriter
 logger = logging.getLogger(__name__)
 
 
+def _normalize_volume_remote_path(remote_path: str) -> str:
+    posix_path = remote_path.replace("\\", "/").strip("/")
+    if any(part in {".", ".."} for part in posix_path.split("/")):
+        msg = "remote_path must not contain '.' or '..' segments"
+        raise ValueError(msg)
+    return PurePosixPath(posix_path).as_posix()
+
+
 class VolumeBulkWriter(LocalBulkWriter):
     """VolumeBulkWriter handles writing local bulk files to a remote volume."""
 
@@ -27,10 +35,11 @@ class VolumeBulkWriter(LocalBulkWriter):
         connect_type: ConnectType = ConnectType.AUTO,
         **kwargs,
     ):
+        remote_base_path = _normalize_volume_remote_path(remote_path)
         local_path = Path.cwd() / "bulk_writer"
         super().__init__(schema, str(local_path), chunk_size, file_type, config, **kwargs)
 
-        remote_dir_path = PurePosixPath(remote_path.replace("\\", "/").strip("/")) / super().uuid
+        remote_dir_path = PurePosixPath(remote_base_path) / super().uuid
         self._remote_path = f"{remote_dir_path.as_posix()}/"
         self._remote_files: List[List[str]] = []
         self._volume_name = volume_name
--- a/pymilvus/bulk_writer/volume_bulk_writer.py
+++ b/pymilvus/bulk_writer/volume_bulk_writer.py
@@ -12,7 +12,11 @@ logger = logging.getLogger(__name__)
 
 
 def _normalize_volume_remote_path(remote_path: str) -> str:
-    posix_path = remote_path.replace("\\", "/").strip("/")
+    if "\\" in remote_path:
+        msg = "remote_path must be a POSIX path and must not contain backslashes"
+        raise ValueError(msg)
+
+    posix_path = remote_path.strip("/")
     if any(part in {".", ".."} for part in posix_path.split("/")):
         msg = "remote_path must not contain '.' or '..' segments"
         raise ValueError(msg)
@@ -120,7 +124,5 @@ class VolumeBulkWriter(LocalBulkWriter):
     def _upload_object(self, file_path: str, object_name: str):
         logger.info(f"Prepare to upload '{file_path}' to '{object_name}'")
         target_volume_path = PurePosixPath(object_name).parent.as_posix()
-        if target_volume_path == ".":
-            target_volume_path = self._remote_path.rstrip("/")
         self._volume_file_manager.upload_file_to_volume(file_path, target_volume_path)
         logger.info(f"Uploaded file '{file_path}' to '{object_name}'")
--- a/tests/unit/test_bulk_writer_stage.py
+++ b/tests/unit/test_bulk_writer_stage.py
@@ -862,9 +862,9 @@ class TestVolumeBulkWriter:
         assert volume_bulk_writer._volume_name == "test_volume"
         assert isinstance(volume_bulk_writer._volume_file_manager, MagicMock)
 
-    def test_init_normalizes_windows_remote_path(self, simple_schema: CollectionSchema) -> None:
-        with patch("pymilvus.bulk_writer.volume_bulk_writer.VolumeFileManager"):
-            writer = VolumeBulkWriter(
+    def test_init_rejects_windows_remote_path(self, simple_schema: CollectionSchema) -> None:
+        with pytest.raises(ValueError, match="remote_path must be a POSIX path"):
+            VolumeBulkWriter(
                 schema=simple_schema,
                 remote_path=r"test\data",
                 cloud_endpoint="https://api.cloud.zilliz.com",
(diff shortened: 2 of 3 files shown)


VectorPeak:
/assign @xuanyang-cn

Hi @xuanyang-cn, this PR is ready for review. It keeps the change scoped to `VolumeBulkWriter` remote path normalization on Windows and adds focused unit coverage for the path behavior. DCO is passing, and the local validation details are included in the PR description.

tests/unit/test_bulk_writer_stage.py
```diff
@@ -857,9 +857,29 @@ def volume_bulk_writer(self, simple_schema: CollectionSchema) -> VolumeBulkWrite
 
     def test_init(self, volume_bulk_writer: VolumeBulkWriter) -> None:
         assert volume_bulk_writer._remote_path.endswith("/")
+        assert volume_bulk_writer._remote_path.startswith("test/data/")
+        assert "\\" not in volume_bulk_writer._remote_path
         assert volume_bulk_writer._volume_name == "test_volume"
         assert isinstance(volume_bulk_writer._volume_file_manager, MagicMock)
 
+    def test_init_normalizes_windows_remote_path(
```
XuanYang-cn: Please run Black on this file. Current CI fails because Black rewrites this new test signature to one line.

pymilvus/bulk_writer/volume_bulk_writer.py
```diff
@@ -110,5 +110,8 @@ def _upload(self, file_list: List[str]) -> List[str]:
 
     def _upload_object(self, file_path: str, object_name: str):
         logger.info(f"Prepare to upload '{file_path}' to '{object_name}'")
-        self._volume_file_manager.upload_file_to_volume(file_path, self._remote_path)
+        target_volume_path = PurePosixPath(object_name).parent.as_posix()
+        if target_volume_path == ".":
```
XuanYang-cn: Please add coverage for this fallback branch, or remove it if it is not needed. Codecov reports patch coverage at 85.71%; add a focused case such as `_upload_object("local_file.parquet", "remote_file.parquet")` so patch coverage gets back to at least 90%.

pymilvus/bulk_writer/volume_bulk_writer.py
```diff
@@ -30,8 +30,8 @@ def __init__(
         local_path = Path.cwd() / "bulk_writer"
         super().__init__(schema, str(local_path), chunk_size, file_type, config, **kwargs)
 
-        remote_dir_path = Path(remote_path) / super().uuid
-        self._remote_path = str(remote_dir_path) + "/"
+        remote_dir_path = PurePosixPath(remote_path.replace("\\", "/").strip("/")) / super().uuid
```
XuanYang-cn: Please normalize or reject dot segments before storing `_remote_path`. `PurePosixPath` preserves internal `..`, while `VolumeFileManager._convert_dir_path()` later normalizes the upload target, so values like `test/../data` can make `batch_files` record a different path than the path actually uploaded.

XuanYang-cn (changes requested):
Please address the inline comments. In particular, patch coverage is currently 85.71%; please raise it to at least 90% before merge.

VectorPeak:
Addressed the review feedback in `723b4591`:

- Ran Black over the touched files and kept the new test signature formatted.
- Added focused coverage for the bare-file `_upload_object(local_file.parquet, remote_file.parquet)` fallback branch.
- Added `_normalize_volume_remote_path()` to normalize Windows separators before storing `_remote_path`, and reject `.` / `..` segments to avoid recorded/uploaded path divergence.
- Amended the commit with DCO sign-off.

Local validation:

- `uv run --python 3.11 --group dev black --check pymilvus/bulk_writer/volume_bulk_writer.py tests/unit/test_bulk_writer_stage.py`
- `uv run --python 3.11 --group dev ruff check pymilvus/bulk_writer/volume_bulk_writer.py tests/unit/test_bulk_writer_stage.py`
- `uv run --python 3.11 --group dev pytest tests/unit/test_bulk_writer_stage.py -k VolumeBulkWriter -q` (`15 passed, 44 deselected`)
- `git diff --check -- pymilvus/bulk_writer/volume_bulk_writer.py tests/unit/test_bulk_writer_stage.py`

@XuanYang-cn could you please re-review when CI/Codecov finishes rerunning?

VectorPeak:
Hi @XuanYang-cn, thanks again for the careful review.

I ran another independent pass over the latest head `723b4591` and confirmed the three requested updates are covered:

- Black formatting is clean for the touched test file.
- The bare-file `_upload_object('local_file.parquet', 'remote_file.parquet')` fallback now has focused coverage.
- `VolumeBulkWriter` now normalizes Windows separators before storing `_remote_path` and rejects `.` / `..` segments before the path is persisted.

Local validation also passes for Black, Ruff, `git diff --check`, and the focused `TestVolumeBulkWriter` pytest selection. Please feel free to take another look whenever convenient.

pymilvus/bulk_writer/volume_bulk_writer.py
```diff
@@ -110,5 +119,8 @@ def _upload(self, file_list: List[str]) -> List[str]:
 
     def _upload_object(self, file_path: str, object_name: str):
         logger.info(f"Prepare to upload '{file_path}' to '{object_name}'")
-        self._volume_file_manager.upload_file_to_volume(file_path, self._remote_path)
+        target_volume_path = PurePosixPath(object_name).parent.as_posix()
+        if target_volume_path == ".":
```
XuanYang-cn: The `rstrip("/")` here is redundant because `upload_file_to_volume()` canonicalizes the
  target directory with `rstrip("/") + "/"` anyway. Also, this fallback branch does not
  actually honor a bare `object_name` as the remote filename; the volume helper uploads
  the source file basename. Since `_upload()` always passes a full remote object path with
  a parent, can we remove the bare-name fallback instead of adding special behavior/tests
  for it?

pymilvus/bulk_writer/volume_bulk_writer.py
```diff
@@ -11,6 +11,14 @@
 logger = logging.getLogger(__name__)
 
 
+def _normalize_volume_remote_path(remote_path: str) -> str:
+    posix_path = remote_path.replace("\\", "/").strip("/")
```
XuanYang-cn: I think it's really odd for an user to pass in a windows-local-style remote_path, the remote are always posix, it should rasie exception if encounters "\\"

XuanYang-cn: I think we should treat `remote_path` as a service-side POSIX object prefix, not as a
  client-local filesystem path. The Windows bug comes from using `pathlib.Path` to build a
  remote path, not necessarily from users passing Windows paths.

  Can we avoid host `Path` for remote paths and use `PurePosixPath` consistently here? I
  would also prefer not silently accepting backslashes unless we clearly want to support
  non-POSIX remote prefixes.

pymilvus/bulk_writer/volume_bulk_writer.py
```diff
@@ -27,11 +35,12 @@ def __init__(
         connect_type: ConnectType = ConnectType.AUTO,
         **kwargs,
     ):
+        remote_base_path = _normalize_volume_remote_path(remote_path)
```
XuanYang-cn, line 42: This validation currently runs before `LocalBulkWriter.__init__()`. When it raises,
  Python later calls `LocalBulkWriter.__del__()` on a partially initialized object and
  emits `PytestUnraisableExceptionWarning` because `_working_thread` was never created.

  Please either move validation after base initialization, or make the destructor path
  safe for partially initialized instances. The new invalid-path tests currently pass but
  emit this warning.

XuanYang-cn (changes requested):

VectorPeak:
Hi @XuanYang-cn, thanks again for the careful review and for clarifying the intended abstraction.

I realized my earlier framing was off: this PR should not be about accepting Windows-style `remote_path` values from users and normalizing them. The better framing is that `remote_path` is a service-side POSIX object prefix, and the client should not let Windows local filesystem semantics leak into the internally constructed remote volume path.

I updated the PR body to match the current implementation and reviewer intent:

- `remote_path` is treated as POSIX-only.
- Backslashes are rejected instead of silently normalized.
- `.` / `..` path segments are rejected before `_remote_path` is stored.
- Remote path construction uses `PurePosixPath` rather than host `Path`.
- The bare-name `_upload_object` fallback was removed.
- The partial-initialization destructor path is guarded.

CI, DCO, and Codecov are currently green. If you think this is still worth carrying as a small correctness/compatibility fix for `VolumeBulkWriter`, I would appreciate another review when convenient. If you feel the value is too small after narrowing the semantics this way, I am also happy to close it rather than keep noise in the queue.

PR #3660, opened 2026-07-06
fix: allow all specified function chain stages in search

junjiejiangjjj:
Relax search-side function chain validation to reject only UNSPECIFIED while forwarding all explicit stages to the server. Update unit tests to cover the new validation behavior.

issue: https://github.com/milvus-io/milvus/issues/51011


XuanYang-cn (approved):
/lgtm

PR #3670, opened 2026-07-07
feat(milvus_client): require bound index_params in add_function_field

MrPresent-Han:
issue: milvus-io/milvus#51104

## Summary
- `MilvusClient.add_function_field` / `AsyncMilvusClient.add_function_field` now take a **mandatory** `index_params` (an `IndexParams` with exactly one entry and an explicit `index_type` for the function output field); missing/empty/AUTOINDEX-only params are rejected client-side.
- The bound index definition is serialized into the existing (previously unused) `AlterCollectionSchemaRequest.FieldInfo.index_name` / `extra_params` fields — **no milvus-proto change needed** — so the server can create the index meta atomically with the schema change and schema-bump/backfill compaction is never blocked by a missing vector index.
- Threaded through `_alter_collection_schema` → grpc handlers (sync + async) → `Prepare.alter_collection_schema_request`, using the same KV encoding as `create_index`.

## Merge order
This PR must merge **before** the server-side PR in milvus-io/milvus (linked from the issue); the server keeps rejecting index-less requests either way, so ordering only affects e2e suite runs.

### DIFF
--- a/pymilvus/client/prepare.py
+++ b/pymilvus/client/prepare.py
@@ -397,6 +397,8 @@ class Prepare:
         drop_field_id: Optional[int] = None,
         drop_function_name: Optional[str] = None,
         drop_function_output_fields: bool = False,
+        index_name: str = "",
+        index_extra_params: Optional[Dict] = None,
     ) -> milvus_types.AlterCollectionSchemaRequest:
         is_drop = any(
             identifier is not None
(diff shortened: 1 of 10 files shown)


pymilvus/client/prepare.py
```diff
@@ -443,7 +445,16 @@ def alter_collection_schema_request(
 
                 field_info = milvus_types.AlterCollectionSchemaRequest.FieldInfo(
                     field_schema=field_schema_proto,
+                    index_name=index_name,
                 )
+                if index_extra_params:
+                    for tk, tv in index_extra_params.items():
+                        if tk == "dim" and (not tv or not isinstance(tv, int)):
+                            raise ParamError(message="dim must be of int!")
```
XuanYang-cn: I think no one will pass in dim for index params right?

XuanYang-cn (changes requested):

pymilvus/client/prepare.py
MrPresent-Han: Right — for function output fields `dim` never comes from index params (BM25 sparse has no dim; MinHash dim lives in the field schema and the server fills/validates it from there), so this client-side guard was redundant. Removed in 5cf83629.

XuanYang-cn (approved):
/lgtm

PR #3679, opened 2026-07-10
feat(function_chain): add xgboost expression helper

junjiejiangjjj:
issue: https://github.com/milvus-io/milvus/issues/51192

Add fn.xgboost for building XGBoost FunctionChain expressions with FileResource-backed UBJ models. Validate feature column refs, model_resource, and output mode, and serialize to the xgboost expression params expected by Milvus.

Add unit coverage for default/raw output serialization and invalid parameter cases.

### DIFF
--- a/pymilvus/function_chain/fn.py
+++ b/pymilvus/function_chain/fn.py
@@ -6,6 +6,7 @@ from .chain import ColumnRef, FunctionChainExpr
 
 _NUM_COMBINE_MODES = {"multiply", "sum", "max", "min", "avg", "weighted"}
 _DECAY_FUNCTIONS = {"gauss", "exp", "linear"}
+_XGBOOST_OUTPUTS = {"default", "raw"}
 
 
 def _check_column_ref(value: Any, param_name: str) -> None:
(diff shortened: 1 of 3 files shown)


pymilvus/function_chain/fn.py
```diff
@@ -204,6 +205,75 @@ def round_decimal(value: ColumnRef, *, decimal: int) -> FunctionChainExpr:
...
+            ``ColumnRef``, ``model_resource`` is empty or not a string, or
+            ``output`` is not one of ``"default"`` or ``"raw"``.
+    """
+    if not features:
+        raise ParamError(message="fn.xgboost requires at least one feature column")
+    for value in features:
+        _check_column_ref(value, "fn.xgboost feature")
+    if not isinstance(model_resource, str) or not model_resource:
+        raise ParamError(message="xgboost model_resource must be a non-empty string")
+    if output not in _XGBOOST_OUTPUTS:
```
liliu-z: At fn.py:265 the `output not in _XGBOOST_OUTPUTS` set-membership test raises a bare `TypeError: unhashable type` for an unhashable `output` (e.g. `output=[]` or `output={}`) instead of a `ParamError` naming the allowed values — an internal inconsistency, since `model_resource` one line above is isinstance-guarded and `output=1` is correctly rejected. An `isinstance(output, str)` guard before the check (plus an `output=[]` regression test) would close it, but note the identical pattern already exists in `num_combine` (line 81) and `decay` (line 150), so this is a pre-existing convention rather than an idiom this helper fails to mirror; guarding only xgboost would make the three inconsistent.

XuanYang-cn:
Relase candidates: 
- [x] milvus 3.0
~~- [ ] milvus 2.6~~

PR #3681, opened 2026-07-14
fix: unpack async schema result in optimize

### LINKED ISSUE #3680: [Bug]: AsyncMilvusClient.optimize fails because _get_schema result is not unpacked
## Environment

- pymilvus: `3.1.0rc62`
- pymilvus master: reproduced by source inspection at `3b7d58e6`
- Python: 3.12.12
- Milvus server: `master-20260713-cdd9136fd7`
- Deployment: 1-CU cluster with MixCoord, Proxy, QueryNode, DataNode, StreamingNode, and Woodpecker

## Problem

`AsyncMilvusClient.optimize()` always fails before compaction submission because `_list_vector_indexes()` treats the result of `AsyncGrpcHandler._get_schema()` as a dictionary. The handler contract returns `(schema_dict, schema_timestamp)`.

The synchronous implementation already unpacks the tuple correctly.

## Minimal reproduction

```python
import asyncio
import time

from pymilvus import AsyncMilvusClient


async def main():
    client = AsyncMilvusClient(uri="http://localhost:19530")
    name = f"async_optimize_repro_{int(time.time())}"
    try:
        await client.create_collection(name, dimension=8)
        await client.optimize(name, target_size="64MB", timeout=30)
    finally:
        await client.drop_collection(name)
        await client.close()


asyncio.run(main())
```

Reproduction rate: 2/2 with the native `AsyncMilvusClient`, without any test wrapper.

## Actual result

```text
AttributeError: 'tuple' object has no attribute 'get'

pymilvus/milvus_client/async_milvus_client.py, in _list_vector_indexes
    for field in schema_dict.get("fields", [])
```

A direct contract probe shows:

```text
type(await conn._get_schema(collection_name)) == tuple
element types: dict, int
```

## Expected result

Async optimize should discover vector indexes, submit Force Merge, wait for compaction/index rebuild, refresh the loaded collection when needed, and return `OptimizeResult`.

## Root cause

The async implementation assigns the complete tuple to `schema_dict`:

```python
schema_dict = await conn._get_schema(...)
```

It should unpack the schema and timestamp, matching other async call sites and the synchronous implementation:

```python
schema_dict, _ = await conn._get_schema(...)
```

The current async optimize unit tests mock `_get_schema.return_value` as a bare dictionary, which does not match the real handler contract and masks the bug. A regression test should return `(schema, timestamp)`.

## Related test task

- milvus-io/milvus#51248 requires dedicated live `AsyncMilvusClient.optimize()` coverage and disallows permanent skip/xfail markers.

zhuwenxing:
## What changed

- Unpack the `(schema_dict, schema_timestamp)` tuple returned by `AsyncGrpcHandler._get_schema()` before discovering vector indexes in `AsyncMilvusClient.optimize()`.
- Update the async optimize unit tests to mock the real handler return contract.

## Why

`AsyncMilvusClient._list_vector_indexes()` treated the complete `_get_schema()` result as a dictionary. Because the async handler returns a tuple, every optimize call failed with `AttributeError: 'tuple' object has no attribute 'get'` before compaction could be submitted.

The synchronous client and other async call sites already unpack this result correctly.

## User impact

`AsyncMilvusClient.optimize()` can now discover vector indexes and complete the compaction, index rebuild, and load refresh workflow.

## Validation

- `pytest -q tests/unit/test_async_milvus_client_ops.py` — 80 passed
- `pytest -q tests/unit/test_optimize_task.py` — 40 passed
- Async handler schema contract tests — 2 passed
- Synchronous vector-index comparison tests — 2 passed
- Ruff check and format check passed
- Live validation against `master-20260714-1acbe20`: inserted and flushed three batches (1,500 rows total), then completed `AsyncMilvusClient.optimize(target_size="1GB")` with compaction ID `467664907943663510`; server state reached `Completed`.

Fixes #3680

### DIFF
--- a/pymilvus/milvus_client/async_milvus_client.py
+++ b/pymilvus/milvus_client/async_milvus_client.py
@@ -2176,7 +2176,7 @@ class AsyncMilvusClient(BaseMilvusClient):
         self, collection_name: str, timeout: Optional[float] = None, **kwargs
     ) -> List[str]:
         conn = await self._get_connection()
-        schema_dict = await conn._get_schema(
+        schema_dict, _ = await conn._get_schema(
             collection_name,
             timeout=timeout,
             context=self._generate_call_context(**kwargs),
--- a/tests/unit/test_async_milvus_client_ops.py
+++ b/tests/unit/test_async_milvus_client_ops.py
@@ -502,7 +502,7 @@ class TestAsyncClientOptimize:
         handler.get_compaction_state.return_value = state
         handler.get_load_state.return_value = LoadState.NotLoad
         schema = {"fields": [{"name": "id", "type": DataType.INT64}]}
-        handler._get_schema.return_value = schema
+        handler._get_schema.return_value = (schema, 0)
         handler.list_indexes.return_value = []
         task = MagicMock()
         task.check_cancelled = MagicMock()
@@ -521,7 +521,7 @@ class TestAsyncClientOptimize:
         handler.get_compaction_state.return_value = state
         handler.get_load_state.return_value = LoadState.NotLoad
         schema = {"fields": [{"name": "id", "type": DataType.INT64}]}
-        handler._get_schema.return_value = schema
+        handler._get_schema.return_value = (schema, 0)
         handler.list_indexes.return_value = []
         result = await client.optimize("col", wait=True)
         assert isinstance(result, OptimizeResult)


zhuwenxing:
/assign @XuanYang-cn 

PTAL

PR #3686, opened 2026-07-14
fix(settings): only load MILVUS_ prefixed keys from dotenv to avoid env pollution

### LINKED ISSUE #3666: [Bug]: load_dotenv() in pymilvus/settings.py causes global environment pollution at import time

### Describe the bug

Importing pymilvus unconditionally calls `load_dotenv()`, which injects all `.env` keys into `os.environ` for the entire process. This breaks test isolation for any project that uses pymilvus alongside a .env-file containing non-Milvus variables.

`load_dotenv()` is meant to be called by the **application** entry point, not by libraries.

python-dotenv's own docs added PYTHON_DOTENV_DISABLED=1 specifically for cases where "you can't modify third-party package calls". This implicitly acknowledges that libraries should not be calling load_dotenv(). All official usage examples position it at the application entry point, not inside library code; using this flag disables loading dotenv process-wide, including our own legitimate usage in our application's `main.py`. 

### Expected Behavior

Importing `pymilvus` should not produce side effects on the global process environment. The library should read its configuration from `os.environ` without modifying it, leaving the responsibility of populating the environment to the application entry point. Non-Milvus variables present in a project's .env file should never be injected into `os.environ` as a consequence of `import pymilvus`.

```markdown
# 1. Create a .env file in the project root with any non-Milvus variable:
# MY_APP_SECRET=sensitive_value
   
# 2. In a Python script or test:

import os

print(os.environ.get("MY_APP_SECRET"))  # None

import pymilvus  # noqa: F401

print(os.environ.get("MY_APP_SECRET"))  # "sensitive_value"
   
# The mere act of importing pymilvus injects every key from .env into os.environ, including variables entirely unrelated to Milvus.
```

### Environment details

```markdown
- Hardware/Softward conditions (OS, CPU, GPU, Memory): Windows 11
- Method of installation (Docker, or from source): pypi
- Milvus version (v0.3.1, or v0.4.0): v2.6.15
```

### Anything else?

_No response_

saitejabandaru-in:
Resolves #3666.

### Description
Currently, `pymilvus` calls `load_dotenv()` unconditionally in global namespace when importing `pymilvus/settings.py`. This causes side effects by loading all key-value pairs from `.env` (such as app secrets or credentials for other systems) into the process-wide `os.environ` environment variables, polluting the host application space.

To solve this, this change replaces the unconditional global `load_dotenv()` with a structured check using `dotenv_values(find_dotenv(usecwd=True))` that filters and only injects keys prefixed with `MILVUS_` (e.g. `MILVUS_URI`, `MILVUS_CONN_ALIAS`) into `os.environ` if they are not already set.

### Changes
- Updated `pymilvus/settings.py` to use `find_dotenv(usecwd=True)` + `dotenv_values()` and only inject keys beginning with `MILVUS_`.
- Added a new unit test suite in `tests/unit/test_settings.py` checking that `MILVUS_` keys are loaded, while non-Milvus keys (like `MY_APP_SECRET`) remain untouched in the host environment.

### DIFF
--- a/pymilvus/settings.py
+++ b/pymilvus/settings.py
@@ -1,9 +1,16 @@
 import logging.config
 import os
 
-from dotenv import load_dotenv
-
-load_dotenv()
+if os.getenv("PYTHON_DOTENV_DISABLED") != "1":
+    try:
+        from dotenv import dotenv_values, find_dotenv
+        dotenv_path = find_dotenv(usecwd=True)
+        if dotenv_path:
+            for k, v in dotenv_values(dotenv_path).items():
+                if k.startswith("MILVUS_") and v is not None and k not in os.environ:
+                    os.environ[k] = v
+    except ImportError:
+        pass
 
 
 class Config:
--- a/tests/unit/test_settings.py
+++ b/tests/unit/test_settings.py
@@ -1,0 +1,102 @@
+import os
+import sys
+from importlib import reload
+
+def test_settings_dotenv_no_pollution(tmp_path):
+    old_uri = os.environ.get("MILVUS_URI")
+    old_secret = os.environ.get("MY_APP_SECRET")
+    old_disabled = os.environ.get("PYTHON_DOTENV_DISABLED")
+
+    if "MILVUS_URI" in os.environ:
+        del os.environ["MILVUS_URI"]
+    if "MY_APP_SECRET" in os.environ:
+        del os.environ["MY_APP_SECRET"]
+    if "PYTHON_DOTENV_DISABLED" in os.environ:
+        del os.environ["PYTHON_DOTENV_DISABLED"]
+
+    orig_cwd = os.getcwd()
+    try:
+        os.chdir(tmp_path)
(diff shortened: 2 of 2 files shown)


saitejabandaru-in:
Hi, I have updated the PR to address Codex feedback:
1. Gated the dotenv loading check on `PYTHON_DOTENV_DISABLED != "1"`.
2. Added a check `v is not None` to prevent errors from valueless environment variable declarations in `.env`.
3. Signed-off the commits to satisfy the DCO requirements.

saitejabandaru-in:
I have checked the codebase and both fixes have been applied: 
1. Respecting .
2. Skipping valueless dotenv entries ().
Thanks for the review!

saitejabandaru-in:
Both fixes have been verified/applied:
1. Respecting `PYTHON_DOTENV_DISABLED`.
2. Skipping valueless dotenv entries with `v is not None`.

Thanks for the review!

PR #3690, opened 2026-07-14
feat: support bloom_match filter parameters

xiaofan-luan:
## Summary

Adds PyMilvus support for the `bloom_match(field, {blob})` expression introduced by milvus-io/milvus#51140.

- exports `build_bloom_filter(members, fpr=0.005)`, a dependency-free MBF1 / Parquet Split-Block Bloom filter builder for homogeneous `int` (signed int64) or `str` members;
- encodes the resulting blob as native `TemplateValue.bytes_val`, with no base64 expansion;
- updates the `milvus-proto` submodule and generated Python stubs to proto #632, which added `TemplateValue.bytes_val`;
- proves byte-for-byte parity against canonical int64 and Unicode-string golden vectors plus a checked-in Apache Arrow C++ fixture containing 100 deterministic random int64 members.

## Dependencies

- Server feature: milvus-io/milvus#51140
- Proto field: milvus-io/milvus-proto#632
- Design: [Bloom Filter Membership Expression](https://github.com/xiaofan-luan/milvus/blob/feat-bloom-filter-expr/docs/design-docs/design_docs/20260707-bloom-filter-expression.md)

## Validation

- `uv run pytest -q tests/unit/test_bloom_filter.py tests/unit/prepare/test_search.py` — 148 passed
- `uv run ruff check pymilvus/client/bloom_filter.py pymilvus/client/prepare.py tests/unit/test_bloom_filter.py`
- `uv run black --check pymilvus/client/bloom_filter.py pymilvus/client/prepare.py tests/unit/test_bloom_filter.py`
- `git diff --check`


pymilvus/client/bloom_filter.py
```diff
@@ -0,0 +1,147 @@
...
+        not isinstance(fpr, (int, float))
+        or not math.isfinite(fpr)
+        or not _MIN_FPR <= fpr <= _MAX_FPR
+    ):
+        raise ParamError(message=f"bloom filter fpr must be in [{_MIN_FPR}, {_MAX_FPR}]")
+    if not isinstance(members, (list, tuple)):
+        raise ParamError(message="bloom filter members must be a list or tuple of int or str")
+    if members and all(isinstance(value, int) and not isinstance(value, bool) for value in members):
+        domain = _DOMAIN_INT64
+        encoded = []
```
XuanYang-cn: This materializes a second list containing one `bytes` object per member, while the original Python member list remains resident. For the design's documented 10M-member use case, that is roughly another 490 MiB on top of the roughly 360 MiB input list and the Bloom body; at the default ~24M-member capacity tier, peak memory can exceed 2 GiB. This undermines the large-set use case. Please size the body from `len(members)`, then validate, encode, and hash each value directly without retaining the complete `encoded` copy.

tests/unit/test_bloom_filter.py
```diff
@@ -0,0 +1,66 @@
...
+    members = [-(1 << 63), -1, 0, 1, 2, 42, 1000000007, (1 << 63) - 1]
+    expected = bytes.fromhex(
+        "4d424631010001000800000000000000fca9f1d24d62503f0100000001000000"
+        "0117100a804845062834804200440612100448a40680089c4040542838410046"
+    )
+
+    assert build_bloom_filter(members, fpr=0.001) == expected
+
+
+def test_build_bloom_filter_matches_parquet_sbbf_string_vector():
```
XuanYang-cn, line 30: Patch coverage is currently 82.14%; please raise it above 90%. In particular, every string vector here is shorter than 32 bytes, so the XXH64 long-input branch is untested. Please add golden vectors around the 31/32/33-byte boundary, plus FPR boundary and signed-int64 overflow cases.

XuanYang-cn (changes requested):
The wire format and client transport are aligned with the Milvus design. Please address the two inline comments before approval.

pymilvus/client/bloom_filter.py
xiaofan-luan: Fixed in `769bcca2`. The intermediate `encoded` list is gone — the body is sized from `len(members)`, and each value is validated, encoded and hashed in a single pass, so no second per-member `bytes` object is retained.

Measured at 10M int64 members: the `encoded` list alone accounted for ~490 MiB (10M × 41 B for the `bytes` objects plus 8 B list slots), and removing it took overall peak from ~1075 MiB to ~585 MiB.

`ee7cc6e4` then takes it further. Of the 48.1 MiB the builder itself still used, only 16 MiB was the actual body — `header + bytes(body)` keeps the bytearray, its `bytes()` copy *and* the joined result alive at the same time, i.e. 3× the body. The header is now reserved in the buffer up front and the blob is assembled in place, so peak is 2× instead:

| | body | builder peak |
|---|---:|---:|
| before `769bcca2` | 16 MiB | ~538 MiB |
| `769bcca2` | 16 MiB | 48.1 MiB |
| `ee7cc6e4` | 16 MiB | **33.1 MiB** |

tests/unit/test_bloom_filter.py
xiaofan-luan: Fixed in `769bcca2`. Added golden vectors at 31/32/33 bytes — that is what actually reaches the XXH64 `length >= 32` stripe branch you pointed at — plus FPR boundaries (`0.0001`/`0.05` accepted, `0.00009`/`0.051`/`nan`/`inf` rejected) and signed-int64 overflow (`1<<63`, `-(1<<63)-1`).

`codecov/patch` now reports 100% of modified lines covered, project 94.08% → 94.12%.

`ee7cc6e4` adds more, because it introduces accelerated paths that have to be pinned to the reference implementation:

- `test_xxh64_python_matches_xxhash_c` — every length 0..79 plus 500 random variable-length payloads
- `test_int64_vectorised_matches_scalar` — includes the 16384 chunk boundary ±1
- `test_int64_vectorised_rejects_bad_members` — bool / float / str / None / overflow all surface as `ParamError`
- `test_build_bloom_filter_fallback_matches_default_path` — monkeypatches the pure-Python hashes back in, so the fallback stays exercised even when the C extra is installed

The pure-Python implementations are retained as `_xxh64_python` / `_xxh64_int64_python` precisely so all three hash routes stay mutually verifiable.

xiaofan-luan:
Both review comments are addressed (see the inline replies). `ee7cc6e4` additionally makes the build substantially faster, since the large-member-set case the memory comment was about was also spending a lot of wall clock in the pure-Python hash loop.

**10M members**

| | before | after | |
|---|---:|---:|---|
| INT64 | 29.03 s | **0.96 s** | 30x |
| VARCHAR (UUID, 36B) | 61.76 s | **13.84 s** | 4.5x |
| builder peak memory | 48.1 MiB | **33.1 MiB** | −31% |

**What changed**

1. **INT64 hashing is vectorised with numpy**, 16K members per chunk. Chunking is required, not incidental: a 256K chunk falls off cache and was 1.7x slower while using 26 MiB more. numpy is already a hard dependency (via pandas, and `client/types.py` imports it), so this adds nothing new.
2. **Optional `xxhash` C accelerator for the VARCHAR path** (`pip install "pymilvus[bloom_filter]"`). Variable-length input cannot be vectorised, since XXH64's stripe and tail loops depend on each member's byte length. Without the extra, the pure-Python fallback runs and the blob is identical.
3. Blob assembled in a single buffer, and one `struct` call per SBBF block instead of eight.

**The output is byte-for-byte unchanged.** Verified against the previous implementation across 66 size/fpr/domain configurations including chunk boundaries, on top of the existing Arrow C++ and Parquet golden vectors, which pass unchanged on every path.

**Semantics are preserved exactly.** This needed care: `np.fromiter` silently coerces where the scalar path raises — `bool` (an int subclass) becomes 0/1 and `float` is truncated. A C-speed `set(map(type, ...))` screen up front rejects both, while still accepting genuine `int` subclasses like the old `isinstance` check did. Range validation is now done by `np.fromiter` in C, with `OverflowError` translated back to the same `ParamError`.

**One thing that needs your call:** the optional `xxhash` dependency. I also added it to the `dev` group so CI exercises the accelerated path rather than only the fallback (`uv.lock` updated accordingly, since CI uses `--frozen`). If you would rather not take the dependency, dropping it is clean — the two accelerated paths are independent, so removing the `try/except` block and its tests leaves the 30x INT64 improvement intact.

Local runs: full suite 4499 passed with the extra installed, 4498 passed / 4 skipped with it hidden.

pymilvus/client/bloom_filter.py
```diff
@@ -0,0 +1,310 @@
...
+    # np.fromiter coerces silently where the scalar path raises: bool (an int subclass) becomes
+    # 0/1 and float is truncated. Screen the distinct types up front -- map(type, ...) runs at C
+    # speed, and the common all-int case settles in one set comparison.
+    member_types = set(map(type, members))
+    if member_types != {int}:
+        for member_type in member_types:
+            if member_type is bool or not issubclass(member_type, int):
+                raise ParamError(message="bloom filter members must be all int or all str")
+
+    words = np.frombuffer(buf, dtype=np.uint32, offset=_HEADER_SIZE)
```
XuanYang-cn: **Please use an explicit little-endian dtype here.** `np.uint32` uses the host's native byte order, while the SBBF wire format requires every block word to be encoded as little-endian uint32. On a big-endian client such as `s390x` or `ppc64`, this path writes byte-swapped words; a little-endian QueryNode then checks different bit positions, potentially introducing false negatives and breaking cross-language byte-for-byte parity. The header and scalar path already specify `<` explicitly. Please change this to `dtype="<u4"` (or a named little-endian dtype constant) and add a regression check.

XuanYang-cn (changes requested):
The previous memory and coverage concerns are addressed. One cross-platform wire-format correctness issue remains; please see the inline comment.

pymilvus/client/bloom_filter.py
xiaofan-luan: Fixed in `7c09786c`. The wire view is pinned as a named constant `_WORD_DTYPE = np.dtype("<u4")` and `_fill_int64_vectorised` builds the block-word view from it, matching the header and the scalar `struct("<8I")` path. This was the only numpy spot that materialises wire bytes — the `lanes.view(np.uint64)` hash arithmetic is value-based and therefore endian-independent.

On the regression check: on a little-endian host `np.dtype("<u4")` and `np.uint32` resolve to the *same* dtype, so no runtime assertion can distinguish them where CI runs. `test_int64_vectorised_block_words_pinned_little_endian` instead proves the fill routes through the constant: it forces `_WORD_DTYPE` to `">u4"` (what native order resolves to on s390x/ppc64) and asserts the header is byte-identical while every block word carries the same value in the opposite byte order. If the fill ever regresses to a hard-coded native dtype, overriding the constant changes nothing and the test fails on little-endian CI too. On an actual big-endian host there are three independent tripwires: the `newbyteorder("<")` assert, the vectorised-vs-scalar cross-check, and the Parquet/Arrow golden vectors.

No cost on little-endian hosts — the dtype resolves to native there, so the fast path is unchanged; `np.bitwise_or.at` handles the swapped view on big-endian hosts correctly (verified on numpy 2.0).

XuanYang-cn:
/lgtm
/approve
@xiaofan-luan Please mark this PR ready to revew if it's ready to merge, lgtm in myside.

xiaofan-luan:
@XuanYang-cn friendly ping for a re-review when you get a chance 🙏

All three rounds of your feedback are addressed:

- **Peak memory on the 10M-member path** — `769bcca2`: the intermediate `encoded` list is gone; the body is sized from `len(members)` and each value is validated / encoded / hashed in a single pass.
- **Patch coverage + untested XXH64 long-input branch** — `769bcca2`: golden vectors at 31/32/33 bytes (the sizes that actually reach the `length >= 32` stripe branch), plus FPR boundary cases. `50939f8d` additionally exercises the INT64 pure-Python fallback. codecov patch + project are both green now.
- **Host-native `np.uint32` byte order** — `7c09786c`: the wire view is pinned to `_WORD_DTYPE = np.dtype("<u4")`, so `_fill_int64_vectorised` matches the header and the scalar `struct("<8I")` path on big-endian hosts too.

CI is fully green (3.9 / 3.13 / 3.14 on ubuntu + windows, lint, proto check, DCO, codecov). The server side landed in milvus-io/milvus#51140, so this client PR is the last piece for `bloom_match`.

Your earlier `CHANGES_REQUESTED` is still blocking — could you take another look and dismiss it if the fixes look right? Happy to iterate on anything still off.

PR #3570, opened 2026-06-05
feat: route collection schema drops through AlterCollectionSchema

sijie-ni-0214:
issue: #3128
related: https://github.com/milvus-io/milvus/issues/48983

## What changed
- Keep the public drop field/function APIs while routing their implementations through AlterCollectionSchema DropRequest.
- Support function_name in AlterCollectionSchema drop requests and validate that exactly one drop identifier is provided.
- Keep schema cache invalidation on AlterCollectionSchema paths and cover sync/async handlers and clients.

## Tests
- python -m ruff check pymilvus/client/prepare.py pymilvus/client/grpc_handler.py pymilvus/client/async_grpc_handler.py pymilvus/milvus_client/milvus_client.py pymilvus/milvus_client/async_milvus_client.py tests/prepare/test_collection.py tests/grpc_handler/test_collection.py tests/async_grpc_handler/test_async_collection.py tests/test_milvus_client.py tests/test_async_milvus_client.py
- python -m pytest tests/prepare/test_collection.py::TestAlterCollectionSchemaRequest tests/grpc_handler/test_collection.py::TestGrpcHandlerCollectionProperties tests/async_grpc_handler/test_async_collection.py::TestAsyncGrpcHandlerCollectionProperties tests/test_milvus_client.py::TestMilvusClientCollectionMgmt tests/test_async_milvus_client.py::TestAsyncAlterCollectionSchema -q
- git diff --check

PR #3704, opened 2026-07-21
fix: align add field and drop function RPC routes

sijie-ni-0214:
## What changed

- Route synchronous and asynchronous `add_collection_field` calls through a field-only `AlterCollectionSchema` AddRequest.
- Fall back to the legacy `AddCollectionField` RPC only when an older server returns gRPC `UNIMPLEMENTED`.
- Invalidate the client schema cache after either add-field path succeeds.
- Restore synchronous and asynchronous `drop_collection_function` calls to the original `DropCollectionFunction` RPC.
- Keep `drop_function_field` on `AlterCollectionSchema` with output-field deletion enabled.
- Deprecate `add_collection_function` and `drop_collection_function` in favor of `add_function_field` and `drop_function_field`, because Milvus 3.0 no longer supports separate function attachment and detachment.

## Why

`add_collection_field` is moving to the unified schema-mutation RPC. Server-side support for external collections will be enabled on that path, so PyMilvus should use the same route on new servers.

Older Milvus servers do not expose `AlterCollectionSchema` but still support `AddCollectionField`. The compatibility path is therefore limited to gRPC `UNIMPLEMENTED`; permission failures, validation failures, and server status errors are never retried through the legacy RPC.

`drop_collection_function` has different semantics from `drop_function_field`: it represents function detachment while retaining the output field. The current server rejects both the legacy RPC and `AlterCollectionSchema` detachment, but routing through Alter additionally changes the established wire and privilege contract. Restoring the original RPC preserves the existing public API behavior without claiming to enable function detachment.

The supported destructive operation remains `drop_function_field`, which deletes the function together with its output field and index.

issue: #3128

Related Milvus issue: https://github.com/milvus-io/milvus/issues/48983

## Test plan

- [x] Sync and async collection handler unit tests: 71 passed
- [x] Ruff check and format verification for changed files
- [x] `git diff --check`
- [x] Local Milvus 2.6 E2E for sync and async clients: `AlterCollectionSchema` returned `UNIMPLEMENTED`, `AddCollectionField` was invoked once, the new field appeared in the schema, and inserts using the field succeeded

### DIFF
--- a/pymilvus/client/async_grpc_handler.py
+++ b/pymilvus/client/async_grpc_handler.py
@@ -1592,11 +1592,24 @@ class AsyncGrpcHandler:
         **kwargs,
     ):
         check_pass_param(collection_name=collection_name, timeout=timeout)
-        request = Prepare.add_collection_field_request(collection_name, field_schema)
-        status = await self._async_stub.AddCollectionField(
-            request, timeout=timeout, metadata=_api_level_md(context)
+        request = Prepare.alter_collection_schema_request(
+            collection_name=collection_name,
+            field_schema=field_schema,
         )
-        check_status(status)
+        try:
+            response = await self._async_stub.AlterCollectionSchema(
+                request, timeout=timeout, metadata=_api_level_md(context)
+            )
+            check_status(response.alter_status)
+        except grpc.RpcError as e:
+            if e.code() != grpc.StatusCode.UNIMPLEMENTED:
+                raise
+            legacy_request = Prepare.add_collection_field_request(collection_name, field_schema)
+            status = await self._async_stub.AddCollectionField(
+                legacy_request, timeout=timeout, metadata=_api_level_md(context)
+            )
+            check_status(status)
+        self._invalidate_schema(collection_name, db_name=(context.get_db_name() if context else ""))
 
     @retry_on_rpc_failure()
     async def add_collection_struct_field(
@@ -1670,14 +1683,13 @@ class AsyncGrpcHandler:
         context: Optional[CallContext] = None,
         **kwargs,
     ):
-        await self._alter_collection_schema_drop(
-            collection_name,
-            function_name=function_name,
-            timeout=timeout,
-            context=context,
-            **kwargs,
+        check_pass_param(collection_name=collection_name, timeout=timeout)
+        request = Prepare.drop_collection_function_request(collection_name, function_name)
+
+        status = await self._async_stub.DropCollectionFunction(
+            request, timeout=timeout, metadata=_api_level_md(context)
         )
-        self._invalidate_schema(collection_name, db_name=(context.get_db_name() if context else ""))
+        check_status(status)
 
     @retry_on_rpc_failure()
     async def add_collection_function(
--- a/pymilvus/client/grpc_handler.py
+++ b/pymilvus/client/grpc_handler.py
@@ -547,11 +547,24 @@ class GrpcHandler:
         **kwargs,
     ):
         check_pass_param(collection_name=collection_name, timeout=timeout)
-        request = Prepare.add_collection_field_request(collection_name, field_schema)
-        status = self._stub.AddCollectionField(
-            request, timeout=timeout, metadata=_api_level_md(context)
+        request = Prepare.alter_collection_schema_request(
+            collection_name=collection_name,
+            field_schema=field_schema,
         )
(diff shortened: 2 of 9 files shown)


pymilvus/client/grpc_handler.py
```diff
@@ -624,14 +628,13 @@ def drop_collection_function(
...
-            context=context,
-            **kwargs,
+        check_pass_param(collection_name=collection_name, timeout=timeout)
+        request = Prepare.drop_collection_function_request(collection_name, function_name)
+
+        status = self._stub.DropCollectionFunction(
+            request, timeout=timeout, metadata=_api_level_md(context)
         )
-        self._invalidate_schema(collection_name, db_name=(context.get_db_name() if context else ""))
+        check_status(status)
```
XuanYang-cn, line 646: Successful function detachment changes both function list and retained output field’s `is_function_output` flag, but this path no longer invalidates `GlobalCache.schema`. A later insert/upsert can keep treating that field as function-generated and reject valid input. Please call `_invalidate_schema(collection_name, db_name=...)` after `check_status(status)` here and in async handler, and keep cache assertions in both tests.

pymilvus/client/async_grpc_handler.py
```diff
@@ -1670,14 +1674,13 @@ async def drop_collection_function(
...
         **kwargs,
     ):
-        await self._alter_collection_schema_drop(
-            collection_name,
-            function_name=function_name,
-            timeout=timeout,
-            context=context,
-            **kwargs,
+        check_pass_param(collection_name=collection_name, timeout=timeout)
+        request = Prepare.drop_collection_function_request(collection_name, function_name)
```
XuanYang-cn, line 1687: Switching back to legacy request drops validation previously provided by `alter_collection_schema_request`: `function_name=""` used to raise `ParamError`, but now gets sent to server. Please preserve non-empty-name validation in both sync and async paths and add tests.

pymilvus/client/grpc_handler.py
```diff
@@ -624,14 +628,13 @@ def drop_collection_function(
...
         **kwargs,
     ):
-        self._alter_collection_schema_drop(
-            collection_name,
-            function_name=function_name,
-            timeout=timeout,
-            context=context,
-            **kwargs,
+        check_pass_param(collection_name=collection_name, timeout=timeout)
+        request = Prepare.drop_collection_function_request(collection_name, function_name)
```
XuanYang-cn, line 641: Non-blocking: this request → RPC → status flow is duplicated in sync and async handlers, same as new add-field route. Consider sharing request/result handling where practical to reduce parity bugs; okay to keep if handler mirroring is intentional project pattern.

XuanYang-cn (changes requested):
please address the following review comments

pymilvus/client/grpc_handler.py
sijie-ni-0214: Agreed that sync/async parity matters. These handlers intentionally mirror each other throughout the module, and this patch is restoring the existing legacy flow. We will keep the explicit implementations to keep the change surgical and avoid a one-off abstraction.

pymilvus/client/grpc_handler.py
sijie-ni-0214: Thanks for flagging this. This PR intentionally restores the pre-#3570 legacy `DropCollectionFunction` handler flow exactly; that implementation did not invalidate `GlobalCache.schema`. Function detachment is rejected by the current server and is not planned to be supported going forward, so there is no successful schema mutation on the supported path that could leave this cache stale. We will keep the legacy behavior unchanged here.

sijie-ni-0214:
Updated the commit body with `See also: #3128` as requested. I also replied inline on the two behavioral comments and the non-blocking abstraction suggestion. PTAL.

pymilvus/client/async_grpc_handler.py
sijie-ni-0214: Thanks. The legacy implementation also validated only `collection_name` and `timeout`; `Prepare.drop_collection_function_request` forwarded `function_name` as-is. Since the purpose is exact restoration of the old RPC contract and function detachment remains unsupported, we will not introduce a new SDK-side validation contract in this patch.

XuanYang-cn (approved):
/lgtm

Issue #3726, opened 2026-07-28
[Bug]: add_collection_field fails for external collections after schema RPC routing


sijie-ni-0214:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

After https://github.com/milvus-io/pymilvus/pull/3704, synchronous and asynchronous `add_collection_field` first use `AlterCollectionSchema` and fall back to the legacy `AddCollectionField` RPC only when the transport returns gRPC `UNIMPLEMENTED`.

Milvus currently rejects `AlterCollectionSchema` for an external collection with a business status instead:

```text
code: 1100
error_code: IllegalArgument
reason: alter collection schema operation is not supported for external collection <name>: invalid parameter
```

The legacy `AddCollectionField` RPC supports external collections and validates the required `external_field` mapping. PyMilvus therefore raises the business error without reaching the compatible RPC.

### Expected Behavior

When `AlterCollectionSchema` returns the explicit external-collection unsupported status, `add_collection_field` should fall back to `AddCollectionField` for both synchronous and asynchronous clients.

Other validation or permission errors must continue to propagate without fallback.

### Steps/Code To Reproduce behavior

```python
client.add_collection_field(
    collection_name=external_collection_name,
    field_name="score",
    data_type=DataType.DOUBLE,
    nullable=True,
    external_field="score",
)
```

1. Create an external collection with valid external field mappings.
2. Call `add_collection_field` with a valid `external_field` mapping.
3. Observe that `AlterCollectionSchema` returns code `1100` and PyMilvus raises the error.
4. Calling the legacy `AddCollectionField` RPC with the same field schema succeeds.

### Environment details

```markdown
- PyMilvus version: master after de103b83 / PR #3704
- Milvus version: master with external collection support
- SDK: PyMilvus master
```

### Anything else?

The protocol does not currently expose a dedicated error code for this compatibility case, so the client must narrowly match the parameter-invalid status and the external-collection unsupported reason.

Related schema routing PR: https://github.com/milvus-io/pymilvus/pull/3704
Related schema routing issue: https://github.com/milvus-io/pymilvus/issues/3705

XuanYang-cn:
should be fixed

PR #3727, opened 2026-07-28
fix: fallback add field for external collections

sijie-ni-0214:
## What changed

- Detect the explicit parameter-invalid status returned when `AlterCollectionSchema` is used on an external collection.
- Fall back to the legacy `AddCollectionField` RPC for that compatibility case in synchronous and asynchronous handlers.
- Preserve all unrelated business and permission errors without fallback.
- Cover the real wire status and verify that `external_field` metadata survives the legacy request.

## Why

After https://github.com/milvus-io/pymilvus/pull/3704, `add_collection_field` first uses the unified `AlterCollectionSchema` RPC. Milvus currently rejects that RPC for external collections with code `1100`, legacy `IllegalArgument`, and the reason `alter collection schema operation is not supported for external collection ...`.

The legacy `AddCollectionField` RPC supports external collections and validates their `external_field` mappings. Without this compatibility fallback, a valid public `add_collection_field` call fails before reaching the supported path.

issue: #3726

Related schema-routing PR: https://github.com/milvus-io/pymilvus/pull/3704

## Test plan

- [x] Sync and async collection handler unit tests: `74 passed`
- [x] Ruff checks for all changed files
- [x] Black format verification for all changed files
- [x] `git diff --check`
- [x] Local Milvus external-table L0 E2E: add scalar field, refresh, load, query, and filter (`1 passed`)
- [x] Local RPC spy verified `AlterCollectionSchema(code=1100, error_code=IllegalArgument)` followed by `AddCollectionField(success)`
- [x] Described schema retained `nullable=True` and `external_field="score"`

### DIFF
--- a/pymilvus/client/async_grpc_handler.py
+++ b/pymilvus/client/async_grpc_handler.py
@@ -69,6 +69,7 @@ from .utils import (
     check_status,
     get_server_type,
     immutable_message_to_dict,
+    is_external_collection_schema_alter_unsupported,
     is_successful,
     len_of,
     replicate_checkpoint_to_dict,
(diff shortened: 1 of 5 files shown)


XuanYang-cn (approved):
/lgtm

PR #3741, opened 2026-08-03
fix(cache): stop evicting collection timestamps under LRU pressure (#3721)

### LINKED ISSUE #3721: [Bug]: CollectionTsCache should not be LRU

### Describe the bug

The CollectionTsCache is a LRU cache capped at DEFAULT_CAPACITY=4096. Once the cap is exceeded. After eviction CollectionTsCache::get returs 0 and construct_guarantee_ts (ts_utils.py) falls back to guarantee_ts = 1, the weakest bound, so a SESSION-consistency read of a collection that was written but then evicted silently degrades to eventually-consistent with no error or signal. This is reachable under many-collection multi-tenant load. 

### Expected Behavior

_No response_

```markdown

```

### Environment details

```markdown
```

### Anything else?

_No response_

Anai-Guo:
Fixes #3721.

### What happens today

`CollectionTsCache` inherits `CacheRegion`, whose store is an `LRUCache(maxsize=DEFAULT_CAPACITY)` with `DEFAULT_CAPACITY = 4096`. The keys are `(endpoint, db_name, collection_name)`, so a process that touches more than 4096 collections starts evicting entries.

An evicted timestamp is indistinguishable from "never written":

```python
# pymilvus/client/cache.py
def get(self, endpoint, db_name, collection_name) -> int:
    return super().get(key) or 0          # evicted -> 0

# pymilvus/client/ts_utils.py — construct_guarantee_ts
kwargs[GUARANTEE_TIMESTAMP] = (
    get_collection_ts(collection_name, endpoint, db_name) or get_eventually_ts()
)                                          # 0 -> EVENTUALLY_TS (1)
```

So a `Session`-consistency read (and the default path, which uses the same fallback) of a collection that *was* written but has since been evicted quietly drops to the weakest bound. Nothing is raised and nothing is logged — the caller just gets stale data. It is reachable under exactly the workload the cache exists for: many collections, multi-tenant.

### The change

- `CacheRegion.__init__` now accepts `capacity: Optional[int]`; `None` selects a plain `dict` instead of an `LRUCache`, i.e. entries are only removed by `invalidate()` / `clear()`.
- `CollectionTsCache` defaults to `capacity=None`.
- `SchemaCache` and the `CacheRegion` default are untouched and stay LRU-bounded: an evicted schema costs a `describe_collection` round trip, it does not change read semantics.

On the memory question: a `CollectionTsCache` entry is one small int keyed by a 3-tuple of strings, versus a full schema dict in `SchemaCache`, and `invalidate()` still drops entries. Trading that for "a Session read never silently degrades" is the direction the issue asks for.

### Tests

Two tests in `tests/unit/test_collection_ts_cache.py`:

- `test_cache_is_not_bounded_by_the_default_capacity` — writes `DEFAULT_CAPACITY + 10` collections, then asserts the *first* one is still remembered and that `construct_guarantee_ts` still produces its real timestamp instead of `EVENTUALLY_TS`. On `master` this fails with `AssertionError: 4096 != 4106`.
- `test_invalidate_still_drops_an_entry` — the unbounded region still honours explicit invalidation.

Verified locally against the full unit suite (`pytest tests/unit`, excluding the modules that need `pymilvus[bulk_writer]`/network in my environment):

| | passed | failed |
|---|---|---|
| `master` | 4231 | 3 |
| this branch | 4233 | 3 |

The same 3 failures occur on `master` before the change (`orm/test_iterator.py`, `test_check.py::test_get_commit`, `test_iterator_ownership.py`) and are unrelated to this patch. `black --line-length 100 --check` and `ruff check` (0.12.9 and 0.15.12, repo `pyproject.toml`) report no new findings on either file.

### DIFF
--- a/pymilvus/client/cache.py
+++ b/pymilvus/client/cache.py
@@ -45,6 +45,18 @@ class CacheRegion:
         with self._lock:
             self._cache.pop(key, None)
 
+    def invalidate_prefix(self, prefix: Tuple[Any, ...]) -> None:
+        """Remove every entry whose tuple key starts with ``prefix``.
+
+        Used to drop a whole database or endpoint at once, so an unbounded
+        region still shrinks when the objects it describes are gone.
+        """
+        size = len(prefix)
+        with self._lock:
+            stale = [k for k in self._cache if isinstance(k, tuple) and k[:size] == prefix]
+            for key in stale:
+                self._cache.pop(key, None)
+
     def clear(self) -> None:
         """Clear all entries from cache."""
         with self._lock:
@@ -81,11 +93,11 @@ class SchemaCache(CacheRegion):
 
     def invalidate_db(self, endpoint: str, db_name: str) -> None:
         """Invalidate all schemas for a database."""
-        prefix = (endpoint, db_name or "default")
-        with self._lock:
-            keys_to_remove = [k for k in self._cache if k[:2] == prefix]
-            for key in keys_to_remove:
-                self._cache.pop(key, None)
+        self.invalidate_prefix((endpoint, db_name or "default"))
+
+    def invalidate_endpoint(self, endpoint: str) -> None:
+        """Invalidate all schemas cached for an endpoint."""
+        self.invalidate_prefix((endpoint,))
 
     @staticmethod
     def _make_key(endpoint: str, db_name: str, collection_name: str) -> Tuple[str, str, str]:
@@ -129,6 +141,14 @@ class CollectionTsCache(CacheRegion):
         """Invalidate timestamp for a specific collection."""
         key = self._make_key(endpoint, db_name, collection_name)
         super().invalidate(key)
+
+    def invalidate_db(self, endpoint: str, db_name: str) -> None:
+        """Invalidate all timestamps for a database."""
+        self.invalidate_prefix((endpoint, db_name or "default"))
+
+    def invalidate_endpoint(self, endpoint: str) -> None:
+        """Invalidate all timestamps cached for an endpoint."""
+        self.invalidate_prefix((endpoint,))
 
     @staticmethod
     def _make_key(endpoint: str, db_name: str, collection_name: str) -> Tuple[str, str, str]:
--- a/pymilvus/client/cache.py
+++ b/pymilvus/client/cache.py
@@ -1,6 +1,6 @@
 import logging
 import threading
-from typing import Any, ClassVar, Optional, Tuple
+from typing import Any, ClassVar, MutableMapping, Optional, Tuple
 
 from cachetools import LRUCache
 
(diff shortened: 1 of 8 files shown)


pymilvus/client/cache.py
```diff
@@ -90,8 +100,17 @@ class CollectionTsCache(CacheRegion):
 
     Key: (endpoint, db_name, collection_name)
     Value: timestamp (int)
+
+    Unbounded on purpose. A missing entry makes ``get`` return 0, and
+    ``construct_guarantee_ts`` then falls back to ``EVENTUALLY_TS``, so an
+    evicted collection silently downgrades a Session-consistency read to the
+    weakest guarantee with nothing logged. An entry is one small int, and
+    stale ones are dropped by ``invalidate``.
```
yhmo, line 120: Medium ---- Making this region unbounded fixes the consistency downgrade, but no production path currently calls `CollectionTsCache.invalidate` or `clear`: sync/async `drop_collection` only invalidate the schema cache, `drop_database` does neither, and connection teardown leaves the global entries intact. A process that repeatedly creates/writes/drops unique collections or endpoints will therefore retain every tuple forever, and recreating a collection keeps the prior incarnation’s timestamp. Please wire collection/db/endpoint cleanup into the corresponding lifecycle paths and add drop/recreate plus churn regression coverage.

Anai-Guo:
Thanks @yhmo — you're right, and the leak is worse than "no caller": `_invalidate_db_schemas` is defined in **both** handlers and never called from anywhere, so `drop_database` was already leaving stale *schemas* behind before this PR. Pushed `f167de8`.

**Wired into the lifecycle paths**

- `CacheRegion.invalidate_prefix` — `SchemaCache.invalidate_db` now reuses it instead of open-coding the scan, and both regions get `invalidate_db` / `invalidate_endpoint`.
- `drop_collection` (sync + async) also drops the collection's timestamp. This fixes the recreate case you called out: `set()` only moves a timestamp forward, so a leftover entry pinned the new incarnation to the old one's value.
- `drop_database` (sync + async) drops that database's schemas *and* timestamps.

**Deliberately not wired: `close()`**

`GlobalCache` is process-wide and keyed by endpoint, so two handlers pointed at the same address share entries by design. Purging on `close()` would strip a live handler's timestamps when an unrelated one disconnects — which is exactly the silent Session-to-Eventually downgrade #3721 is about. Doing it safely needs refcounting live handlers per endpoint and purging on the last close. `invalidate_endpoint` is there and tested so that's a small follow-up; happy to add it here if you'd rather it land together, but it seemed worth keeping separate from the correctness fix.

The remaining unreclaimed case is therefore endpoint churn, not collection churn — one small int per endpoint ever connected to.

**Coverage**

Drop/recreate, per-database and per-endpoint invalidation, and `DEFAULT_CAPACITY + 10` rounds of both collection churn and database churn asserting `len(cache) == 0`. Plus a guard that schema-only invalidation (`drop_collection_field`) *keeps* the timestamp — the collection still exists there, so dropping it would reintroduce the downgrade.

Verified the seven new tests fail on the production code without this commit. Full unit suite: 4287 passed; the two failures in `test_iterator.py` / `test_check.py` reproduce on the base commit untouched. `ruff check` and `ruff format --check` clean on all eight files.

Issue #3742, opened 2026-08-04
[Enhancement]: Add query_iterator to AsyncMilvusClient


csy1204:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What would you like to be added?

`MilvusClient.query_iterator()` has no counterpart on `AsyncMilvusClient`. Please add `AsyncMilvusClient.query_iterator()`, returning an async iterator with `next()` / `close()` / `get_cursor()` and matching the sync signature.

### Why is this needed?

Async users who need to walk a whole filtered result set currently have to fall back to `query(..., limit=N)`. `query` has no ordering guarantee, so *which* N rows come back is undefined — it depends on segment scan order and compaction. The only correct alternative today is hand-rolling primary-key cursor pagination in application code, including the mvcc-timestamp pinning that the sync iterator already does.

### Anything else?

A previous attempt (#3483) bundled this with six unrelated changes and was closed before review. That patch also passed the async handler into the *sync* `QueryIterator`, whose `next()` calls `handler.query(...)` synchronously — it would receive a coroutine instead of rows. A working implementation needs an async driver.

I have a focused patch ready and will open a PR referencing this issue.

PR #3483, opened 2026-05-12
fix: add param checks, flush progress logs, and async iterators

Adil2009700:
## Summary

This PR addresses seven open issues with code-only changes:

### 1. Fix #1939 - Add missing param checks
- Added check_pass_param calls to insert_rows, upsert, search, and create_index in both GrpcHandler and AsyncGrpcHandler.
- Validates collection_name, partition_name, field_name, and index_name before RPC calls.

### 2. Fix #747 - Add progress log while waiting for flush
- Added logging every 60 seconds to _wait_for_flushed, _wait_for_flush_all, and async _wait_for_flushed.
- Log message: \Waiting for flush of collection {name} done after {X} seconds.\
- Improves visibility when flush operations take a long time.

### 3. Fix #2412 - Add iterator methods to AsyncMilvusClient
- Added query_iterator() method returning QueryIterator (with get_cursor() support).
- Added search_iterator() method returning SearchIterator.
- Matches the sync MilvusClient API for feature parity.

### 4. Fix #3169 - Raise error when compact returns invalid compactionID
- Added compactionID validation in GrpcHandler.compact(): raise MilvusException if compactionID <= 0.
- Added compactionID validation in AsyncGrpcHandler.compact(): same check.
- Added tests: test_compact_raises_on_negative_compaction_id and test_compact_raises_on_zero_compaction_id.

### 5. Fix #2827 - Prevent AssertionError crash when no .env file exists
- load_dotenv() internally calls find_dotenv() which traverses call frames and can raise AssertionError.
- Use find_dotenv(usecwd=True) to avoid frame traversal, and only call load_dotenv if a .env file is found.

### 6. Fix #2589 - Use correct param name 'partition_names' in validation error message
- Added 'partition_names' to the validation check dict (mapping to same validator).
- Updated check_pass_param calls in prepare.py to use the correct user-facing parameter name.
- Error messages now report \partition_names\ instead of \partition_name_array\.

### 7. Fix #2538 - Serialize DataType/FunctionType enums as ints in to_dict()
- FieldSchema.to_dict() and Function.to_dict() now return integer values for type fields.
- Added to_dict() and from_dict() to DataType and FunctionType enum classes.
- to_dict() output is now fully JSON-serializable.

## Testing
- All existing tests pass: 294 misc tests, 239 schema tests, 175+ async client tests, 50+ grpc_handler tests, 143 partition tests.
- Linting (black + ruff) passes.

## Files Changed
- pymilvus/client/grpc_handler.py
[...]

PR #3743, opened 2026-08-04
enhance: add query_iterator to AsyncMilvusClient

csy1204:
related to: #3742

## Problem

`AsyncMilvusClient` has no `query_iterator()`. Async callers who need a full filtered result set fall back to `query(..., limit=N)` — which has no ordering guarantee, so which N rows come back is undefined — or hand-roll primary-key cursor pagination plus mvcc-timestamp pinning in application code.

## What this does

The sync `QueryIterator` touches the network in only four places, all through `self._handler`; everything else is pure logic. So instead of a parallel implementation, this splits the state machine out and drives it from both transports.

- **`refactor: extract transport-agnostic _QueryIteratorBase from QueryIterator`** — the base holds cursor bookkeeping, expression assembly, batch/limit accounting, the result cache and the checkpoint file, and exposes `_*_kwargs()` builders paired with `_consume_*()` methods. `QueryIterator` becomes a thin sync driver over it. Two private helpers move to a single underscore (`_setup_next_expr`, `_save_pk_cursor`) because the subclass driver calls them and existing tests reference the mangled names.
- **`enhance: add AsyncQueryIterator on top of the shared iterator base`** — the async driver. Setup RPCs run in a `create()` classmethod because `__init__` cannot await. `await next()`, `await close()`, `get_cursor()`, and `__aiter__`/`__anext__` for `async for`.
- **`enhance: add query_iterator to AsyncMilvusClient`** — the factory, mirroring the sync one (same parameter names/order/defaults, same `filter` type check, same schema-cache lookup). `AsyncMilvusClientSession` delegates it the way the sync session does.
- **`test: cover AsyncQueryIterator limits, cursors, checkpoint resume and async for`**

```python
it = await client.query_iterator(collection_name="c", filter="pk > 0", batch_size=1000)
try:
    while True:
        batch = await it.next()
        if not batch:
            break
        handle(batch)
finally:
    await it.close()

# or
async for batch in it:
    handle(batch)
```

## Behavior note

The refactor moves parameter validation ahead of the first RPC, so `QueryIterator(batch_size=-1)` now raises `ParamError` without the `describe_collection` call the old code made first (same for a schema dict with no `is_primary` field). This is the only intentional behavior difference, and it makes sync and async validate identically. It is stated in the refactor commit's message too.

## How the refactor was verified

Commit 1 renames the class at the top of the file and appends a subclass, so GitHub renders it as close to a rewrite — "existing tests pass" is weak evidence for a refactor whose tests were written against the old structure. So I also compared old vs new directly with a throwaway differential harness (not part of this PR), driving both versions against identical recording handlers and diffing the full ordered sequence of `describe_collection`/`query` calls with all kwargs, plus returned batches, final cursor and checkpoint-file contents:

- **19 scenarios, old vs new: identical except the validation-ordering difference above.** Covered int/varchar/None pk expressions, the cache hit-and-refill path, `limit` cut-off, offset seek (full / drained-early / partial-loop), `element_filter` with and without an element cursor, extra `rpc_options` passthrough, and all six checkpoint branches (fresh, resume, one-line `ParamError`, unparseable ts, element-cursor resume, resume-with-residual-offset).
- **24 scenarios, new sync vs new async: identical, zero divergence.** Also covered `limit=0`, `reduce_stop_for_best=False`, the mvccTs fallback when the server returns 0, and the 100-line checkpoint-file truncation path.

## Tests

`tests/unit/test_async_query_iterator.py` (new, no server needed) covers batch exhaustion, int and varchar PK cursor expressions, `batch_size` validation before any RPC, `limit` cut-off, checkpoint-file write / resume / removal, `describe_collection` called exactly once, `async for` parity with the `next()` loop, and `get_cursor()`. `tests/unit/test_iterator_ownership.py` gains an ownership assertion for the new class, and the session-delegation parametrize list in `tests/unit/test_async_milvus_client.py` gains a `query_iterator` row so `cluster_id` forwarding is pinned the way the sync suite pins it.

Full unit suite: `4553 passed, 3 skipped`. `black --check` and `ruff check` clean.

Also ran it against a live Milvus deployment: iterating one filtered set returned 4,621 rows over 47 pages, and `AsyncMilvusClient.query_iterator`, `async for`, and the sync `query_iterator` all produced the identical primary-key set with no duplicates.

## Scope

Deliberately limited to `query_iterator`, since #3483 was closed for bundling unrelated changes:

- `search_iterator` async support is left for a separate PR, which is why `AsyncMilvusClientSession` delegates only `query_iterator` where the sync session delegates both.
- `__aiter__`/`__anext__` has no sync counterpart on purpose — it is the idiomatic async consumption protocol and six lines. I did not add `__iter__` to the sync iterator.
- Checkpoint-file I/O stays synchronous inside the async coroutines. It lives in the shared base, it is one buffered line plus a flush per batch next to a network round-trip, and moving it off-thread would mean either a thread-pool hop per batch or lifting the checkpoint logic back out of the shared base.
- `AsyncQueryIterator` is not exported from the top-level `pymilvus` namespace, matching `QueryIterator`.

Happy to split, rename or adjust anything here — including the base-class extraction, if you would rather see the async driver built a different way.

### DIFF
--- a/pymilvus/client/iterator/async_query_iterator.py
+++ b/pymilvus/client/iterator/async_query_iterator.py
@@ -1,0 +1,133 @@
+from __future__ import annotations
+
+import logging
+import time
+from typing import TYPE_CHECKING, Any, Mapping, Protocol
+
+from pymilvus.client.constants import MAX_BATCH_SIZE
+
+from .query_iterator import _QueryIteratorBase
+
+if TYPE_CHECKING:
+    from pymilvus.client.call_context import CallContext
+
+log = logging.getLogger(__name__)
+
+
+class AsyncQueryIteratorHandler(Protocol):
+    async def describe_collection(
+        self, collection_name: str, **kwargs: Any
+    ) -> Mapping[str, Any]: ...
+
+    async def query(self, collection_name: str, **kwargs: Any) -> Any: ...
+
+
+class AsyncQueryIterator(_QueryIteratorBase):
+    """Non-blocking counterpart of :class:`QueryIterator`.
+
+    ``__init__`` cannot await, so setup RPCs (describe collection, mvccTs probe and the
+    optional offset seek) run in :meth:`create`. Build instances through that classmethod,
+    or through :meth:`AsyncMilvusClient.query_iterator`, never by calling the constructor
+    directly.
+
+    A single iterator must not be advanced concurrently: calling ``await it.next()``
+    from two tasks at once corrupts the primary-key cursor, because the state mutation
+    straddles the await. (The sync iterator has the same hazard across threads.)
+
+    When a checkpoint file is configured (the ``iterator_cp_file`` option), every batch
+    write goes through :meth:`_finalize_batch`, which writes and flushes one cursor line
+    to that file synchronously, on the event-loop thread; the same is true of the file
+    open/read done by :meth:`create` and the close/unlink done by :meth:`close`. A slow
+    or network-mounted checkpoint path therefore blocks the loop and stalls unrelated
+    coroutines. Leave checkpointing off, or point it at local disk, on latency-sensitive
+    loops.
+
+    Examples:
+        >>> it = await client.query_iterator(collection_name="c", filter="pk > 0")
+        >>> try:
+        ...     async for batch in it:
+        ...         handle(batch)
+        ... finally:
+        ...     await it.close()
+    """
+
+    @classmethod
+    async def create(
+        cls,
+        handler: AsyncQueryIteratorHandler,
+        context: CallContext | None,
+        collection_name: str,
+        batch_size: int | None = 1000,
+        limit: int | None = -1,
+        expr: str | None = None,
+        output_fields: list[str] | None = None,
+        partition_names: list[str] | None = None,
+        schema: Mapping[str, Any] | None = None,
+        timeout: float | None = None,
+        rpc_options: Mapping[str, Any] | None = None,
+    ) -> AsyncQueryIterator:
+        iterator = cls(
+            handler=handler,
+            context=context,
+            collection_name=collection_name,
+            batch_size=batch_size,
+            limit=limit,
+            expr=expr,
+            output_fields=output_fields,
+            partition_names=partition_names,
+            schema=schema,
+            timeout=timeout,
+            rpc_options=rpc_options,
+        )
+        await iterator._setup()
+        return iterator
+
+    async def _setup(self) -> None:
+        self._apply_collection_id(
+            await self._handler.describe_collection(
+                self._collection_name, **self._describe_kwargs()
+            )
+        )
+        if self._prepare_ts_cp():
+            self._consume_ts_response(await self._handler.query(**self._ts_query_kwargs()))
+            self._save_mvcc_ts_if_needed()
+        await self._seek_to_offset()
+
+    async def _seek_to_offset(self) -> None:
+        offset = self._seek_offset_start()
+        if offset <= 0:
+            return
+        start_time = time.time()
+        while offset > 0:
+            batch_size = min(MAX_BATCH_SIZE, offset)
+            query_kwargs = self._seek_query_kwargs(batch_size)
+            seeked_count = self._consume_seek_batch(await self._handler.query(**query_kwargs))
+            log.debug(
+                f"seeked offset, seek_expr:{query_kwargs['expr']} "
+                f"batch_size:{batch_size} seeked_count:{seeked_count}"
+            )
+            if seeked_count == 0:
+                log.info("seek offset has drained all matched results for query iterator, break")
+                break
+            offset -= seeked_count
+        self._finish_seek(offset, start_time)
+
+    async def next(self) -> list:
+        ret = self._take_from_cache()
+        if ret is None:
+            ret = self._consume_next_response(
+                await self._handler.query(**self._next_query_kwargs())
+            )
+        return self._finalize_batch(ret)
+
+    async def close(self) -> None:
+        self._close_common()
+
+    def __aiter__(self) -> AsyncQueryIterator:
+        return self
+
+    async def __anext__(self) -> list:
+        batch = await self.next()
+        if not batch:
+            raise StopAsyncIteration
+        return batch
--- a/pymilvus/client/iterator/query_iterator.py
+++ b/pymilvus/client/iterator/query_iterator.py
@@ -82,10 +82,18 @@ def io_operation(io_func: Callable[[Any], None], message: str):
         raise MilvusException(message=message) from ose
 
 
-class QueryIterator:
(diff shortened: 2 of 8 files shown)


csy1204:
/assign @XuanYang-CN

For context: in #3483 you asked for that PR to be closed and split into one PR per requirement. This is the async-iterator item, on its own, with focused tests — no other changes bundled in.

pymilvus/client/iterator/query_iterator.py
```diff
@@ -121,67 +128,169 @@ def __init__(
...
+        self._need_save_cp = True
+        self._cp_file_path = Path(self._cp_file_path_str)
+        if not self.__init_cp_file_handler():
+            # input cp file is empty, set up mvccTs by query request
+            return True
+        try:
+            # input cp file is not emtpy, init mvccTs by reading cp file
+            lines = self._cp_file_handler.readlines()
+            line_count = len(lines)
+            if line_count < 2:
```
yhmo, line 188: pymilvus/client/iterator/query_iterator.py line:188
Medium ---- An already-existing empty checkpoint (including one left by a timed-out or cancelled setup after line 181 opens it) takes the r+ path and is rejected here as malformed. The documented empty-file initialization path only covers a nonexistent file opened with w, so retrying with the same checkpoint path cannot recover. Please treat a zero-line file as fresh and add a retry regression test before requiring the two-line resume format.

pymilvus/client/iterator/query_iterator.py
```diff
@@ -121,67 +128,169 @@ def __init__(
...
+            **self.__setup_query_params(),
+        }
+
+    def _consume_next_response(self, res: Any) -> list:
+        self.__maybe_cache(res)
+        return res[0 : min(self._query_options[BATCH_SIZE], len(res))]
+
+    def _finalize_batch(self, ret: list) -> list:
+        ret = self.__check_reached_limit(ret)
+        self.__update_cursor(ret)
```
yhmo, line 277: pymilvus/client/iterator/query_iterator.py line:277
Medium ---- This advances _next_id before the checkpoint write at line 278. If writelines, flush, or truncate raises, next() returns an exception without delivering ret, but a retry uses the advanced in-memory cursor and skips that entire batch. Please preserve or roll back the old cursor on checkpoint failure, or persist the candidate cursor before committing it, and cover the write-failure retry path.

pymilvus/client/iterator/query_iterator.py
csy1204: Thanks — this is a real bug, and I confirmed it before answering: it reproduces unchanged on plain `master`, so this PR moves that code but does not alter it.

Same script against `master` (c5e73a6e) and this branch, with a zero-byte checkpoint file present:

```
master : ParamError: input cp file:/tmp/.../cursor.cp should contain at least two lines, but only:0 lines
this PR: ParamError: input cp file:/tmp/.../cursor.cp should contain at least two lines, but only:0 lines
```

`__init_cp_file_handler()` returns `mode == "r+"` on file existence alone, identical to master. One thing I should concede, though: `AsyncQueryIterator` makes that state *easier* to reach. In sync, only a raising mvccTs RPC left an empty file behind; with the async driver a cancelled task between opening the file and writing the timestamp does the same.

So I filed it separately — #3744, with a runnable repro — and fixed it in #3746, which also accepts a one-line file (the state right after `__save_mvcc_ts()`, which the unreachable `if line_count > 1:` guard suggests was always intended).

pymilvus/client/iterator/query_iterator.py
csy1204: Also a real bug, and also pre-existing: `__update_cursor(ret)` before `io_operation(self.__save_pk_cursor, ...)` is master's exact order in `next()`.

```
master : MilvusException: failed to save pk cursor | _next_id None -> 2, returned_count=0
this PR: MilvusException: failed to save pk cursor | _next_id None -> 2, returned_count=0
```

Note `returned_count=0` — nothing was delivered, yet the cursor already sits past the batch.

Fixed in #3746 (issue #3744). While writing that fix I found the rollback needs to cover more than the cursor: `next()` also mutates `iterator_cache` before the write, so with `batch_size=2` and a 7-row result the retry after a failed write returned rows `[4,5]` instead of the `[2,3]` it never delivered — rows lost, no RPC involved. #3746 makes the checkpoint write the commit point, so the cursor *and* the cache stay untouched on failure and the retry re-delivers the identical batch.

I kept both out of this PR because its reviewability rests on the refactor being behavior-preserving, and because #3483 was closed for bundling unrelated changes. Happy to be told otherwise: if you would rather see them here, say so and I will fold #3746's commits in. The two branches conflict textually, so whichever you take first, I will rebase the other onto it.

pymilvus/client/iterator/async_query_iterator.py
```diff
@@ -0,0 +1,125 @@
...
+            offset -= seeked_count
+        self._finish_seek(offset, start_time)
+
+    async def next(self) -> list:
+        ret = self._take_from_cache()
+        if ret is None:
+            ret = self._consume_next_response(
+                await self._handler.query(**self._next_query_kwargs())
+            )
+        return self._finalize_batch(ret)
```
yhmo, line 121: pymilvus/client/iterator/async_query_iterator.py line:113
Low ---- `_finalize_batch()` can synchronously write and flush the checkpoint file on every page, so a slow or network-mounted checkpoint path blocks the event loop and stalls unrelated coroutines despite this class's non-blocking contract. Please offload checkpoint I/O (including setup/close) from the event-loop thread, or at minimum qualify the API documentation so callers can avoid checkpointing on latency-sensitive loops.

csy1204: Fair point, and I took the documentation option rather than the offload — pushed as an amend to the two commits that own those docstrings.

`AsyncQueryIterator`'s class docstring and `AsyncMilvusClient.query_iterator`'s `Note:` section now both say that with `iterator_cp_file` configured, one cursor line is written **and flushed synchronously on the event-loop thread per batch**, that the same applies to the checkpoint open/read in `create()` and the close/unlink in `close()`, and that callers on latency-sensitive loops should leave checkpointing off or point it at local disk.

I chose documentation over a thread hop deliberately, and I'd rather you overrule me than have me guess: the checkpoint logic lives in the state machine that the sync and async drivers share, so offloading it means either a thread-pool round trip per batch or lifting that logic back out of the base — and the latter undoes the reason this PR is small. If you'd prefer the offload anyway (an `asyncio.to_thread` hop in the async driver only, leaving sync untouched, would be the contained version), say so and I'll add it.

Separately, #3746 makes the checkpoint write itself rollbackable, which is the other half of trusting that file.

yhmo: The issue is confirmed fixed.

Issue #3744, opened 2026-08-04
[Bug]: QueryIterator cannot resume from a checkpoint file it left behind, and a failed checkpoint write skips a batch


csy1204:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

```markdown
- PyMilvus version: master (c5e73a6e)
```

### Current Behavior

`QueryIterator`'s checkpoint file (`iterator_cp_file`) cannot be resumed from two states that the iterator itself can leave behind. Both were found while reviewing #3743 and both reproduce on plain `master`.

**1. An existing but empty checkpoint file is rejected as malformed.**

`__init_cp_file_handler()` picks its mode from file *existence* alone, so a zero-byte file opens with `r+` and takes the resume path, where the `line_count < 2` check rejects it. A previous run that was interrupted after creating the file but before writing the mvcc timestamp (a raising `query` RPC, a cancelled task, a killed process) leaves exactly that file — so retrying with the same checkpoint path can never recover, which defeats the point of a checkpoint.

The same check also rejects a one-line file, which is the legitimate state after `__save_mvcc_ts()` wrote the timestamp but no batch has been consumed yet. The `if line_count > 1:` guard right below the raise is unreachable today, which suggests one-line files were meant to be accepted.

**2. The pk cursor advances before the checkpoint write, so a failed write skips a batch.**

`next()` calls `__update_cursor(ret)` and only then `io_operation(self.__save_pk_cursor, ...)`. If the write raises, `next()` propagates the exception without ever delivering `ret`, but the in-memory cursor has already moved past that batch. A caller that catches and retries gets the *next* batch, silently losing the rows it never received.

### Expected Behavior

1. A checkpoint file with no usable session timestamp should be initialized like a fresh one instead of raising, and a one-line file should resume from its timestamp with no pk cursor.
2. A failed checkpoint write should leave the cursor where it was, so a retry re-delivers the batch that was never returned.

### Steps To Reproduce

```python
import tempfile
from pathlib import Path

from pymilvus.client.constants import (
    COLLECTION_ID, ITERATOR_SESSION_CP_FILE, ITERATOR_SESSION_TS_FIELD,
)
from pymilvus.client.iterator import QueryIterator
from pymilvus.client.types import DataType

SCHEMA = {"fields": [{"name": "pk", "type": DataType.INT64, "is_primary": True}]}


class Res(list):
    def __init__(self, rows, ts=100):
        super().__init__(rows)
[...]

PR #3746, opened 2026-08-04
fix: make QueryIterator checkpoint recovery resumable and crash-safe

csy1204:
related to: #3744

Two `QueryIterator` checkpoint (`iterator_cp_file`) defects, both reproducible on plain `master` — repro scripts and output in #3744. They surfaced as review comments on #3743, which refactors this file without changing either behavior, so they are fixed here instead to keep that PR behavior-preserving.

## 1. A checkpoint file the iterator left behind cannot be resumed

`__init_cp_file_handler()` picks its mode from file *existence* alone, so a zero-byte file opens `r+`, takes the resume path in `__set_up_ts_cp()`, and is rejected by `if line_count < 2: raise ParamError(...)`. A run interrupted after the file was created but before the mvcc timestamp was written — a raising `query` RPC, a cancelled task, a killed process — leaves exactly that file, so retrying with the same checkpoint path can never recover. That defeats the point of having a checkpoint.

The same check also rejected a **one-line** file, which is the legitimate state after `__save_mvcc_ts()` wrote the timestamp but before any batch was consumed. The `if line_count > 1:` guard directly below the raise is unreachable today, which suggests one-line files were always meant to be accepted.

Now handled explicitly: 0 lines initializes like a brand-new file (fetch a session ts and persist it), 1 line resumes that timestamp with no pk cursor, 2+ lines behaves exactly as before. An unparseable first line still raises `ParamError`, and a failing `readlines()` still raises `MilvusException`.

## 2. A failed checkpoint write loses the batch it never delivered

`next()` advanced the pk cursor and mutated the result cache *before* persisting the checkpoint. When the write raised, `next()` propagated without ever returning `ret`, but the cursor had already moved past that batch and the cache had already been spliced — so a caller that caught the error and retried silently lost the rows it never received:

```
batch_size=2, one query returns 7 rows (the 5-row overflow is cached)
call 1 -> [0, 1]
call 2 -> served from cache as [2, 3], checkpoint write fails
call 3 (retry) -> [4, 5]        # rows 2 and 3 are gone for good
```

`next()` now treats the checkpoint write as its commit point: it decides the batch without touching `iterator_cache`, updates the cursor, attempts the write, and only on success commits the cache changes and the returned count. On failure the cursor rolls back and the cache is untouched, so a retry re-delivers the identical batch. Same scenario after the fix:

```
call 3 (retry) -> [2, 3]        # no loss, no extra RPC
```

The cache-miss branch's existing `release_cache()` stays where it is — that remainder is a stale suffix of what the server returns for the current cursor, so a re-query after the rollback re-fetches those rows.

## Tests

`tests/unit/orm/test_iterator.py`, +3 net: zero-byte and one-line checkpoint resume, an unparseable one-line file still raising, checkpoint-write failure re-delivering the same batch on retry from both the cache-hit path and the fresh-query-overflow path, and `_next_element_offset` rollback for an element-filter iterator. Write failures are simulated with a file-handle stub whose `writelines` raises `OSError` rather than by patching the method under test.

`test_cp_file_too_few_lines_raises` is removed: its assertion is the one-line rejection this PR fixes. `test_cp_file_single_line_resumes_ts_only` and `test_cp_file_single_line_invalid_ts_raises` replace its coverage. The existing two-line resume and element-cursor tests are untouched.

Full unit suite: `4542 passed, 3 skipped`. `black --check` and `ruff check` clean.

## Note on #3743

This branch is cut from `master` and is independent of #3743, but both touch `next()` and `__set_up_ts_cp()`, so they will conflict textually. Whichever you take first, I will rebase the other onto it.

### DIFF
--- a/pymilvus/client/iterator/query_iterator.py
+++ b/pymilvus/client/iterator/query_iterator.py
@@ -307,6 +307,39 @@ class QueryIterator:
             self._session_ts = fall_back_to_latest_session_ts()
         self._query_options[GUARANTEE_TIMESTAMP] = self._session_ts
 
+    def __init_fresh_cp(self):
+        # (re-)initialize as if the cp file had no content at all yet:
+        # set up mvccTs by a query request and persist it.
+        self.__setup_ts_by_request()
+        io_operation(self.__save_mvcc_ts, "Failed to save mvcc ts")
+
+    def __read_intact_cp_lines(self) -> list[str]:
+        # Every line this iterator writes to the cp file is newline-
+        # terminated: __save_mvcc_ts() writes str(self._session_ts) + "\n",
+        # and __save_pk_cursor() writes __dump_cursor_line() + "\n" -- and
+        # nothing else ever appends to this file. Both writers hand a
+        # single already-terminated "<content>\n" string to writelines()
+        # in one call, so a write interrupted partway through can only
+        # have landed a strict byte-prefix of that string; its trailing
+        # "\n" is its last byte, so a partial write can never include it.
+        # readlines() preserves line terminators, so a final line missing
+        # one is a definitive, content-independent signal that it was only
+        # partially written -- whatever shape the partial content takes
+        # (a bare "-1" left behind by an intended "-1234", a truncated
+        # JSON object, half a varchar pk, even half a ts). Discard it
+        # instead of parsing it: a fully written line is never missing its
+        # terminator, so this can never drop content that actually
+        # finished landing.
+        try:
+            lines = self._cp_file_handler.readlines()
+        except OSError as ose:
+            raise MilvusException(
+                message=f"Failed to read cp info from file:{self._cp_file_path_str}"
+            ) from ose
+        if lines and not lines[-1].endswith("\n"):
+            lines = lines[:-1]
+        return lines
+
     def __set_up_ts_cp(self):
         self._buffer_cursor_lines_number = 0
         self._cp_file_path_str = self._query_options.get(ITERATOR_SESSION_CP_FILE, None)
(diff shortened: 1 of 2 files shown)


pymilvus/client/iterator/query_iterator.py
```diff
@@ -389,12 +413,30 @@ def next(self):
...
             ret = res[0 : min(self._query_options[BATCH_SIZE], len(res))]
 
         ret = self.__check_reached_limit(ret)
+        prev_next_id = self._next_id
+        prev_next_element_offset = self._next_element_offset
         self.__update_cursor(ret)
-        io_operation(self.__save_pk_cursor, "failed to save pk cursor")
+        try:
+            io_operation(self.__save_pk_cursor, "failed to save pk cursor")
+        except MilvusException:
```
yhmo, line 508: pymilvus/client/iterator/query_iterator.py line:424
High ---- This rollback assumes any `MilvusException` means the checkpoint was untouched, but `__save_pk_cursor()` writes/truncates before `flush()`, and `io_operation()` also wraps a `flush()` `OSError`. If the write was partial or the flush fails after accepting the cursor, memory/cache roll back while the file may already point past the undelivered batch (or contain a partial cursor); restarting from it can still skip rows. Please make the file update rollbackable/atomic and cover a post-write/flush failure, rather than treating every write exception as an aborted commit.

csy1204: You were right, and it was worse than the flush case alone. Force-pushed; the two commits are rewritten rather than patched on top.

**The rollback is now file-level.** `__save_pk_cursor()` captures the handle offset before the append and, if the write path raises, seeks/truncates/flushes back to it so the file matches the rolled-back in-memory cursor. If that recovery *itself* fails the state is genuinely unknown, so it raises a `MilvusException` saying the checkpoint file can no longer be trusted, chaining the original error rather than reporting a clean abort.

The decisive test is cross-restart, which nothing covered before: a wrapper forwards `writelines` to the real handle and really flushes it so the bytes land, then raises `OSError` from its own `flush()`. It asserts the on-disk bytes are unchanged and that a **freshly constructed** `QueryIterator` over that same path resumes at the pre-failure cursor. Against the previous tip that test fails with `b'200\n5\n9\n'` where `b'200\n5\n'` was expected — exactly your scenario.

**Your reasoning also caught a second thing.** I had claimed plain pk cursor lines were safe from truncation "by prefix truncation". That is false for negative INT64 keys: `json.loads("-1")` succeeds as a bare int, so a `-1234` line truncated to `-1` parses, isn't a dict, and gets adopted — and `-1 > -1234`, so the resume silently skips that range. No exception, wrong answer. Measured on the old tip vs the new one:

```
before: _next_id='-1'      next expr='pk > -1 and (pk > 0)'      # rows in (-1234, -1] skipped
after : _next_id='-1234'   next expr='pk > -1234 and (pk > 0)'
```

Rather than document that as a limitation I closed it structurally: every line the iterator writes is newline-terminated and writes are sequential, so a partially written line cannot carry its terminator. The reader now **discards an unterminated final line** before parsing it, which covers int, varchar and JSON element cursors uniformly and falls back to the last intact cursor line (or to the ts-only / fresh-init states that the first commit made valid). The claim in the code comment is corrected accordingly.

Residual exposure is stated in the commit message rather than implied away: corruption that preserves a line's terminator (bit rot, or a cp file reused across mismatched collections) is not detected by this rule, and durability is `flush()` not `fsync()` throughout — both unchanged from before this PR.

`tests/unit/orm/test_iterator.py` 160 passed, full suite 4553 passed / 3 skipped, black and ruff clean. Note the repo's CI hasn't run on either of my PRs — the workflows sit in `action_required`, so they need a maintainer to approve them.

yhmo: The issue is confirmed fixed.

PR #3753, opened 2026-08-10
enhance: resolve global cluster topology via DNS SRV records

huanghaoyuanhhy:
## Why

Global cluster connections previously fetched topology by hitting the global endpoint directly. This PR switches the SDK to discover ha-manager seed servers through the `_grpc._tcp.<host>` SRV record set, probe them concurrently, and use the topology with the highest version — with strict no-rollback semantics for every subsequent refresh path.

## What changed

- **SRV discovery**: resolve `_grpc._tcp.<endpoint hostname>` to the ha-manager seed list (typically 1-5 servers, ordered; priority/weight are not differentiated in deployment). Endpoints without SRV records raise instead of falling back to the legacy direct-endpoint path.
- **Seed selection & probing**: probe at least two seeds concurrently when more than one is returned — uniformly at random when all weights are zero, weighted otherwise. The first fetch (no cached version) returns the first seed's answer immediately; the remaining seeds keep being polled in the background and trigger the topology replacement flow if a higher version arrives. Refreshes with a cached version poll all seeds and take the highest.
- **Version management (never roll back)**: memory topology only advances when the fetched version is strictly higher; failures keep the cached topology. Version comparisons now run against the connection strategy's shared topology (single source of truth) instead of a private refresher snapshot, so no path can regress the version.
- **Retry semantics**: transient DNS failures and unreachable seeds retry with exponential backoff (3 attempts); deterministic rejections (no SRV records, server-side auth errors) fail fast without retry.
- The endpoint hostname is passed verbatim via `?endpoint=`; the server derives the global cluster id from its first DNS label, so the SDK does no gcid parsing.
- Adds a `dnspython` dependency for SRV resolution.

## Tests

- Unit tests covering SRV resolution, candidate selection, concurrent probing (first-answer fast path, background higher-version replacement, wait-all highest-version), retry/backoff, deterministic-failure fast-fail, and refresher version guarding.
- `193 passed`

## Deployment note

HA manager SRV records must be in place before this SDK version ships — endpoints without SRV records will raise instead of falling back to the legacy path.


pymilvus/client/connection_manager.py
```diff
@@ -321,50 +321,80 @@ def __init__(self):
...
 
         self._refresher = TopologyRefresher(
             global_endpoint=config.uri,
             token=config.token,
             topology=topology,
             on_topology_change=self._on_topology_change,
+            get_current=self.get_topology,
         )
         self._refresher.start()
         return topology
```
bigsheeper: **[1] Returns the first answer even after `_accept_topology` rejected it.**

When the `_watch_remaining` thread delivers a newer topology through `on_topology_change` before `fetch_topology` returns, `_accept_topology(topology)` above rejects the first answer, but this line still returns it. `GlobalStrategy.create_handler` then builds `GrpcHandler(uri=P1)` from the superseded primary while `self._topology` says P2.

Scenario: seed A answers v5 (primary P1); watcher installs v9 (primary P2) before this method returns. Writes go to a demoted cluster from the first request. `test_initial_fetch_does_not_overwrite_newer_callback` checks `get_topology()` but never checks which uri the handler was created with.

Fix: `return self.get_topology()` (the accepted topology) instead of the raw first answer, and pass that to `TopologyRefresher` as well.

pymilvus/client/connection_manager.py
```diff
@@ -321,50 +321,80 @@ def __init__(self):
...
+        # None means no seed has anything newer; a stale answer is rejected by
+        # the compare-and-set either way, so the cached topology never regresses.
+        if new_topology is not None:
+            self._accept_topology(new_topology)
+
+        current = self.get_topology()
+        new_primary = current.primary.endpoint if current else None
 
         # Only recover if primary actually changed
         if old_primary != new_primary:
```
bigsheeper: **[2] `on_unavailable` compares the cached primary with itself, so a connection stuck on a demoted primary never recovers.**

Both the background watcher and `TopologyRefresher` update the cache through `_accept_topology` without reconnecting. When the handler is still on P1 and the cache already says P2:

1. `old_primary` = P2 (from cache)
2. `fetch_topology(cached_version=9)` returns `None` (nothing newer)
3. `new_primary` = P2
4. `old_primary == new_primary` -> return `False` -> `ConnectionManager.handle_error` never calls `_recover`

Every retry hits P1 again until the process restarts. Combined with [1] this happens right at connection time.

The decision should compare the current primary against the address the handler is actually connected to, not against the previous cached primary.

pymilvus/client/global_topology.py
```diff
@@ -80,56 +100,309 @@ def _parse_topology_response(data: dict) -> GlobalTopology:
...
+
+        if not targets:
+            raise MilvusException(
+                message=(
+                    f"No SRV records found for global cluster endpoint "
+                    f"'{hostname}' ({SRV_SERVICE_PREFIX}{hostname})"
                 )
-                time.sleep(delay)
+            )
+        candidates = _pick_candidates(targets)
```
bigsheeper: **[3] Refresh path polls only `DEFAULT_PROBE_COUNT=2` seeds, not "every seed".**

The `_probe_candidates` docstring and the PR description say that with a cached version every seed is polled and the highest version wins, but `_pick_candidates(targets)` always uses `count=2`.

Scenario: SRV returns 5 seeds, a failover bumps the version to 10 but only 3 seeds have replicated it. If both picked seeds still report v9, `fetch_topology` returns `None`, `on_unavailable` returns `False` (no recovery), and the refresher keeps the stale topology for another 5 minutes.

Either pass `count=len(targets)` when `wait_all` is true, or correct the docstring and PR text.

pymilvus/client/global_topology.py
```diff
@@ -80,56 +100,309 @@ def _parse_topology_response(data: dict) -> GlobalTopology:
...
+    answers wins so a slow seed does not stall the initial connection; the
+    seeds still in flight keep being polled in the background and
+    ``on_higher_version`` triggers the replacement flow when one returns a
+    newer version. An API-level error (e.g. auth failure) is the same across
+    seeds, so it is captured and surfaced when no seed succeeds instead of a
+    generic "unreachable".
+    """
+    best: Optional[GlobalTopology] = None
+    api_error: Optional[MilvusException] = None
+    executor = ThreadPoolExecutor(max_workers=len(candidates))
```
bigsheeper: **[4] Leftover probe workers are non-daemon threads and are joined at interpreter exit.**

After the first-answer fast path returns, the remaining `requests.get` calls keep running in this executor's worker threads. `shutdown(wait=False, cancel_futures=True)` cannot cancel a running future, and `concurrent.futures.thread` registers an atexit hook that joins every worker.

Scenario: short script (connect, one query, exit). Seed A answers fast, seed B drops SYNs. The interpreter hangs on exit for up to connect+read timeout (~2x `REQUEST_TIMEOUT`, 20s).

Use daemon threads for the background probes (or a plain `threading.Thread(daemon=True)` per seed), or use a shorter timeout for the seeds that are still in flight after the first answer.

tests/unit/test_global_topology.py
```diff
@@ -127,65 +135,259 @@ def test_raises_when_no_primary(self):
...
+            _FakeSrvRecord("ha-mgr-ap1.global-cluster.example.com", priority=10),
+            _FakeSrvRecord("ha-mgr-us1.global-cluster.example.com", priority=20),
+        ), patch(
+            "pymilvus.client.global_topology.requests.get",
+            side_effect=[_mock_response(version="5"), _mock_response(version="9")],
+        ):
+            topology = fetch_topology(_GLOBAL_URL, _TOKEN, cached_version=1)
+            assert topology.version == 9
+
+    def test_first_answer_wins_without_cached_version(self):
```
bigsheeper, line 330: **[5] Flaky: completion order of two concurrent probes is not deterministic.**

This test and `test_background_higher_version_triggers_replacement` (line 297) feed one shared `requests.get` `side_effect=[v5, v9]` to two probe threads and assert that the v5 answer completes first. The thread that calls `requests.get` second (and receives v9) can finish its future first under GIL scheduling; `as_completed` then yields v9, `fetch_topology` returns version 9, and `assert topology.version == 5` fails intermittently. In the replacement test the event then never fires and the test times out.

Make the v9 response block on a `threading.Event` that the test releases after the first answer is returned, or key the response on the target hostname, so the order is deterministic.

pymilvus/client/global_topology.py
```diff
@@ -80,56 +100,309 @@ def _parse_topology_response(data: dict) -> GlobalTopology:
...
+            delay = _backoff_sleep(attempt)
+            logger.warning(
+                f"All {len(candidates)} topology seeds unreachable "
+                f"(attempt {attempt + 1}); retrying in {delay:.1f}s"
+            )
 
     raise MilvusException(
-        message=f"Failed to fetch global topology after {MAX_RETRIES} attempts: {last_error}"
+        message=(
+            f"Failed to fetch global topology from {len(candidates)} seed(s) "
```
bigsheeper: **[6] Final error drops the underlying cause.**

The old code appended `last_error` to the final `MilvusException`. Now the caller only sees `Failed to fetch global topology from N seed(s) ... after 3 attempts` with no status code, response body, or exception text. The per-attempt warnings are at WARNING level and are often not captured by applications.

Keep the last seed error (e.g. the last `RuntimeError` from `_probe_seed`) and append it to the message.

pymilvus/client/global_topology.py
```diff
@@ -80,56 +100,309 @@ def _parse_topology_response(data: dict) -> GlobalTopology:
...
-            if attempt < MAX_RETRIES - 1:
-                delay = min(BASE_DELAY * (2**attempt), MAX_DELAY)
-                delay += random.uniform(0, delay * 0.1)  # noqa: S311
-                logger.warning(
-                    f"Topology fetch attempt {attempt + 1} failed: {e}. Retrying in {delay:.1f}s"
+            targets = _resolve_srv(hostname)
+        except MilvusException as e:
+            if attempt == MAX_RETRIES - 1:
+                raise
+            delay = _backoff_sleep(attempt)
```
bigsheeper: **[7] Sleep happens before the "retrying in Xs" log (same at line 395).**

`_backoff_sleep` blocks first, then the warning is emitted, so the process is silent for 1s, 2s, 4s before each message and the log timeline reads as if the delay comes after the message. The old code computed the delay, logged, then slept. Split `_backoff_sleep` into `_backoff_delay(attempt)` + `time.sleep(delay)`, or log inside the helper before sleeping.

pymilvus/client/global_topology.py
```diff
@@ -80,56 +100,309 @@ def _parse_topology_response(data: dict) -> GlobalTopology:
...
+    """Watch the seeds still in flight and trigger replacement on newer answers.
+
+    Runs in a daemon thread after the first answer is handed over, so a slow
+    seed does not stall the initial connection while a newer version from the
+    remaining seeds can still win the replacement flow.
+    """
+
+    def watch() -> None:
+        try:
+            for future in as_completed(futures):
```
bigsheeper: **[8] Watcher re-processes futures that `_probe_candidates` already handled.**

`as_completed(futures)` starts from scratch, so seeds that failed before the first success are read again here: their failure is logged a second time without the target name, and a `MilvusException` (e.g. an auth error captured as `api_error` in the caller) is downgraded to a generic "Topology probe failed" warning.

Pass only the futures that are still pending (the ones the caller's loop did not consume) to the watcher.

bigsheeper (commented):
Code review of the SRV topology resolution change. 10 findings, ranked most severe first as inline comments (1 = most severe). Findings 1-5 are correctness issues; 6-10 are diagnostics and cleanup.

pymilvus/client/global_topology.py
```diff
@@ -80,56 +100,309 @@ def _parse_topology_response(data: dict) -> GlobalTopology:
...
+        return []
+    except dns.exception.DNSException as e:
+        raise MilvusException(message=f"SRV resolution failed for {srv_name}: {e}") from e
+
+    targets = [
+        SrvTarget(
+            priority=rr.priority,
+            weight=rr.weight,
+            port=rr.port,
+            target=str(rr.target).rstrip("."),
```
bigsheeper: **[9] SRV target `.` (RFC 2782 "service decidedly not available") becomes an empty hostname.**

`str(rr.target).rstrip(".")` turns `.` into `""`, so `_probe_seed` builds `https://:<port>/global-cluster/topology`, `requests` raises `InvalidURL` (caught as a generic exception), and `fetch_topology` spends ~7s in backoff before raising a misleading "seeds unreachable" error, instead of failing fast like NXDOMAIN/NoAnswer.

Filter out targets equal to `""` (or `.` before stripping) and let the empty result take the "No SRV records found" path.

pymilvus/client/global_topology.py
```diff
@@ -199,24 +478,43 @@ def _refresh_loop(self) -> None:
...
-                        except Exception:
-                            logger.warning("Topology change callback failed", exc_info=True)
+            current = self._current_topology()
+            new_topology = fetch_topology(
+                self._global_endpoint, self._token, cached_version=current.version
+            )
+
+            if new_topology is not None and new_topology.version > current.version:
+                logger.info(
+                    f"Topology updated: version {current.version} -> {new_topology.version}"
```
bigsheeper: **[10] Misleading "Topology updated" log and a dead private snapshot when `get_current` is set.**

This is logged before the compare-and-set in `_on_topology_change` decides. Scenario: refresher reads v7, fetch returns v8, meanwhile `on_unavailable` installed v10. The log says `Topology updated: version 7 -> 8`, `_accept_topology` rejects v8 at DEBUG level, and an INFO reader believes the topology rolled back.

Also, with `get_current` set, `self._topology` is never updated, so `TopologyRefresher.get_topology()` returns the initial topology forever, and the `_get_current is None` branch duplicates the CAS logic the strategy already owns.

Simpler form: make the strategy the single owner (require `get_current`, drop the private snapshot and this branch), have the callback return the CAS result, and log only when it was accepted.

PR #3757, opened 2026-08-12
fix: coerce uuid.UUID and os.PathLike to str for VARCHAR/JSON fields

### LINKED ISSUE #2917: [Bug]: type support issue : unrecognized dtype for key (WindwosPath、UUID)

pangwangshu:
## Summary

Fixes #2917. Inserting a `uuid.UUID` or `pathlib.Path` value into a VARCHAR field — or nesting one inside a JSON/dynamic field — failed instead of auto-converting to a string, as the issue requested.

#3408 attempted this by patching dtype inference only. @XuanYang-cn correctly flagged that as incomplete ([comment](https://github.com/milvus-io/pymilvus/pull/3408#issuecomment-4241839977)): `convert_to_str_array`/`entity_to_str_arr` reject non-`str` values independently of dtype inference. This PR fixes all three points on that path:

- **`orm/types.py`** — `infer_dtype_by_scalar_data`/`infer_dtype_bydata` recognize `UUID`/`PathLike` as `VARCHAR` (previously `UNKNOWN`; `PathLike` could also crash `infer_dtype_bydata` via an uncaught `TypeError`).
- **`client/entity_helper.py`** — `convert_to_str_array` coerces `UUID`→`str` / `PathLike`→`os.fspath()`, for both the row-insert (scalar) and column-insert (list) call paths.
- **`client/entity_helper.py`** — `convert_to_json` now handles `PathLike` nested in JSON/dynamic fields (previously a raw `orjson` `TypeError`).

## Test plan

- [x] Unit tests for dtype inference, `convert_to_str_array` (both call paths), and `convert_to_json`
- [x] End-to-end tests via `pack_field_value_to_field_data` / `entity_to_field_data` — the real insert path, not just the helpers in isolation (the gap that made #3408 look sufficient)
- [x] `make coverage` (full unit suite) — passing
- [x] `make integration-lite` — passing
- [x] `make lint` — clean


pymilvus/client/entity_helper.py
```diff
@@ -254,6 +275,15 @@ def assign_to_parent(value: Any) -> None:
...
                 assign_to_parent(bool(current))
+            elif isinstance(current, os.PathLike):
+                # pathlib.Path (and subclasses like PosixPath/WindowsPath,
+                # plus the pure variants PurePosixPath/PureWindowsPath) are
+                # not natively JSON-serializable by orjson, which raises a
+                # raw, unhelpful "Type is not JSON serializable" TypeError.
+                # Path values are unambiguous as their string form, so
+                # convert them here alongside the other leaf-type
+                # normalizations. See GH-2917.
+                assign_to_parent(os.fspath(current))
```
yhmo: pymilvus/client/entity_helper.py line:286
Medium ---- `preprocess_numpy_types` only descends into `dict` and `list`, but `orjson` also accepts tuples as JSON arrays. A value such as `{"paths": (PurePosixPath("/a"),)}` therefore leaves the path untouched and still raises `TypeError`, so nested PathLike support remains container-dependent. Please traverse tuples as well (converting them to a list/tuple of processed values) and add a regression case.

pymilvus/client/entity_helper.py
```diff
@@ -197,7 +199,26 @@ def get_max_len_of_var_char(field_info: Dict) -> int:
     return field_info.get("params", {}).get(k, v)
 
 
+def _coerce_str_like(value: Any) -> Any:
+    """Coerce a uuid.UUID or os.PathLike value to its string form; pass through otherwise."""
+    if isinstance(value, os.PathLike):
+        return os.fspath(value)
```
yhmo: pymilvus/client/entity_helper.py line:205
Medium ---- `os.PathLike.__fspath__()` is allowed to return either `str` or `bytes`; this returns the bytes form unchanged. Such a valid PathLike is then rejected by the VARCHAR string check and also remains non-JSON-serializable in `convert_to_json`, despite the new public behavior claiming PathLike coercion. Please normalize the result to text (for example with `os.fsdecode`) or narrow the accepted contract, and cover a custom bytes-returning PathLike.

pymilvus/client/entity_helper.py
pangwangshu: Good catch — fixed in 79e1054e. `preprocess_numpy_types` now treats `tuple` the same as `list` (staged as a mutable list for traversal, then serialized as a JSON array either way — orjson doesn't distinguish), so a PathLike nested in a tuple gets converted too. Added `test_json_with_path_nested_in_tuple`.

pymilvus/client/entity_helper.py
pangwangshu: Good catch — fixed in 79e1054e. Switched both coercion points (`_coerce_str_like` and the `convert_to_json` PathLike branch) from `os.fspath()` to `os.fsdecode()`, which normalizes a bytes-returning `__fspath__()` to text instead of leaving raw bytes behind. Added `test_convert_to_str_array_coerces_bytes_returning_pathlike` and `test_json_with_bytes_returning_pathlike` with a custom bytes-returning PathLike.

shashvat-singham (commented):
Ran the branch. The VARCHAR side does what it says, and the tuple fix in the walker is a nice catch — `(PurePosixPath("/a/b"),)` really did stay unconverted before.

I went looking for a `uuid.UUID` branch in `preprocess_numpy_types` to match the `os.PathLike` one and couldn't find it, assumed that was an oversight, then tested it and found it works anyway — orjson serialises `UUID` natively, so the walker doesn't need to touch it. Worth noting in case a future reader has the same reaction I did.

That does leave one real gap, though: it works *because* orjson handles it, and orjson isn't always the serialiser. `convert_to_json` falls back to stdlib `json` past orjson's ~500-level recursion limit, and stdlib `json` has no native `UUID` support:

```python
def nest(leaf, depth):
    o = leaf
    for _ in range(depth):
        o = {"n": o}
    return o

convert_to_json(nest(uuid.UUID(int=0), 10))    # b'{"n":...{"n":"00000000-..."}}'
convert_to_json(nest(uuid.UUID(int=0), 600))   # TypeError: Object of type UUID is not JSON serializable
convert_to_json(nest(PurePosixPath("/a/b"), 600))  # fine
```

`PathLike` survives the deep case precisely because this PR normalises it to `str` *in the walker*, before either serialiser sees it. `UUID` doesn't, so a JSON field nested past ~500 levels still raises the raw `TypeError` that GH-2917 is about.

To be clear this fails the same way on `master`, so it's not a regression from this PR — but since the PR's stated goal is making `UUID` and `PathLike` work in VARCHAR and JSON fields, and it fixes the deep case for one and not the other, it seems worth closing here. Adding

```python
elif isinstance(current, uuid.UUID):
    assign_to_parent(str(current))
```

next to the `PathLike` branch would make both paths serialiser-independent and match the reasoning already in that comment block.

Admittedly 500-deep JSON is an unusual payload, so if you'd rather keep this PR tight I'm happy to send that as a follow-up — just flagging it while the context is fresh.

pangwangshu:
@shashvat-singham Good catch, thanks for digging into the fallback path. Added an explicit `uuid.UUID` branch next to the `PathLike` one in `preprocess_numpy_types` so it's normalized to `str` before either serializer sees it, instead of relying on orjson's native support: 07cfff71

Verified with your exact repro — `convert_to_json(nest(uuid.UUID(int=0), 600))` now succeeds instead of raising `TypeError`, and added a regression test covering the >500-level stdlib `json` fallback case.

PR #3764, opened 2026-08-18
feat: support client-side roaring bitmap blobs for roaring_match

xiaofan-luan:
`roaring_match(field, {ids})` is the exact-membership sibling of `bloom_match`, added server-side in milvus-io/milvus#51968 (merged). The client compresses a membership set into a roaring bitmap and ships the blob as a template value, so a set that would blow past the proxy's gRPC receive limit as an `in [...]` list rides the wire in a fraction of the space — and, unlike a bloom filter, with no false positives, which is why the server also permits it in delete expressions.

The expression accepts **only** a `{template}` placeholder resolving to protobuf `bytes` holding a complete MRB1 blob. There is no literal-list form. So without an SDK-side builder the feature is unreachable from Python entirely.

## What this adds

- `pymilvus/client/roaring_filter.py` — `build_roaring_bitmap(members) -> bytes` and a streaming `RoaringBitmapBuilder`, both exported from `pymilvus/__init__.py` next to `build_bloom_filter` / `BloomFilterBuilder`.
- 180 tests plus the shared cross-SDK golden vectors, beside the bloom fixture.

No new dependency, and the template path is **not** touched: the bytes branch added for `bloom_match` in #3690 already carries the blob unchanged through query, search and delete. A test asserts that end of it rather than re-plumbing it.

## Why no pyroaring dependency

Most roaring libraries' default serialization is *not* the portable format segcore decodes — that mismatch is the documented trap in the design doc. The writer is ~200 lines and the client only ever writes, never parses, so hand-writing it avoids a native dependency and a serialization dialect we don't control.

## Usage

```python
from pymilvus import build_roaring_bitmap

ids = build_roaring_bitmap([-7, 42, 1001])          # reuse this across queries
client.query(
    collection_name="docs",
    filter="roaring_match(creator_id, {ids})",
    filter_params={"ids": ids},
)
```

Building the blob is the expensive part, so it is meant to be built once and reused — that amortization is the entire point of the feature. For a set that arrives in pieces, `RoaringBitmapBuilder().add_int64_batch(page)` avoids ever materializing the full member list.

## Verification

The bytes must match what every other SDK produces, so conformance is byte-identity against the Go reference (`client/roaringfilter.Build`), not just "the server accepts it":

- **29/29 shared golden vectors byte-identical**, sha256 included, plus the fixture's own container decomposition (high/low container counts and array/bitmap/run kinds) checked against the builder's.
- **300/300 additional randomized cases byte-identical** to the Go reference, on inputs not in the fixture set.
- The fixtures themselves were verified to pass the server's own validator and to decode under CRoaring — the library segcore probes with — with exact membership.
- A test-only reader recovers the exact member set from all 29 blobs and from randomized shapes.
- 19 deliberate mutations each turned a test red, including the tie-break constant, a 2^63-biased mapping, 32-bit key truncation, dropped/misplaced offset tables, off-by-one descriptors and run lengths, MSB-first bitmap and run-bitmap writes, and signed sorting during builder compaction.

One subtlety worth flagging for reviewers: the run-vs-bitmap tie-break compares against **8224**, not 8192. That is the reference's in-memory `bitmapContainerSizeInBytes()` (`unsafe.Sizeof(bitmapContainer{}) + 65536/8`). Using 8192 flips the encoding for containers with 2048..2055 runs *when the array encoding is larger still*, and desynchronizes this SDK's bytes from every other one. The `run_bitmap_tiebreak_run` / `run_bitmap_tiebreak_bitmap` vectors sit either side of it.

## An honest note on one mutation

`np.dtype("u2")` vs `np.dtype("<u2")` is an **equivalent mutant at runtime on a little-endian host** — numpy resolves the native form eagerly, so `np.dtype("u2") is np.dtype("<u2")` is literally `True` and no runtime assertion can distinguish them. (The `newbyteorder` guard in the bloom tests has the same blind spot.) It is a real bug on a big-endian client, which is exactly what #3690's third review round was about. It is pinned here by a test that greps the module source for `np.dtype("…")` literals and requires the explicit `<`.

## Tests

```
PYTHONPATH=$PWD uv run --group dev pytest tests/unit/test_roaring_filter.py -q   # 180 passed
PYTHONPATH=$PWD uv run --group dev pytest tests/unit -q                          # 4716 passed, 3 skipped
make lint                                                                        # clean
```

Baseline before this change was 4536 passed / 3 skipped / 0 failed, so there are no pre-existing failures in play.

Design doc: `docs/design-docs/design_docs/20260714-roaring-exact-membership-expression.md` in milvus-io/milvus.


pymilvus/client/roaring_filter.py
```diff
@@ -0,0 +1,528 @@
...
+            yield
+        finally:
+            self._lock.release()
+
+    def _compact(self) -> None:
+        """Merges the pending batches into the sorted distinct key set."""
+        if not self._parts:
+            return
+        fresh = np.unique(np.concatenate(self._parts))
+        self._parts = []
```
XuanYang-cn, line 504: These assignments discard the only references to pending members before the later merge allocations and sort complete. If `build()` or `cardinality` raises `MemoryError` after this point, `_operation()` releases the lock without poisoning the builder, so a retry silently emits only the old `_keys`. I reproduced: build `{1}`, add `{2}`, inject failure in the second `np.concatenate`, then retry; cardinality became `1` and member `2` was lost. That creates false negatives in an exact ACL/delete predicate. Please commit the cleared state only after the merge succeeds, or retain/restore the pending parts and poison the builder if rollback is impossible.

pymilvus/client/roaring_filter.py
```diff
@@ -0,0 +1,528 @@
...
+
+def _serialize(keys: np.ndarray) -> bytes:
+    """Wrap the portable Roaring64 serialization of ``keys`` in an MRB1 envelope."""
+    if keys.size == 0:
+        # An empty set is still a well-formed bitmap: zero high containers, an eight-byte body.
+        return struct.pack(
+            _HEADER_FORMAT, _MAGIC, _VERSION, _FORMAT_PORTABLE_ROARING64, 0, 8, 0
+        ) + (b"\x00" * 8)
+
+    _check_high_container_limit(_count_high_containers(keys))
```
XuanYang-cn: The accepted design says the client counts both high and low containers before bitmap construction and rejects impossible sparse shapes early. This pass checks only the high-container cap. A shape such as 200K high containers × 4 low containers is already guaranteed to exceed the 64 MiB decoded limit from fixed overhead alone (76.8 MiB), yet it enters `_plan()` and allocates member/container-sized arrays before failing. Please count both levels in this bounded pass and reject when the overhead-only lower bound exceeds the limit; the later exact body-inclusive check can remain for borderline cases.

tests/unit/test_roaring_filter.py
```diff
@@ -0,0 +1,779 @@
...
+    assert values["rb"].bytes_val == blob
+    assert values["rb"].WhichOneof("val") == "bytes_val"
+
+
+def test_roaring_blob_reaches_query_search_and_delete_requests():
+    """The blob must arrive as protobuf bytes -- not base64, not a string -- on every path."""
+    blob = build_roaring_bitmap([1, -1, 1 << 40])
+    expr = "roaring_match(id, {rb})"
+
+    query = Prepare.query_request("coll", expr, [], None, expr_params={"rb": blob})
```
XuanYang-cn, line 853: The design test plan explicitly requires Search, Query, and Delete bytes transport coverage “without bypassing the high-level client.” Calling `Prepare.*` directly proves protobuf preparation but bypasses every public client route. Please invoke the public query/search/delete APIs with mocked transport and assert that the captured requests preserve the blob in `bytes_val`.

XuanYang-cn (changes requested):
Requesting changes for the exactness/exception-safety bug and the incomplete pre-layout admission check. I reproduced the first issue with an injected allocation failure: after successfully adding `{1, 2}`, retrying the builder returned cardinality `1` and a blob containing only `{1}`.

Focused verification is otherwise clean: 182 roaring-filter tests passed, `make lint` passed, and `git diff --check` passed.

Non-blocking standards cleanup for the next revision: remove or merge the strict-subset/duplicate tests around signed mapping and limit handling, parametrize the two identical monkeypatch-and-error limit tests, and have `_plan()` reuse `_check_high_container_limit()` instead of duplicating the check.

xiaofan-luan:
Thanks @XuanYang-cn — all three addressed in `cdd81126`. The first one was a real bug and your reproduction was exact.

**1. Compaction dropped pending members on failure.** Confirmed, and worse than it looks: `_compact` released `_parts` before the merge that consumed them, and since neither `build()` nor `cardinality` poisons the builder, the retry *succeeded* and returned a bitmap silently short of whatever had been pending. In an exact membership predicate that is a false negative reported as success — no error, no metric, just fewer rows deleted or matched than the caller asked for.

Rollback is possible here, so I took that branch rather than poisoning: `_compact` now computes the merged key set first and commits `_keys`, `_parts` and `_pending` only once every allocation that can fail is behind it. A failure leaves the builder byte-for-byte as it was and the retry sees every member.

The regression test injects `MemoryError` into the second `np.concatenate` — the merge, not the gather — exactly as you described, and asserts the builder still reports both members afterwards. Reverting to the early release turns it red.

**2. Count both levels in the bounded pass.** Agreed, and 200K × 4 is the right example: 76.8 MiB of fixed overhead before a single body byte, so it cannot fit however the bodies encode, and that verdict is decidable from the counts alone. Both levels are counted now and the gate decides both limits when the counts settle it.

One deliberate wrinkle, since you may have an opinion on it. The overhead is a *lower bound* on the estimate, not the estimate, so the message says the decoded size is **"at least" N** rather than quoting N as the size. I had it the other way first and it was wrong: a caller reading that error has to know what to shrink against, and handing them an understated figure misinforms them. Borderline sets, where the body is what tips it over, still fall through to the exact check in `_plan` and get the real number. Two tests pin the split — one asserts `_plan` is never entered for the 200K × 4 shape, the other asserts it *is* entered for a set that only the body pushes over, and that the exact figure is what surfaces.

(The same narrowing applies to the C++ SDK's gate in milvus-io/milvus-sdk-cpp#583. That one has `lgtm` already, so I will follow up there separately rather than reset it.)

**3. Transport coverage bypassed the client.** Fair — it proved protobuf preparation and nothing else. There is now a test that patches only the gRPC stub, builds a real `MilvusClient`, calls the public `query` / `search` / `delete`, and asserts the captured requests carry the blob in `bytes_val`. The mocked stub answers with an empty response which the client rejects while parsing; that happens after the request is sent, and the request is what is under test. The `Prepare`-level test stays as the narrower check of the protobuf round trip.

Verification: 183 roaring tests, 4719 unit tests overall, `make lint` clean, and the 300 randomized cross-check cases against the Go reference still 300/300 byte-identical.

XuanYang-cn (approved):
/lgtm

PR #3766, opened 2026-08-20
fix: preserve top-level type params in FieldSchema.construct_from_dict

### LINKED ISSUE #2046: [Bug]: FieldSchema.construct_from_dict() not support params  max_length and dim

### Describe the bug

When dict has` max_length` or `dim` 
`{'name': 'error_message', 'type': <DataType.VARCHAR: 21>, 'max_length': 1024}`
 ` a_dict={'name': 'error_msg_vector', 'type': <DataType.FLOAT_VECTOR: 101>, 'dim': 768}`
When I new `FieldSchema.construct_from_dict(a_dict)`,  got 

return FieldSchema(raw["name"], raw["type"], raw.get("description", ""), max_length=raw.get("max_length", 128), **kwargs)
TypeError: pymilvus.orm.schema.FieldSchema() got multiple values for keyword argument 'max_length'

and

pymilvus.exceptions.MilvusException: <MilvusException: (code=65535, message=dimension is not defined in field type params, check type param `dim` for vector field)>

### Expected Behavior

When I new `FieldSchema.construct_from_dict(a_dict)`  no error.

```markdown
fields_dict = [
{'name': 'error_message', 'type': <DataType.VARCHAR: 21>, 'max_length': 1024},
{'name': 'error_msg_vector', 'type': <DataType.FLOAT_VECTOR: 101>, 'dim': 768}
]
fields = []
for f_dict in fields_dict:
    fields.append(FieldSchema.construct_from_dict(f_dict))
```

### Environment details

```markdown
pymilvus==2.4.0
```

### Anything else?

add  below  code  at pymilvus/orm/schema.py  can fix it.
# within  def construct_from_dict(cls, raw: Dict):
371         if raw.get("max_length") is not None:
372             kwargs["max_length"] = raw.get("max_length")
373         if raw.get("dim") is not None:
374             kwargs["dim"] = raw.get("dim")
pymilvus orm schema py-construct_from_dict

mameikagou:
## What

`FieldSchema.construct_from_dict()` currently reads type parameters only from the nested `params` mapping. Dictionaries that provide `max_length` or `dim` at the top level therefore lose those values, so the reconstructed VARCHAR or vector field is incomplete.

This preserves top-level `max_length` and `dim` while keeping the canonical nested `params` value authoritative when both forms are present. Explicit top-level `None` values remain ignored.

Fixes #2046

## Tests

- `python -m pytest -q --capture=no tests/unit/orm/test_schema.py` (259 passed)
- `black --check pymilvus/orm/schema.py tests/unit/orm/test_schema.py`
- `ruff check pymilvus/orm/schema.py tests/unit/orm/test_schema.py`


pymilvus/orm/schema.py
```diff
@@ -657,6 +657,9 @@ def _parse_type_params(self):
     def construct_from_dict(cls, raw: Dict):
         kwargs = {}
         kwargs.update(raw.get("params", {}))
+        for key in ("max_length", "dim"):
```
yhmo: pymilvus/orm/schema.py line:660
Low ---- The hardcoded tuple only covers max_length and dim, but the other entries of COMMON_TYPE_PARAMS (max_capacity for ARRAY fields, enable_match/enable_analyzer, analyzer_params) are still silently dropped when provided at top level, so a dict like {"name": "tags", "type": DataType.ARRAY, "element_type": DataType.VARCHAR, "max_capacity": 100} still loses max_capacity. Consider iterating COMMON_TYPE_PARAMS instead, since it already drives _parse_type_params; that covers the whole class of bug and prevents future type params from drifting out of sync with this list.

mameikagou: Good catch. Updated `construct_from_dict` to iterate `COMMON_TYPE_PARAMS` instead of hardcoding `max_length` and `dim`. I also added regression coverage for `max_capacity`, analyzer flags, `analyzer_params`, and `multi_analyzer_params`. Focused tests now pass (29 cases), as does the full schema test file (263 tests).

pymilvus/orm/schema.py
```diff
@@ -657,6 +657,9 @@ def _parse_type_params(self):
     def construct_from_dict(cls, raw: Dict):
         kwargs = {}
         kwargs.update(raw.get("params", {}))
+        for key in COMMON_TYPE_PARAMS:
```
yhmo: pymilvus/orm/schema.py line:660
Low ---- mmap_enabled and warmup are also field type params (FieldSchema.__init__ stores them into self._type_params at lines 603-607 and to_dict() serializes them under "params"), but they are not part of COMMON_TYPE_PARAMS, so a dict like {"name": "x", "type": DataType.FLOAT_VECTOR, "dim": 128, "mmap_enabled": True} still silently drops mmap_enabled — the same class of top-level-loss bug this PR fixes for the other type params. The nested params path still carries them (the describe path maps "mmap.enabled" -> "mmap_enabled" inside params), so round-trips work, but top-level keys do not. Consider extending the merge to cover mmap_enabled/warmup or documenting the limitation. Non-blocking.

mameikagou: Addressed in 4006b1cc: top-level `mmap_enabled` and `warmup` are now preserved, while nested `params` still take precedence. Added regression coverage for both cases; all 267 schema tests pass.

yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #3767, opened 2026-08-22
fix(deps): install the milvus-lite extra on Windows for Python >=3.10 (#3676)

### LINKED ISSUE #3676: [Bug]: Windows: `pymilvus[milvus-lite]` does not install `milvus-lite` despite Windows support

### Describe the bug

````
## Summary

The documentation and README indicate that Milvus Lite supports Windows.

However, installing with the documented command:

pip install -U "pymilvus[milvus-lite]"

(or the equivalent `uv add "pymilvus[milvus-lite]"`)

does not install `milvus-lite` on Windows.

## Investigation

The package metadata contains:

```python
import importlib.metadata as m
print([r for r in m.requires("pymilvus") if "milvus" in r.lower()])
```

Output:

```text
[
  'pymilvus.model>=0.3.0; extra == "model"',
  'milvus-lite>=2.4.0; sys_platform != "win32" and extra == "milvus-lite"'
]
```

The dependency explicitly excludes Windows via:

```
sys_platform != "win32"
```

## Workaround

Installing `milvus-lite` directly works:

```
pip install pymilvus
pip install milvus-lite
```

or

```
uv add pymilvus
uv add milvus-lite
```

After installing `milvus-lite` directly, the following code successfully creates a local database, inserts vectors, and performs a search:

```python

client = MilvusClient("test.db")
```

This suggests that `milvus-lite` itself supports Windows, but the extra dependency metadata in `pymilvus` still excludes Windows.

### Expected Behavior

If Windows is now officially supported by `milvus-lite`, the `milvus-lite` extra in `pymilvus` should install on Windows as documented.

```

---

```markdown
1. Create a clean virtual environment.
2. Install:

   pip install -U "pymilvus[milvus-lite]"

3. Check installed packages.

`milvus-lite` is not installed.

Running:


client = MilvusClient("test.db")
`

raises:


ConnectionConfigException:
milvus-lite is required for local database connections.
Please install it with:
pip install pymilvus[milvus_lite]
```

### Environment details

```markdown
- OS: Windows 11 x64
- Python: 3.12
- pymilvus: 3.0.0
- milvus-lite: 3.0
```

### Anything else?

_No response_

Anai-Guo:
## Summary

Fixes #3676.

`pip install -U "pymilvus[milvus-lite]"` installs nothing on Windows, silently. The extra carries a `sys_platform!='win32'` marker:

```toml
milvus_lite = [
    "milvus-lite>=2.4.0;sys_platform!='win32'",
    ...
]
```

so the resolver drops the requirement, exits 0, and the user gets a pymilvus with no local mode — even though the README and docs list Windows as supported. There is no warning, because an extra whose markers all evaluate false is not an error.

## Why the marker was right, and why it no longer is

It was correct when written. milvus-lite `<3.0` published platform-specific wheels only:

```
milvus_lite-2.5.1-py3-none-macosx_10_9_x86_64.whl
milvus_lite-2.5.1-py3-none-macosx_11_0_arm64.whl
milvus_lite-2.5.1-py3-none-manylinux2014_aarch64.whl
milvus_lite-2.5.1-py3-none-manylinux2014_x86_64.whl
```

No Windows wheel existed, so excluding win32 was the only way to keep the extra resolvable.

Since `3.0` the project ships a single pure-Python wheel:

```
milvus_lite-3.2.0-py3-none-any.whl
```

`py3-none-any` installs on Windows. The exclusion is now stale.

## Why a new marker instead of relaxing the existing one

milvus-lite 3.x declares `requires_python >=3.10`, while pymilvus still supports 3.9 (`requires-python = '>=3.9'`). Simply dropping `!='win32'` would let the resolver try to satisfy `milvus-lite>=2.4.0` on Windows + Python 3.9, where no wheel exists for that platform.

Adding a separate win32 branch pinned to `>=3.0` keeps each environment pointed at a distribution that actually exists:

| platform | Python | resolved |
|---|---|---|
| win32 | 3.9 | *(nothing — correct, no candidate exists)* |
| win32 | 3.10 | `milvus-lite>=3.0` |
| win32 | 3.12 | `milvus-lite>=3.0` |
| linux | 3.9 | `milvus-lite>=2.4.0` *(unchanged)* |
| linux | 3.12 | `milvus-lite>=2.4.0` *(unchanged)* |
| darwin | 3.12 | `milvus-lite>=2.4.0` *(unchanged)* |

Generated by evaluating the three requirement markers with `packaging.markers.Marker(...).evaluate(env)`. **Non-Windows resolution is byte-for-byte unchanged.**

## Verification

Milvus Lite was exercised end-to-end on Windows 11 (Python 3.12.10, clean venv), not just installed:

```python
c = MilvusClient("wintest.db")
c.create_collection(collection_name="demo", dimension=4)
c.insert(collection_name="demo", data=[
    {"id": 1, "vector": [0.1, 0.2, 0.3, 0.4], "text": "hello"},
    {"id": 2, "vector": [0.9, 0.8, 0.7, 0.6], "text": "windows"},
])
print(c.search(collection_name="demo", data=[[0.1, 0.2, 0.3, 0.4]],
               limit=2, output_fields=["text"]))
```

```
milvus-lite 3.2.0
data: [[{'id': 1, 'distance': 1.0, 'entity': {'id': 1, 'text': 'hello'}},
        {'id': 2, 'distance': 0.8427009582519531, 'entity': {'id': 2, 'text': 'windows'}}]]
db file created: True
```

Collection creation, insert, and vector search all succeed, and the `.db` file is written.

## Notes

This supersedes #3701, which proposed the same version-aware split and was closed by its author on 2026-08-12 without maintainer objection. I kept the `>=3.10` gate for the reason given above and left the existing `setuptools<82` pin untouched.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,6 +78,10 @@ bloom_filter = [
 
 milvus_lite = [
     "milvus-lite>=2.4.0;sys_platform!='win32'",
+    # milvus-lite <3.0 only published macOS/manylinux wheels, hence the win32 exclusion
+    # above. 3.x ships a pure-Python py3-none-any wheel that works on Windows, but it
+    # requires Python >=3.10, so 3.9 on Windows still has no installable candidate.
+    "milvus-lite>=3.0;sys_platform=='win32' and python_version>='3.10'",
     # milvus-lite 2.5.1 uses pkg_resources on Python 3.9; setuptools 82 removed it.
     "setuptools<82;python_version<'3.10'",
 ]
--- a/tests/integration/lite/test_milvus_lite.py
+++ b/tests/integration/lite/test_milvus_lite.py
@@ -1,16 +1,12 @@
-import sys
-
 import numpy as np
 import pytest
 from pymilvus.exceptions import ConnectionConfigException
 from pymilvus.milvus_client import MilvusClient
 
-pytestmark = pytest.mark.skipif(
-    sys.platform.startswith("win"), reason="Milvus Lite is not supported on Windows"
-)
-
-if not sys.platform.startswith("win"):
-    pytest.importorskip("milvus_lite", reason="milvus-lite not installed")
+# milvus-lite only ships a Windows-capable wheel from 3.0 onward, and the milvus_lite
+# extra resolves it there only on Python >=3.10 (see pyproject.toml), so let the import
+# decide which runs are skipped instead of hard-skipping every Windows run.
+pytest.importorskip("milvus_lite", reason="milvus-lite not installed")
 
 
 class TestMilvusLite:
--- a/uv.lock
+++ b/uv.lock
@@ -14,7 +14,8 @@ resolution-markers = [
     "python_full_version == '3.13.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
     "python_full_version == '3.12.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
     "python_full_version == '3.11.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
-    "python_full_version == '3.10.*'",
+    "python_full_version == '3.10.*' and sys_platform != 'win32'",
+    "python_full_version == '3.10.*' and sys_platform == 'win32'",
     "python_full_version > '3.9' and python_full_version < '3.10'",
     "python_full_version <= '3.9'",
 ]
@@ -142,7 +143,8 @@ resolution-markers = [
     "python_full_version == '3.13.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
     "python_full_version == '3.12.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
     "python_full_version == '3.11.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
-    "python_full_version == '3.10.*'",
+    "python_full_version == '3.10.*' and sys_platform != 'win32'",
+    "python_full_version == '3.10.*' and sys_platform == 'win32'",
 ]
 dependencies = [
     { name = "requests", version = "2.34.2", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.10'" },
@@ -252,7 +254,8 @@ resolution-markers = [
     "python_full_version == '3.13.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
     "python_full_version == '3.12.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
     "python_full_version == '3.11.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
-    "python_full_version == '3.10.*'",
+    "python_full_version == '3.10.*' and sys_platform != 'win32'",
+    "python_full_version == '3.10.*' and sys_platform == 'win32'",
 ]
 dependencies = [
     { name = "click", version = "8.4.1", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.10'" },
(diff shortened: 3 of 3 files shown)


pyproject.toml
```diff
@@ -78,6 +78,10 @@ bloom_filter = [
 
 milvus_lite = [
     "milvus-lite>=2.4.0;sys_platform!='win32'",
+    # milvus-lite <3.0 only published macOS/manylinux wheels, hence the win32 exclusion
+    # above. 3.x ships a pure-Python py3-none-any wheel that works on Windows, but it
+    # requires Python >=3.10, so 3.9 on Windows still has no installable candidate.
+    "milvus-lite>=3.0;sys_platform=='win32' and python_version>='3.10'",
```
yhmo, line 84: pyproject.toml line:84
Medium ---- This new win32 branch is not reflected in the committed uv.lock. At the PR head the lock still records the milvus_lite extra as `milvus-lite>=2.4.0; sys_platform != 'win32' and extra == 'milvus-lite'` (uv.lock requires-dist, ~line 3559) and keeps milvus-lite 3.0's resolution markers win32-excluded for Python 3.11-3.14 (uv.lock ~lines 2114-2124). Both CI workflows run `uv sync --frozen`, which uses the lockfile as the source of truth and does not check freshness, so the Windows branch of this extra will never be installed in the repo's own environments, and any `uv lock` / `uv sync --locked` run will report drift. The superseded PR #3701 regenerated uv.lock (adding the faiss-cpu win_amd64 wheel) when it changed this extra, and CONTRIBUTING.md documents uv.lock as the locked dependency graph used by uv. Please run `uv lock` and commit the updated lockfile in this PR.

yhmo, line 84: pyproject.toml line:84
Medium ---- tests/integration/lite/test_milvus_lite.py (unchanged by this PR) still hard-skips the whole module on Windows: `pytestmark = pytest.mark.skipif(sys.platform.startswith("win"), reason="Milvus Lite is not supported on Windows")`. That marker now directly contradicts the behavior this PR enables, so the fix is never exercised by the repo's own CI (the windows-2022 matrix collects then skips the Lite tests), and a future reader trusting the marker will keep believing Windows is unsupported. The issue discussion (#3676) explicitly proposed enabling these tests on a supported Windows/Python combination, and the superseded PR #3701 de-skipped the same file. Please re-gate the skip on `python_version < '3.10'` instead of the platform (milvus-lite 3.x requires Python >=3.10), and consider adding a CI step that installs `.[milvus-lite]` on Windows so this metadata fix is actually verified.

Anai-Guo:
Thanks @yhmo — both points were correct and are fixed in 1d83032.

**1. `uv.lock` was stale.** Regenerated with `uv lock` (uv 0.9.x). The diff is +117/-65 and contains **zero package version changes** — it is entirely the mechanical consequence of the new marker: the `python_full_version == '3.10.*'` resolution marker splits into `… and sys_platform != 'win32'` / `… == 'win32'`, and `faiss-cpu` gains its already-published `win_amd64` / `win_arm64` wheel entries. The two lines that matter now read:

```
{ name = "milvus-lite", marker = "python_full_version >= '3.10' and sys_platform == 'win32' and extra == 'milvus-lite'", specifier = ">=3.0" },
{ name = "milvus-lite", marker = "sys_platform != 'win32' and extra == 'milvus-lite'", specifier = ">=2.4.0" },
```

**2. The module-level Windows skip contradicted the fix.** Dropped the `pytestmark = pytest.mark.skipif(sys.platform.startswith("win"), …)` and the `if not sys.platform.startswith("win")` guard around it, leaving the plain `pytest.importorskip("milvus_lite", …)` to decide. That keeps Windows + Python 3.9 skipping (no installable candidate, as the pyproject comment explains) while letting 3.10+ actually run. `import sys` is now unused and removed.

**Verified on real Windows**, not by reasoning — this machine is Windows 11 / `windows-2022`-class:

```
$ uv run --group dev --extra milvus-lite python -c "import milvus_lite, platform, sys; print(...)"
milvus_lite 3.0 py 3.11.15 Windows

$ PYTHONPATH=. uv run --group dev --extra milvus-lite pytest tests/integration/lite -v
tests/integration/lite/test_milvus_lite.py::TestMilvusLite::test_milvus_client_with_local_db_path PASSED [ 25%]
tests/integration/lite/test_milvus_lite.py::TestMilvusLite::test_milvus_lite_insert_search        PASSED [ 50%]
tests/integration/lite/test_milvus_lite.py::TestMilvusLite::test_milvus_lite_multiple_clients_same_db PASSED [ 75%]
tests/integration/lite/test_milvus_lite.py::TestMilvusLite::test_illegal_name                     PASSED [100%]
============================== 4 passed in 4.48s ==============================
```

So the Lite server does start, serve, and answer on Windows under milvus-lite 3.0 — the four tests that CI has been collecting-then-skipping on the `windows-2022` matrix leg now genuinely exercise the path this PR enables.

(The `AllocTimestamp … Method not implemented!` lines in the log are pre-existing and unrelated — milvus-lite does not implement that RPC and the client falls back; the same lines appear on Linux.)

Issue #3768, opened 2026-08-23
[Bug]: check_invalid_binary_vector raises TypeError instead of returning False for non-sized values


lokeshramchand-ctrl:
### Describe the bug

`check_invalid_binary_vector` in `pymilvus/client/utils.py` raises an unhandled `TypeError` instead of returning `False` when a `BINARY_VECTOR` entity's `values` list contains an element that has no `len()` (e.g. `None`, an `int`, or any other non-sized value).

The function checks the per-row length before checking the type:

```python
def check_invalid_binary_vector(entities: List) -> bool:
    for entity in entities:
        if entity["type"] == DataType.BINARY_VECTOR:
            if not isinstance(entity["values"], list) or len(entity["values"]) == 0:
                return False

            dim = len(entity["values"][0]) * 8
            if dim == 0:
                return False

            for values in entity["values"]:
                if len(values) * 8 != dim:          # len() called first
                    return False
                if not isinstance(values, bytes):   # type check comes too late
                    return False
    return True
```

`len(values)` is evaluated before `isinstance(values, bytes)`, so any row element without `__len__` blows up with a raw `TypeError` before the type check ever runs.

This function's entire job (per the intent behind the earlier fix in #3694) is to return `False` cleanly so the caller can raise a proper `ParamError`. The callers are:
- `pymilvus/client/grpc_handler.py:1094`
- `pymilvus/client/grpc_handler.py:1228`
- `pymilvus/client/async_grpc_handler.py:893`

All do:
```python
if not check_invalid_binary_vector(entities):
    raise ParamError(message="Invalid binary vector data exists")
```

So instead of a clean `ParamError("Invalid binary vector data exists")`, the caller gets a raw, confusing `TypeError: object of type 'NoneType' has no len()` out of a validation helper.

Note: this is a distinct, unfixed gap left over from #3694, which fixed the empty-list case (`and` → `or`) in the same function but didn't address the length-before-type-check ordering. The existing regression test `test_non_bytes_values` in `tests/unit/test_utils.py` only exercises nested lists as the "invalid" values (which still have `__len__`), so it never caught this.

### Expected Behavior

`check_invalid_binary_vector` should return `False` for any `BINARY_VECTOR` entity whose `values` contains a non-`bytes` element, regardless of whether that element supports `len()`, so callers can raise the intended `ParamError` instead of crashing with a `TypeError`.

### Steps/Code To Reproduce behavior

```python
[...]

PR #3769, opened 2026-08-23
fix: check_invalid_binary_vector should not raise on non-sized values

lokeshramchand-ctrl:
## Problem

`check_invalid_binary_vector` raises an unhandled `TypeError` instead of returning `False` when a `BINARY_VECTOR` entity's `values` list contains an element with no `__len__` (e.g. `None`, an `int`):

```python
>>> check_invalid_binary_vector([{"type": DataType.BINARY_VECTOR, "values": [b"\x00\x01", None]}])
TypeError: object of type 'NoneType' has no len()
```

The per-row loop calls `len(values)` before checking `isinstance(values, bytes)`:

```python
for values in entity["values"]:
    if len(values) * 8 != dim:        # len() evaluated first
        return False
    if not isinstance(values, bytes):  # too late — never reached on non-sized input
        return False
```

## Impact

This function's whole job is to return `False` so the caller can raise a clean error. `grpc_handler.py:1094` (and `:1228`, plus the async handler) does:

```python
if not check_invalid_binary_vector(entities):
    raise ParamError(message="Invalid binary vector data exists")
```

So instead of `ParamError("Invalid binary vector data exists")`, the user gets a raw `TypeError` out of the validation helper. Reachable from a malformed binary-vector insert, e.g. `Collection.insert(data=[[b"\x00\x01", None]])`.

This is a sibling gap to the one fixed in #3694 (`and` → `or` for the empty-list case), in the same function, that survived that fix. The existing `test_non_bytes_values` test only exercises nested *lists* as the invalid case, which still have `__len__`, so it never caught a genuinely non-sized element.

Filed as #3768.

## Fix

Swap the two checks: verify `isinstance(values, bytes)` before calling `len(values)`.

```python
for values in entity["values"]:
    if not isinstance(values, bytes):
        return False
    if len(values) * 8 != dim:
        return False
```

## Testing

Added `test_non_sized_value` to `TestCheckInvalidBinaryVector` in `tests/unit/test_utils.py`, alongside the existing `test_inconsistent_dimensions` / `test_non_bytes_values` / `test_empty_values` cases.

- `uv run pytest tests/unit/test_utils.py -q` → 71 passed
- `uv run pytest tests/unit/ -q` → 4719 passed, 3 skipped, 1 failed (`tests/unit/test_check.py::TestGetCommit::test_get_commit`, a pre-existing unrelated failure — same one called out in #3694's own PR description — verified to fail identically on `master` before this change)
- Manually verified the repro above raises `TypeError` before the fix and returns `False` after

---

This change was AI-assisted (Claude). I found the bug while doing a broader audit for unclaimed defects, reproduced the `TypeError` directly, traced the call sites in `grpc_handler.py` / `async_grpc_handler.py` to confirm the intended contract, and ran the full unit suite plus the baseline comparison above myself.


pymilvus/client/utils.py
```diff
@@ -164,10 +164,10 @@ def check_invalid_binary_vector(entities: List) -> bool:
                 return False
 
             for values in entity["values"]:
-                if len(values) * 8 != dim:
-                    return False
                 if not isinstance(values, bytes):
                     return False
+                if len(values) * 8 != dim:
```
yhmo, line 172: pymilvus/client/utils.py line:169
High ---- The reordering only protects non-sized elements that appear after the first one. `dim` is still computed from `entity["values"][0]` at line 162, before this loop runs, so `values=[None, b"\x00\x01"]`, `values=[1, 2]`, or `values=[None]` still raise `TypeError: object of type 'NoneType' has no len()` instead of returning `False`. This is reachable end-to-end through the same path this PR fixes: `Collection.insert(data=[[None, b"\x01"]])` passes `check_insert_schema` and `prepare_data` (binary-vector values are passed through raw at pymilvus/orm/prepare.py:152), so the raw `TypeError` still escapes instead of the intended `ParamError` described in #3768. Suggest guarding the first element before computing `dim`, e.g. `if not isinstance(entity["values"][0], bytes): return False`, and add a regression test with a non-sized first element.

tests/unit/test_utils.py
```diff
@@ -259,6 +259,10 @@ def test_empty_values(self):
         entities = [{"type": DataType.BINARY_VECTOR, "values": []}]
         assert utils.check_invalid_binary_vector(entities) is False
 
+    def test_non_sized_value(self):
+        entities = [{"type": DataType.BINARY_VECTOR, "values": [b"\x00\x01", None]}]
+        assert utils.check_invalid_binary_vector(entities) is False
```
yhmo, line 264: tests/unit/test_utils.py line:264
Medium ---- This regression test only covers a non-sized element in the second position, where the first element (`b"\x00\x01"`) is a valid `bytes` value used to compute `dim` at utils.py:162. A non-sized *first* element (e.g. `[None, b"\x00\x01"]`) or an all-non-sized list (e.g. `[1, 2]` or `[None]`) still raises `TypeError` at the dim computation before the loop, so this test passes while the bug class from #3768 is only partially fixed. Adding a first-position non-sized case would have caught the remaining gap.

pymilvus/client/utils.py
lokeshramchand-ctrl: Good catch — fixed in 132ee011 by guarding `entity["values"][0]` before computing `dim`, so a non-sized first element now returns `False` instead of raising. Added regression tests for both a non-sized first element and an all-non-sized list.

tests/unit/test_utils.py
lokeshramchand-ctrl: Added test_non_sized_first_value and test_all_non_sized_values in 132ee011 to cover this gap.

lokeshramchand-ctrl:
Review this and let me know if any edits required - @tedxu @XuanYang-cn

PR #3779, opened 2026-08-29
fix: refresh loads in async client

### LINKED ISSUE #3772: [Bug]: AsyncMilvusClient.refresh_load() never triggers a refresh, so optimize_collection() leaves stale segments loaded

mikemikimike:
Closes #3772

`AsyncMilvusClient.refresh_load()` currently delegates to the low-level progress query instead of requesting a load refresh. As a result, async callers and `optimize_collection()` can return without reloading refreshed segments.

This change reuses the existing async load paths: collections call `load_collection(..., _refresh=True)`, while partition refreshes call `load_partitions(..., _refresh=True)`. Regression tests verify both routes and ensure the progress-only handler is not used.

Validation: async client unit tests passed (81); Ruff format/check and `git diff --check` passed. A targeted mypy run is blocked by existing missing grpc/milvus-lite stubs and generated protobuf stub syntax errors.

This contribution was prepared with assistance from an AI agent.


linhongyu510:
I checked out this PR in an isolated worktree and ran the complete async client operations test module with the PR source forced on `PYTHONPATH`:

```text
pytest tests/unit/test_async_milvus_client_ops.py -q
81 passed, 3 warnings
```

Both new regression cases pass:

- collection refresh delegates to `load_collection(..., _refresh=True)`;
- partition refresh delegates to `load_partitions(..., _refresh=True)`.

I also verified that the progress-only handler is not used by either path. I did not find a correctness issue in the changed scope.

mikemikimike:
I rechecked this PR for issue #3772. The PR is linked to the issue, and there is no human changes-requested review or failing check reported for the current head. Could a maintainer please review whether the change fully resolves the issue and approve it, or advise on any remaining work?

PR #3783, opened 2026-09-05
fix(async): preserve optimization after result wait timeout

Kuang-xianxin:
## Problem

`AsyncOptimizeTask.result(timeout=...)` passes its owned background task directly to `asyncio.wait_for`. A short wait cancels the optimization, affects other result waiters, and prevents a later retry. For the preview `optimize(wait=False)` API this can stop the client workflow during compaction before index rebuild and load refresh.

Cancellation also needs to distinguish ownership: a caller holding an `AsyncOptimizeTask` can wait again or explicitly cancel it, while `optimize(wait=True)` does not return a handle when it exits early.

## Change

- Shield the background task in both timed and untimed `result()` waits. A local wait timeout raises the existing timeout error; caller cancellation propagates as `CancelledError`, allowing standard `asyncio.wait_for` and `asyncio.timeout` wrappers to work without cancelling the shared optimization.
- Keep explicit `task.cancel()` behavior and the optimization's own timeout forwarding.
- In `optimize(wait=True)`, cancel and await unfinished owned work before returning a wait timeout or propagating caller cancellation. This prevents detaching an inaccessible task and retrieves its cancellation exception.
- Document the ownership distinction. Server-side compaction cancellation behavior is unchanged.

Regression coverage includes later success/failure after a short wait, concurrent waiters, explicit task cancellation, external timeout wrappers, and zero/positive timeouts or cancellation of `optimize(wait=True)`. A public `optimize(wait=False)` workflow test uses mocked RPC responses and verifies load refresh after a timed-out result wait.

## Validation

Python 3.13.14 on Windows:

- Original source: 5 initial regression cases failed.
- Before the review revision: 9 additional cancellation/ownership regressions failed on d22d7af.
- Latest revision: **148 passed** with:

```text
uv run --no-sync pytest tests/unit/test_async_optimize_task.py tests/unit/test_async_milvus_client_ops.py tests/unit/test_optimize_task.py -q -o log_cli=false --tb=short
```

- Ruff and Black checks pass on all four changed files.
- The `asyncio.timeout` context test is skipped on Python versions before 3.11; `wait_for` cases cover older versions.
- No live Milvus server or performance benchmark was used; this is a client task-lifecycle fix.

Found through source inspection. Searches did not find a matching optimization/result-timeout fix; the existing async refresh-load PRs concern separate RPC options.

### DIFF
--- a/pymilvus/milvus_client/async_optimize_task.py
+++ b/pymilvus/milvus_client/async_optimize_task.py
@@ -84,18 +84,23 @@ class AsyncOptimizeTask:
             raise MilvusException(message="Optimization task was cancelled")
 
     async def result(self, timeout: Optional[float] = None) -> OptimizeResult:
-        """Wait for the result without cancelling optimization when the wait times out."""
+        """Wait without cancelling optimization on wait timeout or caller cancellation.
+
+        Call cancel() to stop the background optimization explicitly.
+        """
         if not self._task:
             raise MilvusException(message="Task has not been started")
 
         try:
             if timeout is not None:
                 return await asyncio.wait_for(asyncio.shield(self._task), timeout=timeout)
-            return await self._task
+            return await asyncio.shield(self._task)
         except asyncio.TimeoutError as e:
             raise MilvusException(message="Timeout waiting for optimization to complete") from e
         except asyncio.CancelledError as e:
-            raise MilvusException(message="Optimization task was cancelled") from e
+            if self._task.cancelled():
+                raise MilvusException(message="Optimization task was cancelled") from e
+            raise
 
     def set_progress(self, stage: ProgressStage) -> None:
         if self._cancelled:
--- a/pymilvus/milvus_client/async_optimize_task.py
+++ b/pymilvus/milvus_client/async_optimize_task.py
@@ -84,12 +84,13 @@ class AsyncOptimizeTask:
             raise MilvusException(message="Optimization task was cancelled")
 
     async def result(self, timeout: Optional[float] = None) -> OptimizeResult:
+        """Wait for the result without cancelling optimization when the wait times out."""
         if not self._task:
             raise MilvusException(message="Task has not been started")
 
         try:
             if timeout is not None:
-                return await asyncio.wait_for(self._task, timeout=timeout)
+                return await asyncio.wait_for(asyncio.shield(self._task), timeout=timeout)
             return await self._task
         except asyncio.TimeoutError as e:
             raise MilvusException(message="Timeout waiting for optimization to complete") from e
--- a/pymilvus/milvus_client/async_milvus_client.py
+++ b/pymilvus/milvus_client/async_milvus_client.py
@@ -2,6 +2,7 @@ import asyncio
 import copy
 import time
 import types
+from contextlib import suppress
 from typing import Dict, List, Optional, Type, Union
 
 from pymilvus.client import type_info
@@ -2310,6 +2311,7 @@ class AsyncMilvusClient(BaseMilvusClient):
                 If not provided, uses system default.
             wait (bool): Whether to wait for optimization to complete. Defaults to True.
                 If False, returns an OptimizeTask for async tracking.
+                If True, a wait timeout or caller cancellation stops the client-side task.
             timeout (Optional[float]): Maximum time in seconds to wait for optimization.
                 Only applies when wait=True.
             **kwargs: Additional arguments.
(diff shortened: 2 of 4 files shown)


pymilvus/milvus_client/async_optimize_task.py
```diff
@@ -84,12 +84,13 @@ def check_cancelled(self) -> None:
...
 
     async def result(self, timeout: Optional[float] = None) -> OptimizeResult:
+        """Wait for the result without cancelling optimization when the wait times out."""
         if not self._task:
             raise MilvusException(message="Task has not been started")
 
         try:
             if timeout is not None:
-                return await asyncio.wait_for(self._task, timeout=timeout)
+                return await asyncio.wait_for(asyncio.shield(self._task), timeout=timeout)
```
yhmo, line 96: pymilvus/milvus_client/async_optimize_task.py line:93
Medium ---- With wait=True the caller no longer has a handle after a wait timeout: optimize(wait=True, timeout=X) raises "Timeout waiting..." while the detached background task keeps running, cannot be cancelled or re-awaited, and typically ends with its own remaining_timeout()=0 exception that nobody retrieves ("Task exception was never retrieved" at GC). Was losing the old abort-on-timeout behavior for wait=True intended? Consider cancelling the task on that path or keeping the task reachable/documenting the detach.

pymilvus/milvus_client/async_optimize_task.py
```diff
@@ -84,12 +84,13 @@ def check_cancelled(self) -> None:
             raise MilvusException(message="Optimization task was cancelled")
 
     async def result(self, timeout: Optional[float] = None) -> OptimizeResult:
+        """Wait for the result without cancelling optimization when the wait times out."""
```
yhmo: pymilvus/milvus_client/async_optimize_task.py line:87
Low ---- Follow-up: only the timed branch shields the background task. The timeout=None branch (line 94) still awaits self._task directly, so a caller cancellation - e.g. `await asyncio.timeout(X): await task.result()` or `asyncio.wait_for(task.result(), X)` - still cancels the optimization, and result() converts the caller CancelledError into a MilvusException. The exact failure this PR fixes is therefore still reachable through the standard asyncio timeout wrappers; consider shielding both branches.

pymilvus/milvus_client/async_optimize_task.py
Kuang-xianxin: Addressed in 258e5d1. `optimize(wait=True)` now cancels and awaits unfinished owned work in `finally`, retrieving the task's cancellation exception while preserving the original timeout or caller cancellation. This keeps the no-handle path from detaching the optimization. The behavior is documented on `wait`.

New regressions cover zero/positive wait timeouts and caller cancellation with/without a result timeout; they verify that the owned task is done and cancelled before `optimize` exits. Together with the waiter-isolation cases, 9 new cases failed on d22d7af; all 148 related tests pass after the fix. Ruff and Black pass on all four changed files. No live Milvus server was used.

pymilvus/milvus_client/async_optimize_task.py
Kuang-xianxin: Addressed in 258e5d1. Both branches now shield the background task. `result()` propagates cancellation of its caller as `CancelledError`, so `asyncio.wait_for(task.result(), ...)` and `asyncio.timeout(...)` produce their normal timeout result while the optimization remains available to other waiters. Explicit cancellation of the optimization still reports the existing Milvus cancellation error.

The new regressions cover direct caller cancellation and `wait_for` around both timed/untimed `result()`, plus the Python 3.11+ timeout context. They verify subsequent successful retrieval and isolation of another waiter. These cases failed on the previous head and pass in the 148-test related suite.

pymilvus/milvus_client/async_optimize_task.py
yhmo: The issue is confirmed fixed. Resolving this conversation.

pymilvus/milvus_client/async_optimize_task.py
yhmo: The issue is confirmed fixed. Resolving this conversation.

Kuang-xianxin:
Merged current master in 1a2b0d7 and resolved the import conflict, preserving the new segment/compaction APIs and this PR's optimization ownership behavior.

The 148 related async optimization, async client and synchronous optimization tests pass on Windows. Ruff and Black pass on all four PR files. No live Milvus server was used.

PR #3784, opened 2026-09-07
enhance: support custom HTTP headers in bulk import APIs

bigsheeper:
Milvus keeps adding request features carried in HTTP headers. The first is `Idempotency-Key` (milvus-io/milvus#52544): a retry carrying the same key resolves to the original import job instead of creating a second one. Rather than teach the SDK each header by name, this PR lets callers pass any of them, so the next header-driven feature needs no SDK change.

## Changes

- `bulk_import`, `get_import_progress`, `commit_import`, `abort_import` and `list_import_jobs` accept `headers: Optional[Dict[str, str]]`. It is merged over the SDK defaults (User-Agent, Accept, Authorization, DB-Name), so a caller's value wins for the same name. `None` or `{}` adds nothing; a call without it is byte-identical to today.
- The `bulk_import` docstring shows the `Idempotency-Key` example and the server-side constraints (at most 256 bytes of printable ASCII, else error 1100).
- `pymilvus.bulk_writer.IDEMPOTENCY_KEY_HEADER` exports the header name, so it gets autocomplete instead of being retyped at every call site; the raw string works too.

```python
bulk_import(url=url, collection_name="events", files=files,
            headers={"Idempotency-Key": "nightly-2026-09-07-batch-3"})
```

## Scope

Earlier revisions added a named `idempotency_key` parameter and threaded it through `MilvusClient` / `AsyncMilvusClient` `insert` / `upsert` / `delete` as gRPC metadata. Both are dropped. The named parameter would need a new SDK release for every future header; the DML surface would carry three behaviours behind one parameter (the server acts on the key only for `Insert`, milvus-io/milvus#50007, and only when idempotent write is enabled globally and on the collection). A generic per-call gRPC metadata passthrough is a separate follow-up.

## Tests

- `tests/unit/test_bulk_import.py`: header merge and override at `_http_headers`, passthrough at `_post_request`, end-to-end at `bulk_import` with and without headers, and one case per remaining entry point.
- Full `tests/unit` run: 4724 passed; `ruff` and `black` clean.


czs007:
> **Must-fix issues introduced by this PR: 0**
> **Merge recommendation:** Mergeable as-is — this review found no must-fix issue introduced by this PR.

Adversarial review found no issues requiring changes.

**Verified:**
- `pymilvus/client/call_context.py:19-27` — the `idempotency-key` header is appended only when the key is non-empty, so keyless calls emit exactly the same three metadata tuples as before this PR.
- `pymilvus/orm/connections.py:592-596` — `kwargs.get("idempotency_key", "")` mirrors the `client_request_id` precedent; the key stays in `**kwargs` and every downstream terminal (`Prepare.delete_request`, `describe_collection`, `MutationFuture`) absorbs it, so no `TypeError` is possible.
- `pymilvus/client/grpc_handler.py:972-976` — `insert_rows` (and likewise `upsert_rows` at `:1315`, `batch_insert` at `:1101`, `delete` at `:1149`) issues a single RPC per call, so no multiple chunks ever share one key.
- `pymilvus/milvus_client/milvus_client.py:885-907` — delete-by-ids builds the call context twice but both carry the same key and only one `Delete` RPC is sent.
- `pymilvus/client/bulk_import.py:76-81` — `_post_request` consumes the key via `pop` into the HTTP header only; `tests/unit/test_bulk_import.py:216-217` asserts it never reaches the JSON body or lingers in kwargs.
- `tests/unit/test_idempotency_key.py` — `_context_of` reads the real `context=` keyword argument, matching the call shape at `milvus_client.py:280`, so the sync/async insert/upsert/delete assertions exercise the actual forwarding path.

## Low — non-blocking

- `pymilvus/client/grpc_handler.py:1041` — The same `CallContext` that carries `idempotency-key` is passed into schema resolution (`insert_rows` → `_prepare_row_insert_request` → `_get_schema` at `:1000-1002`, `upsert_rows` at `:1280-1282`, ORM columnar `batch_insert` at `:1071-1073`, `MilvusClient.delete(ids=…)` at `milvus_client.py:884-889`, and the async handler at `async_grpc_handler.py:707-709`). On a schema cache miss, `describe_collection` sends `_api_level_md(context)` at `:820-822`, so `DescribeCollection` goes out with the same `idempotency-key` as the following `Insert`/`Upsert`/`Delete`. This contradicts the PR description's statement that the key is gRPC metadata for that DML only. Server-side dedup is keyed by `(messageType, scopedKey)` and only broadcast write requests enter the index, and the proxy `DescribeCollection` path never reads this metadata, so no incorrect behavior is reproducible today; this is a contract/consistency gap rather than a defect. Suggestion: derive a key-less `CallContext` for the schema-resolution path (or strip the field inside `_get_schema`), and add a negative assertion that `DescribeCollection` metadata contains no `idempotency-key` on a cache miss. (raised by xiaocai2333)
- `tests/unit/test_idempotency_key.py:140` — The ORM entry points are only covered at the `connections._generate_call_context` layer (`:140-143`); nothing asserts that `Collection.insert` / `upsert` / `delete` → `_get_connection(**kwargs)` (`collection.py:557/635/695`) → `_generate_call_context` (`:194`) → handler actually delivers a context carrying the key. The chain is correct on reading, but `collection.py` already has several `_get_connection()` calls without kwargs (`:263/:309/:1219/:1361`), and a future change of that kind on the DML methods would go uncaught. Suggestion: mirror `TestMilvusClient` with a mock handler and assert the context passed to `insert_rows` / `batch_insert` / `upsert_rows` / `delete` contains the key. (raised by czs007, xiaocai2333)
- `tests/unit/test_idempotency_key.py:50` — All assertions are positive checks on the DML context, and both the sync fixture (`:50`) and the async fixture (`:96`) mock `_get_schema` outright, so `describe_collection` is never reached and the leak above is invisible to the suite. Suggestion: add one test that uses a real `GrpcHandler` (or mocks only `_stub`), clears `GlobalCache.schema` to force a cache miss, and asserts `_stub.DescribeCollection` metadata lacks `idempotency-key` while `_stub.Insert` metadata carries it. That test fails on the current code and pairs with the first item. (raised by czs007, xiaocai2333)

tinswzy:
Reviewed against the server side of milvus-io/milvus#50007 and #52544. The plumbing is
right, and one thing that matters most is right: `context` is built once in the client
method and `@retry_on_rpc_failure()` sits on `insert_rows`, so every retry re-emits the
same key. That is the property the whole feature rests on.

The wire name matches the server constants exactly, and empty-vs-absent is handled the way
the server reads it. What follows is about what the parameter promises versus what the
server does with it.

## Must fix

**1. `insert` with a key is rejected on a default-configured server — it is not ignored.**

```go
// task_insert.go:192 -> task_insert_idempotency.go:65
if it.idempotencyKey != "" && !globalIdempotencyEnabled {
    return merr.WrapErrParameterInvalidMsg("idempotency key is not accepted when idempotent write is globally disabled")
}
if it.idempotencyKey != "" && !collectionIdempotencyEnabled {
    return merr.WrapErrParameterInvalidMsg("idempotency key is not accepted when collection idempotent write is disabled")
}
```

Both gates default off: `streaming.idempotency.enabled` is `false`, and the collection
property is unset. So adding this argument to working code makes the write start failing
with error 1100. The docstring should say both switches are required and that sending a key
early is an error, not a no-op. (milvus-docs#3635 already states this — worth matching.)

**2. `upsert` and `delete` docstrings should say the server does not act on the key.**

`GetIdempotencyKeyFromContext` has exactly one caller, `Proxy.Insert`. Upsert and delete
never read it. Today the same parameter has three behaviours — insert errors when disabled,
upsert and delete silently drop it — behind three docstrings that read alike. Someone who
sets a key on `delete` and retries gets a second delete with no signal.

## Worth fixing

**3. Nothing validates the key before it reaches the server.** `ValidateIdempotencyKey`
requires printable ASCII (0x20–0x7E) and a length bound (`streaming.idempotency.maxKeyLength`,
256 by default). A Python caller passing a non-ASCII string — a UUID is fine, a batch name
with a non-Latin character is not — gets a 1100 from deep in the server. A few lines client
side would give a better error.

**4. `idempotency_key` is kwargs-only, so a misspelling silently disables idempotency.**
`idempotencyKey=`, `idempotency_id=`, or `**{"idempotency-key": ...}` are all absorbed by
`**kwargs` at both ends — `_generate_call_context` does `.get("idempotency_key", "")` and
`insert_rows(**kwargs)` swallows the rest. No error, and the write is simply not idempotent.

`client_request_id` has the same shape, but the consequences differ: a lost request id costs
a log correlation, a lost idempotency key costs duplicate rows. Making it an explicit keyword
parameter would not stop `**kwargs` from swallowing a typo — that is inherent here — but it
would put the right name in autocomplete, `help()` and the signature, which is where the
name is most likely to be gotten right.

Relatedly: `_generate_call_context` accepts both `client_request_id` and `client-request-id`
but only the underscore form for the new parameter. Rather than accepting both, consider
raising on `idempotency-key` in kwargs — readers meet that spelling as the wire name and
copying it currently fails silently.

**5. The Go SDK and this PR disagree on `upsert`.** `client/milvusclient/write_options.go`
rejects a key on Upsert client-side (`"idempotency key is only supported for Insert"`), while
this sends it and the server drops it. Two official SDKs, opposite contracts for one
operation. Silently dropping is the weakest of the three options; worth a maintainer call
rather than settling it in either PR alone.

## Optional

**6.** `_generate_call_context(**kwargs)` appears 116 times in the sync client and 118 in the
async one, so `search(..., idempotency_key=...)` and `create_collection(...)` also emit the
metadata — undocumented and untested. On the DDL path #52544's broadcaster does act on a key,
so this is either a feature or a surprise; worth deciding deliberately.

**7.** Only async `upsert` got the docstring note; async `insert` and `delete` did not, though
the description covers all three.

tinswzy:
Correcting one thing I got wrong in the review above, in point 3.

I wrote that a non-ASCII key "gets a 1100 from deep in the server". That is not how it fails.
`ValidateIdempotencyKey` — the function carrying the printable-ASCII rule — runs at the REST
door and in the inter-component propagation interceptor, and neither is on the path a gRPC
client takes into `Proxy.Insert`. The proxy's own check there covers length only.

The charset constraint is real, but it is gRPC's: a metadata value outside printable ASCII is
refused by the transport, which is why the server-side validator exists for the REST door at
all — HTTP accepts bytes the gRPC hop then cannot carry. So a pymilvus caller passing a
non-ASCII key gets a transport-level failure from grpcio, not a Milvus parameter error.

The recommendation is unchanged, and if anything stronger for it: a few lines of client-side
validation turn an opaque transport error into one that names the parameter. The length bound
is still enforced server-side and still worth checking early.

Everything else in that review stands, and the two must-fix items are unaffected.

bigsheeper:
@tinswzy Thanks for the review, and for the follow-up on point 3. Verified the grpcio side locally: a non-ASCII value fails with `ValueError: metadata was invalid` at call start, before anything reaches the server.

I have narrowed this PR to bulk import only (b1966bd1). The DML surface (`insert` / `upsert` / `delete` carrying `idempotency-key` gRPC metadata) is dropped: as points 1, 2 and 5 show, one parameter would carry three behaviours across the three methods and disagree with the Go SDK on upsert, so it needs a settled contract first. I will bring it back in a follow-up PR with docstrings covering the two server switches, an explicit keyword parameter, and client-side ASCII/length validation (points 1-4, 7).

What remains is `bulk_import(idempotency_key=...)` -> `Idempotency-Key` header. It is an explicit parameter, and the REST middleware validates charset and length and answers with 1100 naming the parameter, so points 3, 4 and 6 do not apply here.

bigsheeper:
@czs007 Re the adversarial review: the three Low items (`DescribeCollection` sharing the key on a schema cache miss, ORM end-to-end coverage, the negative test) all concern the DML gRPC-metadata path. That path is removed in b1966bd1; this PR now only sends the `Idempotency-Key` HTTP header from `bulk_import`, so no gRPC metadata carries the key anymore. They will be picked up when the DML surface returns in a follow-up.

bigsheeper:
@tinswzy @czs007 One more revision (68221bde): the named `idempotency_key` parameter is gone too. The five bulk import REST functions now take a generic `headers: dict`, merged over the SDK defaults, so `Idempotency-Key` is `headers={"Idempotency-Key": ...}` and the next header-driven server feature needs no SDK change. The `bulk_import` docstring and `examples/bulk_import/example_bulkinsert_parquet.py` show the one-key-per-logical-import, reuse-on-retry pattern. PR title and description updated accordingly.

bigsheeper:
Correction to the comment above: the example script change is reverted in 590086c5, since most callers do not need an idempotency key. The usage is shown only in the `bulk_import` docstring now.

czs007:
> **Must-fix issues introduced by this PR: 0**
> **Merge recommendation:** Mergeable as-is — this review found no must-fix issue introduced by this PR.

Adversarial review found no issues requiring changes.

Verified:
- `pymilvus/bulk_writer/bulk_import.py:18-34` — the extra `headers` argument is merged last into a fresh local dictionary, so caller values override defaults and the `None`/`{}` cases return the exact prior header set without mutating shared state.
- `pymilvus/bulk_writer/bulk_import.py:76-87` — `headers` is popped from `kwargs` before forwarding to `requests.post`, so it cannot be double-passed as a keyword argument or leak into the JSON body.
- `pymilvus/bulk_writer/bulk_import.py:91-197` — all five public entry points (`bulk_import`, `get_import_progress`, `commit_import`, `abort_import`, `list_import_jobs`) forward `headers=headers` explicitly as a named parameter, while preserving their endpoint-specific payloads and DB-name handling.
- `tests/unit/test_bulk_import.py:35-46,85-99` — tests cover header addition, default override, `None`, and empty dictionaries, including equality against the no-arg form for the backward-compatibility claim.
- `tests/unit/test_bulk_import.py:206-239,362-388` — parametrized tests exercise every public entry point and assert that custom headers are sent as HTTP headers and are absent from the request JSON.

No findings survived verification in this round. The three Low items raised in an earlier round (idempotency-key sharing with `DescribeCollection` in `grpc_handler.py`, missing ORM end-to-end assertions, and the missing negative assertion for non-DML RPCs) all concerned the gRPC-metadata DML path, which the author removed in commits `b1966bd1` and `68221bde`. That removal resolves them, so they are not re-raised here.

PR #3788, opened 2026-09-08
fix: check type before len() in check_invalid_binary_vector to avoid TypeError

### LINKED ISSUE #3768: [Bug]: check_invalid_binary_vector raises TypeError instead of returning False for non-sized values

linhongyu510:
## Summary

Fixes #3768.

`check_invalid_binary_vector` in `pymilvus/client/utils.py` raises an unhandled `TypeError` instead of returning `False` when a `BINARY_VECTOR` entity's `values` list contains an element that has no `len()` (e.g. `None`, `int`, or any other non-sized value).

## Root Cause

The function calls `len()` before checking `isinstance(values, bytes)` in two places:

1. **Line 162**: `dim = len(entity["values"][0]) * 8` — called unconditionally on the first element
2. **Line 167**: `len(values) * 8 != dim` — called in the per-row loop before the type check on line 169

When a non-sized value (like `None` or `int`) is present, `len()` raises `TypeError` instead of the function returning `False`.

## Fix

Swap the order of checks in both locations: check `isinstance(values, bytes)` before calling `len(values)`.

## Changes

- `pymilvus/client/utils.py`: 4 insertions, 2 deletions


linhongyu510:
Added `test_non_sized_values` as requested, and rebased onto current `master` (the branch was well behind). New head `4a5a8abe` — the diff is now just the two intended files.

Your reasoning about `test_non_bytes_values` was exactly right, and I verified it rather than taking it on faith. With the source reverted to `master` and only the new test applied:

```
FAILED tests/unit/test_utils.py::TestCheckInvalidBinaryVector::test_non_sized_values
E   TypeError: object of type 'NoneType' has no len()
pymilvus/client/utils.py:167: TypeError

1 failed, 6 passed
```

`test_non_bytes_values` passes in that same run — nested lists have `__len__`, so it never reaches the unguarded call. `None` is the value that does.

I kept both orderings from your snippet, and they are not redundant: `[b"\x00\x01", None]` hits the `len()` inside the loop, while `[None, b"\x00\x01"]` hits the separate `dim = len(entity["values"][0]) * 8` probe above it. That probe was a second instance of the same gap, and only the second case covers it.

With the fix in place: `tests/unit/test_utils.py` **71 passed**, `ruff check` and `ruff format --check` clean.

Issue #3790, opened 2026-09-09
Add collection compaction lifecycle observability APIs


MrPresent-Han:
## Motivation

PyMilvus E2E tests currently cannot combine retrieve results with authoritative evidence about retained segment lineage, the actual serving segment set, and every compaction task still active or retained for a collection.

## Required client support

- Allow persistent segment queries to select lifecycle states and expose partition, channel, storage version, level, and direct compaction sources.
- Provide a collection segment helper whose default state set covers Growing, Sealed, Flushing, Flushed, Importing, and Dropped.
- Expose the actual serving segment set through a clear client-level API.
- List retained collection compaction tasks with complete identifiers, state, type, sources, targets, and failure reason.
- Preserve existing manual compaction job-ID APIs and backward-compatible response fields.

## Scope

The APIs expose retained server state for polling and E2E assertions; they do not provide permanent task or dropped-segment history.

Dependencies: milvus-io/milvus-proto#667 and milvus-io/milvus#53307.

PR #3791, opened 2026-09-09
enhance: add compaction lifecycle observability APIs

MrPresent-Han:
issue: #3790

## Scope after protocol bump

#3797 is merged at `8bd86b706b3f94e3f269d8c330972da4f969fe6f`, and this PR is based on that master revision as one commit.
Diff changes seven SDK implementation files and seven unit-test files; no `pymilvus/grpc_gen` bindings or proto submodule updates.

Expose retained persistent-segment metadata and collection-scoped compaction-task metadata through existing RPCs; existing synchronous `list_loaded_segments()` remains query-serving segment API.

- Add lifecycle-state filtering to sync and async persistent-segment listing, plus sync and async `list_segments()` with shared retained-state default.
- Require `Optional[Sequence[SegmentState]]`; reject empty sequences, strings, raw integers, booleans, and invalid values before RPC construction.
- Add sync and async `list_compaction_tasks()` plus `AsyncGrpcHandler.get_compaction_tasks()`, while retaining synchronous job-scoped `get_compaction_plans()`.
- Share public `CompactionType`, `CompactionTaskState`, and `SegmentState` `IntEnum` definitions, default-state collection, and compaction-plan response parser across sync and async paths.
- Return enum instances for task type/state and `SegmentInfo.state`; unknown response integers fall back to zero-value enum members.
- Return task identity, source/target sets, failure reasons, partition/channel/lineage metadata, and collection context; database context is carried by RPC metadata.
- Preserve empty task outputs as `targets=[]` while retaining legacy positive `target` fallback.
- Initialize persistent `SegmentInfo` metadata directly through constructors in both clients while preserving legacy positional/keyword calls and `LoadedSegmentInfo` construction.

Proto dependency: milvus-io/milvus-proto#668 at `ae7fea6ab2f4e958f2feef0f0edb9a0d23fa7e0c` is inherited from master via #3797, and this PR does not modify those bindings.

Server dependency: milvus-io/milvus#53308; lifecycle filtering and collection-scoped task lookup require matching server support.

Dropped segments and terminal tasks remain subject to server retention/GC; callers needing audit history must collect observations within retention window.

## Validation

- 2,003 related unit tests passed on `3a2214a78ea3f5ae27bf7836bb580cc4ab3798e6`, covering sync/async handlers, request preparation, result types, and clients.
- Coverage includes retained defaults, invalid state rejection before RPC, enum known/unknown response decoding, task targets/failure reasons, selectors, metadata constructors, and async parity.
- Ruff, Black check, and diff whitespace checks passed.
- No server E2E rerun for client-side review fixes.

```bash
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. \
  python -m pytest -q -p no:cacheprovider -p pytest_asyncio.plugin \
  tests/unit/grpc_handler tests/unit/async_grpc_handler tests/unit/prepare \
  tests/unit/test_client_types.py tests/unit/test_types.py \
  tests/unit/test_milvus_client.py tests/unit/test_async_milvus_client.py
```


pymilvus/client/prepare.py
```diff
@@ -2498,6 +2510,16 @@ def get_compaction_state_with_plans(cls, compaction_id: int):
         request.compactionID = compaction_id
         return request
 
+    @classmethod
+    def get_compaction_tasks(cls, collection_name: str, db_name: str = ""):
```
liliu-z, line 2538: `Prepare.get_compaction_tasks` (prepare.py:2514) leaves `compactionID` (field 1) unset and carries the selector only in the new `db_name`/`collection_name` fields (2 and 3), which servers predating milvus#53308 discard as unknown fields; datacoord then executes `getCompactionInfo(triggerID=0)`, which returns `Status: Success`, `state: Completed`, `mergeInfos: []` (`internal/datacoord/compaction_inspector.go:144-148`). `list_compaction_tasks()` therefore reports "no compaction tasks, all completed" for every collection on every released Milvus, and the caller gets no error, warning, or field that distinguishes an unsupported server from a genuinely empty result — a polling loop such as `while client.list_compaction_tasks(c).plans:` exits immediately. The new `states` filter has the same failure mode: field 4 is dropped by older servers, so `list_segments(c, states=["Dropped"])` returns the unfiltered segment set as if it had been filtered.

MrPresent-Han: Thanks for pointing this out. These new observability APIs require the corresponding server support in milvus-io/milvus#53308; using collection-scoped task listing or segment-state filtering against older servers is outside the supported scope. We acknowledge that an empty or otherwise successful response from an older server does not establish the requested semantics, but we do not plan to add capability detection or compatibility fallbacks in this PR.

pymilvus/client/grpc_handler.py
```diff
@@ -99,6 +99,60 @@
...
+            merge.target,
+            plan_id=merge.plan_id,
+            trigger_id=merge.trigger_id,
+            collection_id=merge.collection_id,
+            partition_id=merge.partition_id,
+            channel=merge.channel,
+            compaction_type=_COMPACTION_TYPE_NAMES.get(merge.type, f"unknown({merge.type})"),
+            state=_COMPACTION_TASK_STATE_NAMES.get(merge.state, f"unknown({merge.state})"),
+            failure_reason=merge.failure_reason,
+            targets=list(merge.targets) or None,
```
liliu-z, line 150: `_parse_compaction_plans` collapses an empty `merge.targets` to `None` at `grpc_handler.py:150`, and `Plan.__init__` then falls back to `[target] if target else []` at `types.py:337`. The server sets `Target = -1` whenever `task.GetResultSegments()` is empty (`internal/datacoord/util.go:321-324`), and `-1` is truthy in Python, so every executing, pipelining, failed and timed-out task is returned with `targets == [-1]` — a segment ID that does not exist — and `if plan.targets:` evaluates True for tasks that have produced no output. Both `get_compaction_plans` and `list_compaction_tasks` are affected because they share the parser.

pymilvus/client/grpc_handler.py
```diff
@@ -2224,9 +2281,14 @@ def get_persistent_segment_infos(
         collection_name: str,
         timeout: Optional[float] = None,
         context: Optional[CallContext] = None,
+        states: Optional[Iterable[Union[int, str]]] = None,
```
liliu-z: `get_persistent_segment_infos` is annotated `states: Optional[Iterable[Union[int, str]]]` and builds its request inside the `@retry_on_rpc_failure()` boundary, while the decorator re-invokes `func(*args, **kwargs)` in a loop at `decorators.py:419`. A generator passed as `states` is exhausted building the first request, so any retry sends `states=[]`, which the server resolves to its default Flushing/Flushed/Sealed filter (`internal/proxy/impl.go:4345`) instead of the requested states — the call returns Success with the wrong segment set and no warning. Verified: `states=iter(["Dropped"])` serializes to `[6]` on the first build and `[]` on the second.

pymilvus/client/prepare.py
```diff
@@ -2292,8 +2292,17 @@ def get_collection_stats_request(cls, collection_name: str):
...
+    def get_persistent_segment_info_request(
+        cls,
+        collection_name: str,
+        states: Optional[Iterable[Union[int, str]]] = None,
+        db_name: str = "",
+    ):
+        return milvus_types.GetPersistentSegmentInfoRequest(
+            dbName=db_name,
+            collectionName=collection_name,
+            states=states or [],
```
liliu-z: `states=states or []` hands the value straight to the repeated enum field. Because the annotation is `Iterable[Union[int, str]]`, a bare `str` type-checks, so `states="Flushed"` makes protobuf iterate the characters and raise `ValueError: unknown enum label "F"`; an invalid label raises a bare `ValueError` that escapes the `MilvusException` contract, unlike `Prepare.get_compaction_tasks` at `prepare.py:2515` in this same diff which raises `ParamError`; and `states=[99]` is accepted silently and sent to the server, which matches nothing and returns an empty list.

pymilvus/client/grpc_handler.py
MrPresent-Han: Fixed in debfd411: states now accepts only a repeatable Sequence across sync and async paths, while bare strings/bytes and one-shot or other non-sequence iterables are rejected with ParamError before the RPC. Added sequence-reuse, generator-rejection, and no-RPC coverage; the related suite passes 1,988 tests.

pymilvus/client/prepare.py
MrPresent-Han: Fixed in debfd411: every requested state is normalized and validated against the public SegmentState names and numeric values before protobuf construction, with invalid labels, out-of-range integers, booleans, and invalid containers raising ParamError. Added request and handler no-RPC coverage; the related suite passes 1,988 tests.

pymilvus/client/grpc_handler.py
MrPresent-Han: Fixed in debfd411: the legacy target fallback is now used only for positive segment IDs, so DataCoord's -1 sentinel produces targets=[] while old responses with a positive target remain supported. Added parser coverage for both cases; the related suite passes 1,988 tests.

pymilvus/client/prepare.py
```diff
@@ -2292,8 +2292,38 @@ def get_collection_stats_request(cls, collection_name: str):
...
+        collection_name: str,
+        states: Optional[Sequence[Union[int, str]]] = None,
+        db_name: str = "",
+    ):
+        if states is None:
+            normalized_states = []
+        else:
+            if isinstance(states, (str, bytes)) or not isinstance(states, Sequence):
+                message = "states must be a sequence of SegmentState names or values"
+                raise ParamError(message)
```
liliu-z: `prepare.py:2306` and `2320` call `raise ParamError(message)`, but `MilvusException.__init__` takes `code` as its first positional parameter (`exceptions.py:27`). Both `Prepare.get_persistent_segment_info_request` validation failures produce an exception whose `.message` is `""` and whose `.code` is a string instead of an int, breaking the documented exception contract for every caller that logs `e.message` or type-checks `e.code`. These are the only two positional `ParamError` constructions in the package; the sibling `Prepare.get_compaction_tasks` added in this same PR (`prepare.py:2537`) uses `message=`.

pymilvus/client/prepare.py
```diff
@@ -2292,8 +2292,38 @@ def get_collection_stats_request(cls, collection_name: str):
...
     @classmethod
-    def get_persistent_segment_info_request(cls, collection_name: str):
-        return milvus_types.GetPersistentSegmentInfoRequest(collectionName=collection_name)
+    def get_persistent_segment_info_request(
+        cls,
+        collection_name: str,
+        states: Optional[Sequence[Union[int, str]]] = None,
+        db_name: str = "",
+    ):
+        if states is None:
```
liliu-z, line 2301: An explicitly empty `states` sequence normalizes to the same empty repeated field as `states=None`, so the request carries no filter and the server falls back to its legacy persistent-state result set. A caller that builds the filter dynamically (`states=[s for s in wanted if s in allowed]`) and ends up with an empty list receives flushed segments instead of zero segments, with no way to detect the difference. This is the wire semantics of a new public API and cannot be changed after release.

pymilvus/milvus_client/milvus_client.py
```diff
@@ -2793,16 +2793,28 @@ def list_loaded_segments(
...
+    def list_serving_segments(
+        self,
+        collection_name: str,
+        timeout: Optional[float] = None,
+        **kwargs,
+    ) -> List[LoadedSegmentInfo]:
+        """List the collection segments currently serving queries."""
+        return self.list_loaded_segments(collection_name, timeout=timeout, **kwargs)
+
     def list_persistent_segments(
```
liliu-z, line 2805: `list_segments(collection_name, states, timeout)` delegates to `list_persistent_segments(collection_name, timeout, states)`, so the second and third positional parameters are reversed between a wrapper and the method it wraps. `client.list_segments("c", ["Flushed"])` works while `client.list_persistent_segments("c", ["Flushed"])` binds the list to `timeout` and forwards it as the gRPC deadline, raising a `TypeError` from deep inside grpc. The handler layer uses yet a third order (`get_persistent_segment_infos(collection_name, timeout, context, states)`, `grpc_handler.py:2279`).

pymilvus/milvus_client/milvus_client.py
```diff
@@ -2849,6 +2886,24 @@ def get_compaction_plans(
             job_id, timeout=timeout, context=self._generate_call_context(**kwargs), **kwargs
         )
 
+    def list_compaction_tasks(
```
liliu-z, line 2889: Every other `list_*` method on `MilvusClient` (`list_segments`, `list_persistent_segments`, `list_loaded_segments`) returns a real list, but `list_compaction_tasks` returns a `CompactionPlans`, so `for task in client.list_compaction_tasks("c")` raises `TypeError: 'CompactionPlans' object is not iterable`. The returned object also always carries `compaction_id = 0` because the collection-scoped path never sets one, and its `__repr__` prints `compaction id: 0` for every collection query.

pymilvus/client/types.py
```diff
@@ -1408,13 +1441,19 @@ def __repr__(self) -> str:
...
+            f"partition_id={self.partition_id}, "
+            f"insert_channel='{self.insert_channel}', "
+            f"compaction_from={self.compaction_from})"
         )
 
 
 @dataclass
 class LoadedSegmentInfo(SegmentInfo):
-    partition_id: int
+    partition_id: int = field()
```
liliu-z, line 1453: `QuerySegmentInfo` has no `insert_channel` or `compaction_from` (`milvus_pb2.pyi:1694`), so `LoadedSegmentInfo` masks both with `init=False` and they are always `""` / `[]`. A caller reading `serving_segment.compaction_from` gets `[]`, which is indistinguishable from "server reported no lineage"; neither field appears in `LoadedSegmentInfo.__repr__` (`types.py:1461`). The masking also breaks substitutability: `SegmentInfo` accepts `insert_channel=` while `LoadedSegmentInfo` rejects it, so `LoadedSegmentInfo(**dataclasses.asdict(seg))` raises `TypeError` where the same round-trip works on the base class.

pymilvus/client/prepare.py
MrPresent-Han: Fixed in f3706ed8: both validation paths now construct ParamError with the message keyword, preserving the default integer error code and intended message. Added assertions for both code and message; the related suite passes 1,990 tests.

pymilvus/client/prepare.py
MrPresent-Han: Fixed in f3706ed8: an explicitly empty states sequence is now rejected with ParamError, while states=None continues to preserve the legacy default filter. Added Prepare and handler pre-RPC coverage; the related suite passes 1,990 tests.

pymilvus/milvus_client/milvus_client.py
MrPresent-Han: Thanks, the positional ordering is inconsistent across these layers. We are keeping the current public method signatures unchanged in this PR to avoid introducing a separate compatibility-sensitive API cleanup; callers of the new states parameter should use the keyword form, and the ordering issue will be handled separately.

pymilvus/milvus_client/milvus_client.py
MrPresent-Han: Agreed that the collection-scoped result is not naturally modeled by a job-scoped CompactionPlans object. We are retaining the existing result model in this PR to keep the observability change focused, and will evaluate a dedicated iterable collection-task result type as a separate API change.

pymilvus/client/types.py
MrPresent-Han: Agreed that QuerySegmentInfo does not provide authoritative insert-channel or lineage metadata. In this PR LoadedSegmentInfo is used only to identify the serving segment set, while ownership and compaction lineage come from the DataCoord-backed list_segments API, so we are leaving this model unchanged and will handle unknown metadata and constructor substitutability separately.

pymilvus/client/grpc_handler.py
```diff
@@ -99,6 +99,60 @@
...
+        Plan(
+            list(merge.sources),
+            merge.target,
+            plan_id=merge.plan_id,
+            trigger_id=merge.trigger_id,
+            collection_id=merge.collection_id,
+            partition_id=merge.partition_id,
+            channel=merge.channel,
+            compaction_type=_COMPACTION_TYPE_NAMES.get(merge.type, f"unknown({merge.type})"),
+            state=_COMPACTION_TASK_STATE_NAMES.get(merge.state, f"unknown({merge.state})"),
```
liliu-z, line 148: `GetPersistentSegmentInfoRequest.states` is a new proto field and milvus-io/milvus#53308 is still open, so every released Milvus ignores it and falls back to the hardcoded `[Flushing, Flushed, Sealed]` filter in `proxy.GetPersistentSegmentInfo`. `grpc_handler.py:2296` returns `response.infos` unfiltered, so `list_persistent_segments("c", states=["Dropped"])` returns live Flushed segments with a Success status and no warning, and `list_segments("c")` silently drops Growing/Importing/Dropped. Callers get segments in states they explicitly excluded, with nothing in the response signalling the filter was discarded.

pymilvus/milvus_client/milvus_client.py
liliu-z, line 2889: `Prepare.get_compaction_tasks` leaves `compactionID` at 0 and relies on `collection_name`, which only milvus-io/milvus#53308 (still open) reads. On current Milvus, DataCoord falls through to `getCompactionInfo(ctx, 0)`, `GetCompactionTasksByTriggerID` returns nothing, and `summaryCompactionState` sets `state = Completed` with empty `mergeInfos` and a Success status. The result is byte-identical to a genuine "collection has no in-flight compaction", so a caller gating on "no in-flight compaction tasks" gets a permanent false all-clear. The docstring at line 2895 documents no server requirement.

pymilvus/client/grpc_handler.py
MrPresent-Han: The existing API semantics remain forward-compatible: when states is omitted, the request keeps the empty-field legacy behavior, so existing callers are unchanged. Supplying states is a new capability that requires the corresponding server support from milvus-io/milvus#53308; we will document that requirement rather than add cross-version capability negotiation in this PR.

pymilvus/milvus_client/milvus_client.py
MrPresent-Han: Correct, collection-scoped lookup did not exist in the previous contract; the legacy path only queries by compactionID and remains unchanged. The ambiguous Completed plus empty result occurs only when the new list_compaction_tasks API is used against a server without milvus-io/milvus#53308, so this new API will be documented as requiring the corresponding server support rather than adding capability negotiation here.

pymilvus/client/async_grpc_handler.py
```diff
@@ -800,10 +800,15 @@ async def get_persistent_segment_infos(
...
         context: Optional[CallContext] = None,
+        states: Optional[Sequence[Union[int, str]]] = None,
         **kwargs,
     ) -> List[milvus_types.PersistentSegmentInfo]:
         check_pass_param(collection_name=collection_name, timeout=timeout)
-        req = Prepare.get_persistent_segment_info_request(collection_name)
+        req = Prepare.get_persistent_segment_info_request(
+            collection_name,
+            states=states,
+            db_name=context.get_db_name() if context else "",
```
XuanYang-cn: All db_name in pymilvus are passing through metadata=_api_level_md(context), we dont need to passin db names here.

pymilvus/client/grpc_handler.py
```diff
@@ -1550,7 +1604,10 @@ def get_query_segment_info(
         context: Optional[CallContext] = None,
         **kwargs,
     ) -> List[milvus_types.QuerySegmentInfo]:
-        req = Prepare.get_query_segment_info_request(collection_name)
+        req = Prepare.get_query_segment_info_request(
+            collection_name,
+            db_name=context.get_db_name() if context else "",
```
XuanYang-cn: ditto

pymilvus/client/grpc_handler.py
```diff
@@ -2224,9 +2281,14 @@ def get_persistent_segment_infos(
...
         timeout: Optional[float] = None,
         context: Optional[CallContext] = None,
+        states: Optional[Sequence[Union[int, str]]] = None,
         **kwargs,
     ) -> List[milvus_types.PersistentSegmentInfo]:
-        req = Prepare.get_persistent_segment_info_request(collection_name)
+        req = Prepare.get_persistent_segment_info_request(
+            collection_name,
+            states=states,
+            db_name=context.get_db_name() if context else "",
```
XuanYang-cn: ditto

pymilvus/client/grpc_handler.py
```diff
@@ -2532,11 +2594,25 @@ def get_compaction_plans(
...
+    def get_compaction_tasks(
+        self,
+        collection_name: str,
+        timeout: Optional[float] = None,
+        context: Optional[CallContext] = None,
+        **kwargs,
+    ) -> CompactionPlans:
+        req = Prepare.get_compaction_tasks(
+            collection_name,
+            db_name=context.get_db_name() if context else "",
```
XuanYang-cn: ditto

pymilvus/client/prepare.py
```diff
@@ -2292,8 +2292,41 @@ def get_collection_stats_request(cls, collection_name: str):
         return milvus_types.GetCollectionStatisticsRequest(collection_name=collection_name)
 
     @classmethod
-    def get_persistent_segment_info_request(cls, collection_name: str):
-        return milvus_types.GetPersistentSegmentInfoRequest(collectionName=collection_name)
+    def get_persistent_segment_info_request(
+        cls,
+        collection_name: str,
+        states: Optional[Sequence[Union[int, str]]] = None,
+        db_name: str = "",
```
XuanYang-cn: ditto

pymilvus/client/prepare.py
```diff
@@ -2302,8 +2335,11 @@ def get_flush_state_request(cls, segment_ids: List[int], collection_name: str, f
         )
 
     @classmethod
-    def get_query_segment_info_request(cls, collection_name: str):
-        return milvus_types.GetQuerySegmentInfoRequest(collectionName=collection_name)
+    def get_query_segment_info_request(cls, collection_name: str, db_name: str = ""):
+        return milvus_types.GetQuerySegmentInfoRequest(
+            dbName=db_name,
```
XuanYang-cn: ditto, these db_names are conflicting with the overall design

pymilvus/client/prepare.py
```diff
@@ -2498,6 +2534,16 @@ def get_compaction_state_with_plans(cls, compaction_id: int):
         request.compactionID = compaction_id
         return request
 
+    @classmethod
+    def get_compaction_tasks(cls, collection_name: str, db_name: str = ""):
+        if not isinstance(collection_name, str) or not collection_name:
+            raise ParamError(message=f"collection_name value {collection_name} is illegal")
+
+        return milvus_types.GetCompactionPlansRequest(
+            db_name=db_name,
```
XuanYang-cn: ditto

XuanYang-cn (changes requested):

pymilvus/client/grpc_handler.py
```diff
@@ -99,6 +99,60 @@
 
 logger = logging.getLogger(__name__)
 
+# Keep the Plan API's names stable across the public protobuf enum migration.
+_COMPACTION_TYPE_NAMES = {
```
XuanYang-cn: 请删除 `_COMPACTION_TYPE_NAMES` 和 `_COMPACTION_TASK_STATE_NAMES`，在 `pymilvus/client/types.py` 中参照 `DataType` 定义 `CompactionType`、`CompactionTaskState`、`SegmentState` 三个 `IntEnum`，成员值直接引用 `common_pb2`，完整覆盖对应 proto 的成员。

`CompactionType` 保留 `UndefinedCompaction`、`MergeCompaction`、`MixCompaction` 等完整名称，可以对照 milvus-io/milvus#53308 确认服务端实际返回哪些类型，但不要据此删减 proto 已定义的成员。

三个枚举统一覆盖 `__str__` / `__repr__`，保持与 `DataType` 一致：

```python
def __repr__(self) -> str:
    return f"<{self.__class__.__name__}: {self._name_}>"

def __str__(self) -> str:
    return self._name_
```

这里统一使用引用 proto 数值的 `IntEnum`；例如 `str(CompactionType.MixCompaction)` 应为 `"MixCompaction"`。

XuanYang-cn (commented):
请按以下 inline comments 统一公开枚举类型、顶层导出和同步/异步能力。未知响应枚举值统一回退到对应枚举的 0 成员。

pymilvus/client/types.py
```diff
@@ -308,28 +308,58 @@ def __repr__(self) -> str:
...
+        sources: list,
+        target: int,
+        *,
+        plan_id: int = 0,
+        trigger_id: int = 0,
+        collection_id: int = 0,
+        partition_id: int = 0,
+        channel: str = "",
+        compaction_type: str = "",
+        state: str = "",
```
XuanYang-cn: 这两个公开字段请使用枚举类型及对应的零值默认值：

```python
compaction_type: CompactionType = CompactionType.UndefinedCompaction
state: CompactionTaskState = CompactionTaskState.Unknown
```

同时在 `pymilvus/__init__.py` 中导入 `CompactionType`、`CompactionTaskState`、`SegmentState` 并加入 `__all__`，让用户可以直接 `from pymilvus import CompactionType, CompactionTaskState, SegmentState`。请补充顶层导入和默认字段实际类型的测试。既有 compaction job 聚合状态与 task 状态含义不同，`CompactionPlans.state` 不应混用 `CompactionTaskState`。

pymilvus/client/grpc_handler.py
XuanYang-cn: 响应解析请改为 `CompactionType(merge.type)` / `CompactionTaskState(merge.state)`，返回枚举实例。

未知响应枚举值的行为已确定：统一回退到对应枚举的 `0` 成员，即 `CompactionType.UndefinedCompaction`、`CompactionTaskState.Unknown`、`SegmentState.SegmentStateNone`。可以在枚举的 `_missing_` 中集中处理未知整数值；不要让未知响应值抛出 `ValueError`，也不要再返回 `unknown(N)` 字符串。

同步、异步共用这一转换规则，并更新已知值、未知值测试，断言返回的是对应枚举实例。此回退规则用于响应解码；调用方传入非法 `states` 仍应按参数校验规则拒绝。

pymilvus/client/prepare.py
```diff
@@ -2292,8 +2292,41 @@ def get_collection_stats_request(cls, collection_name: str):
...
+                message = "states must be a sequence of SegmentState names or values"
+                raise ParamError(message=message)
+            if not states:
+                message = "states must not be empty; use None for the default state filter"
+                raise ParamError(message=message)
+
+            valid_names = set(common_types.SegmentState.keys())
+            valid_values = set(common_types.SegmentState.values())
+            normalized_states = []
+            for state in states:
```
XuanYang-cn, line 2312: `states` 请在 Prepare、同步/异步 handler、同步/异步 client 中统一为 `Optional[Sequence[SegmentState]]`。

保留序列容器与空序列校验，元素只接受 `isinstance(state, SegmentState)`，删除字符串名称和裸整数值的归一化分支。字符串、裸整数、布尔值都应抛出 `ParamError`，并在发起 RPC 前被拒绝。请同步更新签名、docstring 以及参数校验测试。

pymilvus/client/types.py
```diff
@@ -1390,6 +1420,9 @@ class SegmentInfo:
     state: common_pb2.SegmentState
```
XuanYang-cn: 公开返回字段也应统一 typed：请把 `SegmentInfo.state` 改为 SDK 的 `SegmentState`，并在同步、异步客户端构造返回对象时将 proto 值转换为该枚举，保证实际返回枚举实例，而不只是修改注解。相关返回路径（包括继承该字段的 `LoadedSegmentInfo`）保持一致。

未知响应值回退为 `SegmentState.SegmentStateNone`。请同步检查 `state_name` 的实现，并测试已知、未知状态的实际返回类型。

pymilvus/milvus_client/milvus_client.py
```diff
@@ -2811,24 +2814,49 @@ def list_persistent_segments(
...
+        timeout: Optional[float] = None,
+        **kwargs,
+    ) -> List[SegmentInfo]:
+        """List collection segments still retained in the requested lifecycle states.
+
+        Dropped segment metadata is subject to server-side garbage collection, so callers that
+        need lineage history must poll and cache it during the retention window.
+        """
+        if states is None:
+            states = ["Growing", "Sealed", "Flushing", "Flushed", "Importing", "Dropped"]
```
XuanYang-cn: 默认状态请使用同步、异步共用的枚举元组：

```python
_DEFAULT_RETAINED_SEGMENT_STATES = (
    SegmentState.Growing,
    SegmentState.Sealed,
    SegmentState.Flushing,
    SegmentState.Flushed,
    SegmentState.Importing,
    SegmentState.Dropped,
)
```

`list_segments(states=None)` 使用该默认值；既有 `list_persistent_segments(states=None)` 保留原来的服务端默认过滤语义。请覆盖两者默认行为及同步、异步一致性的测试。

pymilvus/milvus_client/milvus_client.py
```diff
@@ -2849,6 +2877,24 @@ def get_compaction_plans(
             job_id, timeout=timeout, context=self._generate_call_context(**kwargs), **kwargs
         )
 
+    def list_compaction_tasks(
```
XuanYang-cn, line 2882: 这次新增的能力请同步补齐异步端：

- `AsyncMilvusClient.list_segments`
- `AsyncMilvusClient.list_compaction_tasks`
- `AsyncGrpcHandler.get_compaction_tasks`

同步、异步的参数、默认过滤、返回类型、调用上下文和错误处理应保持一致，共用枚举定义、默认状态集合及 compaction 响应解析逻辑，并补充对应异步测试。

pymilvus/client/async_grpc_handler.py
MrPresent-Han: Updated as suggested: this async request no longer writes db_name explicitly, and the active database is carried only by _api_level_md(context) metadata; coverage now asserts the metadata path.

pymilvus/client/grpc_handler.py
MrPresent-Han: Updated under the same design: this synchronous request no longer writes db_name explicitly, so the request field remains empty and the database is carried in metadata.

pymilvus/client/grpc_handler.py
MrPresent-Han: Updated this handler to remove redundant db_name request plumbing, with coverage confirming that the context database remains present in RPC metadata.

pymilvus/client/grpc_handler.py
MrPresent-Han: Removed the explicit db_name from the collection-scoped compaction-task request; collection_name remains in the request and the database continues to come from metadata.

pymilvus/client/prepare.py
MrPresent-Han: Removed the db_name parameter and field assignment from Prepare.get_persistent_segment_info_request, so this call chain consistently relies on RPC metadata.

pymilvus/client/prepare.py
MrPresent-Han: Removed the db_name parameter and field assignment from Prepare.get_query_segment_info_request to avoid duplicating the existing database metadata design.

pymilvus/client/prepare.py
MrPresent-Han: Removed the db_name parameter and field assignment from Prepare.get_compaction_tasks while retaining the protocol field for Proxy-side metadata normalization.

pymilvus/client/grpc_handler.py
MrPresent-Han: Removed both string mapping tables and added CompactionType, CompactionTaskState, and SegmentState IntEnums in client/types.py, fully covering the corresponding proto members with consistent name-based str/repr behavior.

pymilvus/client/types.py
MrPresent-Han: Plan.compaction_type and Plan.state now use typed enums with zero-value defaults, all three enums are exported from pymilvus, and CompactionPlans.state remains the distinct job aggregate state.

pymilvus/client/grpc_handler.py
MrPresent-Han: Sync and async paths now share parse_compaction_plans, return enum instances, and use _missing_ to map unknown wire integers to the appropriate zero member, with known and unknown wire-value coverage.

pymilvus/client/prepare.py
MrPresent-Han: states is now Optional[Sequence[SegmentState]] across Prepare, both handlers, and both clients; empty sequences, strings, raw integers, and booleans raise ParamError before the RPC.

pymilvus/client/types.py
MrPresent-Han: SegmentInfo.state is now converted to the SDK SegmentState enum, LoadedSegmentInfo inherits the same behavior, unknown values fall back to SegmentStateNone, and state_name uses the enum name.

pymilvus/milvus_client/milvus_client.py
MrPresent-Han: Sync and async list_segments now share the _DEFAULT_RETAINED_SEGMENT_STATES enum tuple, while list_persistent_segments(states=None) retains the legacy server-side default filter.

pymilvus/milvus_client/milvus_client.py
MrPresent-Han: Added AsyncMilvusClient.list_segments, AsyncMilvusClient.list_compaction_tasks, and AsyncGrpcHandler.get_compaction_tasks, reusing the synchronous enum, default-state, and response-parsing rules.

XuanYang-cn (approved):

PR #3793, opened 2026-09-13
fix: QueryIterator checkpoint file bugs on empty/short cp files and f…

### LINKED ISSUE #3744: [Bug]: QueryIterator cannot resume from a checkpoint file it left behind, and a failed checkpoint write skips a batch

vickyandvicky:
…ailed saves

Two related correctness bugs in QueryIterator's checkpoint handling:

1. __init_cp_file_handler() decided resume-vs-fresh based only on whether the cp file exists, not whether it has usable content. A cp file left behind by a process interrupted before writing anything (0 lines) or interrupted right after the first save (1 line, ts only, no cursor) was incorrectly rejected with ParamError instead of being resumed/started fresh.

2. next() advanced the in-memory cursor before confirming the checkpoint write succeeded. A failed save left the cursor pointing past a batch that was never actually returned to the caller -- silently dropping that batch on any retry.

Fixes #3744


pymilvus/client/iterator/query_iterator.py
```diff
@@ -393,8 +399,23 @@ def next(self):
...
+        # the in-memory cursor pointing past a batch that was never
+        # actually returned to the caller, and that batch is silently
+        # skipped on any retry.
+        prev_next_id = self._next_id
+        prev_next_element_offset = self._next_element_offset
+
         self.__update_cursor(ret)
-        io_operation(self.__save_pk_cursor, "failed to save pk cursor")
+        try:
+            io_operation(self.__save_pk_cursor, "failed to save pk cursor")
```
yhmo, line 413: pymilvus/client/iterator/query_iterator.py line:413
Medium ---- The rollback restores only _next_id and _next_element_offset, not the shared result cache. When the batch is served from the cache (cached_res path), the cache is advanced before the checkpoint write; if the write then fails and the remaining cache is still sufficient (>= 2 * BATCH_SIZE rows, i.e. the query returned >= 4 * BATCH_SIZE rows), a retry serves the batch AFTER the one that was never returned, silently skipping it. With BATCH_SIZE=10 and a 40-row query result: next() #1 caches rows 10-39 and delivers 0-9; next() #2 delivers 10-19 from cache and advances it to 20-39; if the save fails here, the retry delivers 20-29 and rows 10-19 are lost -- the exact silent-drop failure this PR intends to fix. Please roll the cache back on failure as well (or defer the cache advance until after the save succeeds) and add a regression test with a query result >= 4 * BATCH_SIZE rows.

tests/unit/test_query_iterator_checkpoint.py
```diff
@@ -0,0 +1,122 @@
...
+            # Simulate a disk failure on the checkpoint write specifically,
+            # independent of the (already-succeeded) Milvus query.
+            broken_handle = Mock()
+            broken_handle.writelines.side_effect = OSError("simulated disk failure")
+            iterator._cp_file_handler = broken_handle
+
+            with pytest.raises(Exception, match="failed to save pk cursor"):
+                iterator.next()
+
+            assert (
```
yhmo, line 106: tests/unit/test_query_iterator_checkpoint.py line:106
Low ---- The failed-save test only asserts _next_id did not advance. It does not verify that a retry actually re-delivers the same batch (the acceptance criterion of #3744), it does not exercise the cache-hit path (a query result >= 2 * BATCH_SIZE rows) where the rollback in next() is incomplete, and it does not assert _next_element_offset is restored on failure. Suggest adding a cache-path failure test and asserting the element offset rollback.

vickyandvicky: Added a regression test reproducing your exact scenario (BATCH_SIZE=10, 40-row result, failure on the cache-serving call) — confirmed it fails against the previous fix and passes now.
https://github.com/milvus-io/pymilvus/pull/3793/changes/9dc9dbd07a662f1aad3cda4505b59236f6fd2104

pymilvus/client/iterator/query_iterator.py
vickyandvicky: Fixed — the cache advance is now rolled back alongside the cursor fields for both the cache-hit and cache-miss paths.

PR #3805, opened 2026-09-18
[2.6] feat: add scalable Volume upload policies

lentitude2tk:
Backport #3803 to `2.6` (cherry-picked from `6a9a0b4a676f1e43eeeb1caea38dfbcc5f187e67`).

Adds the same four Volume upload policies, defaulting to `SKIP_IF_SAME_SIZE`, with paginated LIST checks, bounded concurrency, an automatic memory-to-SQLite manifest, expired-credential recovery, and preparation/progress reporting. VolumeBulkWriter retains overwrite behavior. No new dependency is required.

Branch adaptation: preserve the existing example while adding the default policy, and replace `Path.is_relative_to` with an equivalent ancestor check so directory-cycle detection works on Python 3.8.

Validation:
- Full unit suite on Python 3.14: 4,034 passed, 3 skipped. The isolated 20,000-path test verifies SQLite spill after 10,000 entries, streaming matching, cleanup, and bounded Python heap.
- Python 3.8 compatibility checks passed for symlink aliases, directory cycles, 10,001-entry SQLite spill/matching/cleanup, and millisecond mtime comparisons. The full Python 3.8 dependency environment was not available locally; that full matrix is left to CI.
- Repository-wide Ruff 0.15.14 and Black 26.5.1 checks passed.
- Sensitive-data and diff checks passed; no live credentials, endpoints, test data, or temporary scripts are included.


XuanYang-cn (approved):
/lgtm

